Issue #292, opened 2015-01-10
Copywriting for fellowship page


olasitarska:
- [x] Checking for typos
- [x] Checking for grammar
- [x] Checking if all the links are properly linked
- [x] Checking if everything "feels" right, or maybe we can put more emotions there

asendecka:
It looks great!

PR #302, opened 2015-01-10
Design and HTML markup for fellowship page

olasitarska:
I used the styling from [overview page](https://www.djangoproject.com/start/overview/) for this. Not sure about the last part (What's Fellowship and Who are Fellows), maybe it should be the same width as the rest? It feels quite good if you look at the website, looks weird on the screenshot though. 

I recommend clicking on the screenshot below and zooming in to 100% for a good experience ;)

screencapture-127-0-0-1-8000-fundraising

Suggestions welcome :) 

cc @asendecka @bmispelon @jezdez @treyhunner


asendecka:
Great work! :+1: 

I suggest putting heart higher, before the text about supporting the program. It would be awesome to see immediately the heart and see call for action on top of the page.

asendecka:
I also think it's a good idea to hide old "Heros": display 4 weeks only and the rest as a toggle.

olasitarska:
@asendecka :heart: 

- Good point with the heart, I think swapping two sections (intro and heart) would be definitely better :) 
- I thought about only showing 2 weeks, notice the pagination after second week -- I thought we could just paginate them there?:)

djangoproject/templates/fundraising/index.html
```diff
@@ -0,0 +1,162 @@
...
+  </div>
+
+</div>
+
+  <div class="section">
+
+  </div>
+  {% endblock %}
+
+  {% block content-related %}
```
timgraham: remove this block? Not sure it'll ever be relevant to this page and if so, I'd expect it'd be an included template.

olasitarska: Definitely, not sure how I missed that! :)

djangoproject/templates/fundraising/index.html
```diff
@@ -0,0 +1,162 @@
...
+
+  {% endblock %}
+
+  {% block content-extra %}
+
+  <div class="layout-tertiary" id="about">
+    <div class="container">
+        <h2>What is Django Fellowship?</h2>
+        <p><a href="https://www.djangoproject.com/weblog/2014/sep/18/dsf-calls-applicants-django-fellow/">Django Fellowship program</a> is a project where paid contractors have been engaged to manage some of the administrative and community management tasks of Django project. The initial pilot took three months and came to close in mid-January; one of the major goals is to secure funding for this program on an ongoing basis.</p>
+        <p>The first three months of the Django Fellowship program have been very successful. In this short time Django Fellows worked on around 240 pull requests (including revising some of the major features that will be in Django 1.8), triaged at least 90 different tickets and made around 200 commits in total!</p>
```
timgraham: It's much easier to review the text if it isn't on one long line. I can do some copyediting on the text -- do you want me to add it to your branch or merge this first and then create a separate PR?

olasitarska: Let's focus on design / html markup here, there is #292 for all copywriting issues (which I'm sure there are :)). Make sense?

timgraham: Sure, overall it looks like a fine starting point then.

olasitarska:
Ok, I'll add some design for testimonial when I get home (in an hour?) and this will be ready to be merged :)

olasitarska:
Final design:

screencapture-127-0-0-1-8000-fundraising 1

PR #314, opened 2015-01-17
Don't reveal heart pixel until pixel subgoal complete

treyhunner:
Problem:

- 1% was filling in 1 heart instead of 0 hearts (it rounded up)
- 99% was filling in all 27 hearts instead of just 26 (also rounded up)

Now instead 1% fills in 0 hearts and 99% fills in all but one heart.


bmispelon:
There are only two hard problems in computer science: cache invalidation, naming things, and off-by-one errors :)

PR #352, opened 2015-01-22
Randomize donations on the fundraising index page

bmispelon:
The randomization is weighted by the donation amount so
that bigger donations have more chances to appear at the top.

This is basically a revert of f1e067e6ccee3a39622042a61badfbf90bde07a9.
Refs #334


fundraising/templatetags/fundraising_extras.py
```diff
@@ -29,3 +30,14 @@ def donation_snippet():
         donation = None
 
     return {'donation': donation}
+
+
+@register.filter
+def shuffle_donations(donations, random=random):
+    """
+    Randomize the give list of donations but in a way that big donations
```
charettes: the given list

fundraising/tests.py
```diff
@@ -29,9 +38,18 @@ def test_in_period_ordering(self):
...
         Donation.objects.create(donor=h3, amount='10')
-        today = date.today()
-        self.assertQuerysetEqual(
-            DjangoHero.objects.in_period(today, today + timedelta(days=1)),
-            [Decimal('15.00'), Decimal('5.00'), Decimal('10.00')],
-            attrgetter('donated_amount')
-        )
+        self.today = date.today()
+
+    def test_donation_suffling(self):
```
jezdez: Please don't pass in a random object but use mock instead, like here: http://stackoverflow.com/a/22457486

fundraising/templatetags/fundraising_extras.py
```diff
@@ -29,3 +30,14 @@ def donation_snippet():
         donation = None
 
     return {'donation': donation}
+
+
+@register.filter
+def shuffle_donations(donations, random=random):
```
jezdez: I'd rather see that applied in the view than the template layer to reduce effort to port this over to Jinja2 later

fundraising/tests.py
bmispelon: `mock` isn't currently installed though, right?

jezdez: nope, add it to the `requirements/tests.txt`

bmispelon:
Switched to `mock` and moved the logic inside the view rather than the template.

Issue #364, opened 2015-01-27
Large payload for fundraising page


treyhunner:
The [fundraising page][] requires a 4MB network transfer, mostly due to all of the images on that page.

The 5 largest avatars take up 2MB on their own and some are 1000 x 1000 pixels.

Some things that might help:

- Enable gzip/deflate encodings for smaller network transfer size
- Use [easy-thumbnails][] to create small thumbnails

[easy-thumbnails]: https://github.com/SmileyChris/easy-thumbnails
[fundraising page]: https://www.djangoproject.com/fundraising/

jezdez:
Gzip is enabled for files where it makes sense, basically text based files. I don't think it's common practice to enable it for image files. But I agree that thumbnailing the images to a sensible size makes sense. I'm a bit reluctant to add easy-thumbnails though since I'm not certain about its maintenance status. I think sorl-thumbnail is nowadays better supported and works on 1.7 and Python 3: https://github.com/mariocesar/sorl-thumbnail

treyhunner:
Yes I forgot about the inefficiency of compressing images. It does seem like sorl-thumbnail is the best solution.

PR #365, opened 2015-01-27
Create thumbnails for django-hero logos

relekang:
Related to #364

### DIFF
--- a/fundraising/tests.py
+++ b/fundraising/tests.py
@@ -235,14 +235,18 @@ class TestDjangoHero(TestCase):
                 )
 
     def test_thumbnail(self):
+        try:
+            os.makedirs(os.path.join(settings.MEDIA_ROOT, 'fundraising/logos/'))
+        except OSError:  # directory may already exist
+            pass
         image_path = os.path.join(settings.MEDIA_ROOT, 'fundraising/logos/test_logo.jpg')
         image = Image.new('L', (500, 500))
         image.save(image_path)
         self.h1.logo = image_path
         self.h1.save()
         thumbnail = self.h1.thumbnail
-        self.assertEqual(thumbnail.x, 340)
-        self.assertEqual(thumbnail.y, 340)
+        self.assertEqual(thumbnail.x, 170)
+        self.assertEqual(thumbnail.y, 170)
         os.remove(image_path)
         self.assertTrue(
             os.path.exists(
--- a/fundraising/tests.py
+++ b/fundraising/tests.py
@@ -209,14 +209,14 @@ class TestDjangoHero(TestCase):
             'is_visible': True,
             'is_amount_displayed': True,
         }
-        h1 = DjangoHero.objects.create(**kwargs)
-        Donation.objects.create(donor=h1, amount='5')
-        h2 = DjangoHero.objects.create(**kwargs)
-        Donation.objects.create(donor=h2, amount='15')
+        self.h1 = DjangoHero.objects.create(**kwargs)
+        Donation.objects.create(donor=self.h1, amount='5')
+        self.h2 = DjangoHero.objects.create(**kwargs)
+        Donation.objects.create(donor=self.h2, amount='15')
         # hidden donation amount should display last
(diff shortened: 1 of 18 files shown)


jezdez:
@relekang I think it'd be awesome to use sorl-thumbnails' ImageField and admin integration so that we can see the thumbnail there as well. http://sorl-thumbnail.readthedocs.org/en/latest/examples.html

About the 340px width and height, is that related to the 170px max-height currently set to images? 2x for retina? Is this something we should conditionally handle via http://sorl-thumbnail.readthedocs.org/en/latest/template.html#resolution?

jezdez:
Ah, also, is there a reason not to use the thumbnail template tag btw?

jezdez:
Could you set `THUMBNAIL_DEBUG` to `True` in the dev settings?

How about setting `THUMBNAIL_FORMAT` to `'PNG'` so that we can preserve transparency? Alternatively set `THUMBNAIL_PRESERVE_FORMAT` to `True`?

relekang:
> @relekang I think it'd be awesome to use sorl-thumbnails' ImageField and admin integration so that we can see the thumbnail there as well. http://sorl-thumbnail.readthedocs.org/en/latest/examples.html

I can add that :)

>About the 340px width and height, is that related to the 170px max-height currently set to images? 2x for retina? Is this something we should conditionally handle via http://sorl-thumbnail.readthedocs.org/en/latest/template.html#resolution?

I thought that for images that small it would not make a large enough impact to not just use 2x on all instead of adding javascript to check the dpi. I can change that by using `THUMBNAIL_ALTERNATIVE_RESOLUTIONS` and something like [retina.js](http://imulus.github.io/retinajs/).

> Ah, also, is there a reason not to use the thumbnail template tag btw?

Yes, using `[hero.thumbnail for hero in DjangoHero.objects.all().exclude(logo=None)]` in shell will generate all thumbnails, and it makes it easier to test.

> Could you set THUMBNAIL_DEBUG to True in the dev settings?

Sure, I'll fix it later today.

>How about setting THUMBNAIL_FORMAT to 'PNG' so that we can preserve transparency? Alternatively set THUMBNAIL_PRESERVE_FORMAT to True?

Sounds like a good idea.

jezdez:
About checking with retina.js, we're already using Modernizr, so maybe someting like https://github.com/joaocunha/modernizr-retina-test is enough?

relekang:
We still have to change the path of the src on the img tag, but it is just a string replace.

fundraising/tests.py
```diff
@@ -232,6 +234,25 @@ def test_donation_shuffling(self):
                     attrgetter('donated_amount')
                 )
 
+    def test_thumbnail(self):
+        image_path = os.path.join(settings.MEDIA_ROOT, 'fundraising/logos/test_logo.jpg')
```
jezdez: Seems like we need to create that directory first

jezdez:
Nice work @relekang!

fundraising/tests.py
relekang: Yeah, somehow I managed [to do this](https://github.com/relekang/djangoproject.com/commit/a387effe289454601c20643812bdf450663f6e75#diff-dd3055d7945a1d7f057a725ff660ef79L238). Fixed it now :)

fundraising/tests.py
```diff
@@ -232,6 +234,29 @@ def test_donation_shuffling(self):
...
+            os.makedirs(os.path.join(settings.MEDIA_ROOT, 'fundraising/logos/'))
+        except OSError:  # directory may already exist
+            pass
+        image_path = os.path.join(settings.MEDIA_ROOT, 'fundraising/logos/test_logo.jpg')
+        image = Image.new('L', (500, 500))
+        image.save(image_path)
+        self.h1.logo = image_path
+        self.h1.save()
+        thumbnail = self.h1.thumbnail
+        self.assertEqual(thumbnail.x, 340)
```
jezdez: Must be 170 :)

relekang: Ah, yeah. Forgot that

PR #376, opened 2015-02-03
Handled anonymous donors on the fundraising index page

bmispelon:
If someone opted to appear in the list but didn't provide
a name, "Anonymous Hero" will be used instead.

### DIFF
--- a/djangoproject/templates/fundraising/index.html
+++ b/djangoproject/templates/fundraising/index.html
@@ -123,7 +123,7 @@
         </div>
         <div class="hero-name">
           {% if donor.url %}<a href="{{ donor.url }}" rel="nofollow">{% endif %}{% spaceless %}
-            {{ donor.name }}
+            {{ donor.name_with_fallback }}
           {% endspaceless %}{% if donor.url %}</a>{% endif %}
           {% if donor.is_amount_displayed %}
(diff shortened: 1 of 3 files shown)


djangoproject/templates/fundraising/index.html
```diff
@@ -123,7 +123,7 @@ <h1 class="visuallyhidden">Support Django</h1>
         </div>
         <div class="hero-name">
           {% if donor.url %}<a href="{{ donor.url }}" rel="nofollow">{% endif %}{% spaceless %}
-            {{ donor.name }}
+            {{ donor.name|default:'Anonymous Hero' }}
```
jezdez: Could we move that to a model method to not have to repeat ourselves here and make that translation aware?

fundraising/models.py
```diff
@@ -81,6 +81,10 @@ class Meta:
     def thumbnail(self):
         return get_thumbnail(self.logo, '170x170', quality=100)
 
+    @property
+    def name_with_fallback(self):
+        return self.name if self.name else 'Anonymous Hero'
```
jezdez: Mark `Anonymous Hero` for translation please

jezdez: Scratch that, let's do it in #377

jezdez:
LGTM

Issue #381, opened 2015-02-08
Support for multiple fundraisings


olasitarska:
We plan to run a couple of fundraising in the future, so it'd be nice to create a Fundraising model that would tie donations to specific campaigns. 

This ticket also covers displaying fundraisings on individual pages for each fundraising with a unique URL

PR #387, opened 2015-02-09
Fundraising multi campaign support

olasitarska:
This pretty much covers the whole #381 issue. It turned out to be a really big refactor, but hopefully for the best :)


olasitarska:
Oh, I don't think I want to merge it into master, but rather something like `fundraising-phase-2`. Closing this, second pr coming soon :)

PR #396, opened 2015-02-23
Fix #doc-versions block background

a1tus:
In versions block switcher every link has border-radius but background is set to `li` items so on dark background we can see an issue (see screenshot).

doc-versions


bmispelon:
Nice catch, thanks!

:octopus:

PR #481, opened 2015-06-04
Update Travis and Coveralls badges

xavierdutreilh:
Hello again,

I have updated the badges for Travis-CI and Coveralls in the README. There are two reasons to that:
1. The minor reason is that URLs have been changed. And even if the old ones are still working, it needs to be changed.
2. The major reason is that the badges (in particular the one for Travis) report *the results for the last build*, not *the results for the last build on the master branch*. So, if any pull request introduces some changes that break the source code, then the badge reports a broken build, even if the master branch is fine.

EDIT : This is what the README looks like at the moment:
screen shot 2015-06-04 at 10 30 13
and this is what it looks like with the fix:
screen shot 2015-06-04 at 11 42 29

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,11 +1,11 @@
 djangoproject.com source code
 =============================
 
-.. image:: https://img.shields.io/travis/django/djangoproject.com.svg
-    :target: http://travis-ci.org/django/djangoproject.com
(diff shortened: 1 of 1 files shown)


olasitarska:
:sparkles: :rocket: thank you!

Issue #470, opened 2015-05-21
Definition term css could benefit from some styling


claudep:
Currently, definition terms (`<dt>` tags) are not styled. See for example https://docs.djangoproject.com/en/1.8/topics/i18n/#definitions
This could be improved, IMHO.

PR #482, opened 2015-06-04
Style definition terms in glossaries

xavierdutreilh:
Hello again,

As mentionned in #470, definition terms (`<dt>` tags) are not styled in glossaries. So, you have a "hard" time reading them and guessing which one is the term and which one is the definition. Currently, it means that they look like this on the current website:
screen shot 2015-06-04 at 14 08 09

After applying some very basic styling, they become more apparent:
screen shot 2015-06-04 at 14 07 37

Please note that I have been very careful. So, the CSS rule only applies to definition terms in glossaries (elements using the class `glossary`) and it does not affect all other definition terms like those in the right sidebar.

Voilà, and feel free to ask for modifications. I am not too familiar with the design rules for djangoproject.com. :)

### DIFF
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -71,6 +71,12 @@ dl {
 			margin-bottom: 0;
 		}
 	}
+
+	&.glossary {
+		dt {
+			font-weight: 600;
+		}
+	}
 }
 h1,
 h2,


olasitarska:
It looks good! Thanks! :sparkling_heart: 

If you're planning to make any more changes to the design, here is a helpful style guide: http://djangoproject.com/styleguide/

PR #595, opened 2015-12-17
Improve heart filling

treyhunner:
I got some feedback from fresh eyes (friends who hadn't seen this page before). Feel free to cherry pick specific commits if you don't want any of them (they should be fairly atomic).

Changes:

- Fill heart from the bottom up instead of randomly
- Remove (somewhat confusing) hover effect on heart pixels that also doesn't work over text sections
- Make overlaid heart percentage not act like text
- Don't randomly move heart pixels (just leave them static)
- Make pixel pulsing more colorful/contrasted and empty pixels slightly more colored/opaque


treyhunner:
Demos

timgraham:
Looks nice, thanks Trey!

PR #640, opened 2016-02-25
Fixed a 404 on the Diversity page

ewjoachim:
The GitHub link was not pointing to the right path.

Note : if accepted, this commit will be listed in the page pointed by the url it corrects ! :smile:


ewjoachim:
Yay, thank you !

PR #656, opened 2016-04-10
Made the surrounding div for the language and version chooser click-through

MarkusH:


timgraham:
I think I understand what the intent is but I can't detect a behavior change on Firefox. Could you explain this a bit more?

MarkusH:
The problem is that long items in a TOC of a documentation page may be underneath the language switcher which is notably smaller than the version switcher.

The div `#version-switcher` covers parts of the TOC. 
screenshot1

When you want to click a TOC link somewhere in that area the link is not clickable
screenshot2

With this change the `#version-chooser` has a click-through behavior and the `ul>li` tags have the normal behavior. The `ul` tags have the normal behavior when the `li` tags are hovered to allow for selection of the choices (otherwise the chooser would collapse when moving the mouse out of the current shown version)

MarkusH:
Thanks for noticing, @timgraham. There was indeed some weird behavior with FF going on. Works for me, now.

timgraham:
Looks good.

PR #667, opened 2016-06-30
PyCharm Fundraising Promotion

frankwiles:
- Added block to allow overriding header containers
    - Added promotion banner image
    - Added .copy-banner-black SCSS to change background color
    - FYI: This will be removed with another PR at the end of the
      fundraising campaign. Please deploy as soon as possible thanks!


frankwiles:
Awesome thanks!

PR #674, opened 2016-07-02
Fixed #660 -- Added a `min_score` clause of .01 to ES search

### LINKED ISSUE #660: Search yields the same results regardless of the query
I'm not sure how long this has been happening but it seems that searching for anything always yields the same number of results (360 when searching in the 1.9 version).

For example, the search for [`reinhardt`](https://docs.djangoproject.com/en/1.9/search/?q=reinhardt) and for [`asdf`](https://docs.djangoproject.com/en/1.9/search/?q=asdf) both yield 360 results.

Maybe related to #636?

nanuxbe:
.01 is low enough to show a significant amount of results while weeding
out results with very low (ie. 0.0) scores


bmispelon:
Thanks! 🚴

PR #675, opened 2016-07-02
Fixed #670 -- Added a `messages` block to the main template

### LINKED ISSUE #670: RSS feed submission
When submitting a new rss feed to the "community" section of the message, upon form submission, you are sent back to the list page without any failure or success message

nanuxbe:
Also overridden said block for existing messages in fundraising/index and fundraising/manage-donations


bmispelon:
Could you post some screenshot of how it looks like?

I'd also be interested to see if we can get rid of the `<h1>` message display in the fundraising app and use the `base.html` default display. How would that look?

Thanks!

nanuxbe:
Here are a few screenshots

screenshot 2016-07-02 at 18 06 13

screenshot 2016-07-02 at 18 10 00


screenshot 2016-07-02 at 18 12 31

bmispelon:
Big :+1: from me, but I would appreciate another set of eyes.

PR #680, opened 2016-07-05
Increased DocumentRelease.lang max_length

claudep:


claudep:
I think that URLs are OK because pt_BR is "pt-br" (BCP format) in settings.LANGUAGES.

PR #795, opened 2017-11-09
Fix a few template errors

motet-a:
Last few days, I wrote a [quick-and-dirty linter for Jinja-like HTML
templates][jinjalint]. I tried it on this project and found out a few
HTML errors, most of them are forgotten or mismatched closing tags.

I also fixed a few other quirks that looked wrong to me. Feel free to leave
any feedback.

[jinjalint]: https://github.com/motet-a/jinjalint

### DIFF
--- a/djangoproject/templates/styleguide.html
+++ b/djangoproject/templates/styleguide.html
@@ -217,7 +217,7 @@
             <option>Option 3</option>
           </select>
         </fieldset>
-          
+
         <fieldset>
           <label>
             <input type="checkbox" />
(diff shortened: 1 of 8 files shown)


djangoproject/templates/styleguide.html
```diff
@@ -267,15 +267,24 @@ <h3 id="elements-lists">
...
+                  <li><a class="reference internal" href="#saving-foreignkey-and-manytomanyfield-fields">Saving <tt class="docutils literal"><span class="pre">ForeignKey</span></tt> and <tt class="docutils literal"><span class="pre">ManyToManyField</span></tt> fields</a></li>
+                </ul>
+              </li>
+            </ul>
           </li>
-          <li><a class="reference internal" href="#retrieving-objects">Retrieving objects</a><ul>
-          <li><a class="reference internal" href="#retrieving-all-objects">Retrieving all objects</a></li>
-          <li><a class="reference internal" href="#retrieving-specific-objects-with-filters">Retrieving specific objects with filters</a>
+          <li>
+            <a class="reference internal" href="#retrieving-objects">Retrieving objects</a>
```
timgraham: I think the changes here aren't correct. "Making queries" should be the only top-level heading but this moves "Retrieving objects" to the top-level.

motet-a: You are right! To be fair, I didn't install this website and didn't make sure this template renders properly.

Here is how the fixed markup renders (without styles):

image

PR #812, opened 2018-01-29
Add new index to optimize filters in docs search.

pauloxnet:
Without Index

```SQL
QUERY PLAN
Limit  (cost=193.72..193.72 rows=1 width=544) (actual time=0.737..0.737 rows=0 loops=1)
  ->  Sort  (cost=193.72..193.72 rows=1 width=544) (actual time=0.735..0.735 rows=0 loops=1)
        Sort Key: id
        Sort Method: quicksort  Memory: 25kB
        ->  Seq Scan on docs_document  (cost=0.00..193.71 rows=1 width=544) (actual time=0.696..0.696 rows=0 loops=1)
              Filter: ((release_id = 22) AND ((title)::text = 'json'::text))
              Rows Removed by Filter: 1114
Planning time: 0.629 ms
Execution time: 0.819 ms
```

With Index

```SQL
QUERY PLAN
Limit  (cost=8.31..8.31 rows=1 width=544) (actual time=0.031..0.031 rows=0 loops=1)
  ->  Sort  (cost=8.31..8.31 rows=1 width=544) (actual time=0.030..0.030 rows=0 loops=1)
        Sort Key: id
        Sort Method: quicksort  Memory: 25kB
        ->  Index Scan using document_release_title_idx on docs_document  (cost=0.28..8.30 rows=1 width=544) (actual time=0.011..0.011 rows=0 loops=1)
              Index Cond: ((release_id = 22) AND ((title)::text = 'json'::text))
Planning time: 0.436 ms
Execution time: 0.064 ms
```


pauloxnet:
@timgraham is that commit small enough to be merged with no side effects ?

PR #830, opened 2018-05-01
Isolated building of DocumentReleases in a separate update_docs method

claudep:


claudep:
The diff is unfortunately not very readable, but basically this was just transforming the content of the for loop in a new method.

PR #847, opened 2018-10-12
Add PyCharm 2018 Fundraiser Banner

frankwiles:
Please merge as soon as possible so we can get the banner live on the site prior to DjangoCon.  The promotion is live so merge away!


frankwiles:
Awesome Tim thank you!

PR #862, opened 2019-02-17
disable LocaleMiddleware for docs.djangoproject.com

tobiasmcnulty:
When looking into adding a CDN, I noticed docs.djangoproject.com is setting `Vary: Cookie` and `Vary: Accept-Language` headers, neither of which should be needed (there are no cookies, and we use URL-based rather than header-based language selection).

While no other apps currently have translation, work is in progress (#866) to enable that to happen, so we may not want to disable `LocaleMiddleware` entirely.

This PR selectively disables `LocaleMiddleware` for the domain(s) specified in settings, and standardizes how language activation is done in the `docs` app to always use the language in the URL (rather than a mix of the two).

### DIFF
--- a/djangoproject/middleware.py
+++ b/djangoproject/middleware.py
@@ -1,3 +1,10 @@
+from django.conf import settings
+from django.http.request import split_domain_port
+from django.middleware.locale import LocaleMiddleware
+from django.utils.functional import cached_property
+from django.utils.http import is_same_domain
+
+
 class CORSMiddleware(object):
     """
     Set the CORS 'Access-Control-Allow-Origin' header to allow the debug
@@ -9,4 +16,34 @@ class CORSMiddleware(object):
     def __call__(self, request):
         response = self.get_response(request)
         response['Access-Control-Allow-Origin'] = '*'
+        return response
+
+
+class ExcludeHostsLocaleMiddleware(LocaleMiddleware):
+    """
+    Locale middleware that lets us exclude requests to certain hosts (e.g.,
+    docs.djangoproject.com) from being processed by LocaleMiddleware.
+    """
+
+    @cached_property
+    def _excluded_hosts(self):
+        return frozenset(getattr(settings, 'LOCALE_MIDDLEWARE_EXCLUDED_HOSTS', []))
+
+    def _is_host_included(self, host):
+        """
+        Mirrors the behavior of django.http.request.validate_host(), but does
+        not match '*' (which would exclude all hosts). To exclude all requests
+        from being processed by LocaleMiddleware one should simply remove this
+        class from settings.MIDDLEWARE.
+        """
+        domain, _ = split_domain_port(host)
+        return not any(is_same_domain(domain, pattern) for pattern in self._excluded_hosts)
+
+    def process_request(self, request):
+        if self._is_host_included(request.get_host()):
+            super().process_request(request)
+
+    def process_response(self, request, response):
+        if self._is_host_included(request.get_host()):
+            return super().process_response(request, response)
         return response
--- a/djangoproject/tests.py
+++ b/djangoproject/tests.py
@@ -1,5 +1,9 @@
+from http import HTTPStatus
+
 from django.test import TestCase
 from django_hosts.resolvers import reverse
+
+from docs.models import DocumentRelease, Release
 
 
 class TemplateViewTests(TestCase):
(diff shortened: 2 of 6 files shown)


charettes:
While it looks like the removal of `LocaleMiddleware` could deal with the unintentional weird translation of timestamps



I wonder if we shouldn't aim at translating more parts of the site instead. By the way, [searching for `trans` in the existing HTML templates](https://github.com/django/djangoproject.com/search?l=HTML&q=trans) it looks like it's not only the docs that are translated and we still need to `Vary: Accept-Language`.

Happy to provide help with the French translation of the missing strings as well

tobiasmcnulty:
Thanks @charettes -- to be clear I am not suggesting at all that we don't try to translate more parts of the site, but rather, that `LocaleMiddleware` may not be necessary for switching languages. It checks in many places to try to do the right thing, which is why we end up with all these `Vary:` headers. I feel something more granular may be a better fit...which maybe means I am volunteering myself to add some configurability to `LocaleMiddleware`...

I am missing something with `makemessages` right now, too; I think its setup may have broken in a Django upgrade since the command was last run.

It breaks things if I run it from the project directory, in part because the `docs` templates are in the `djangoproject` folder, I think. I *believe* the options are to move the `locale` directory up one (and have it house all the translations for the project in one place, though that seems to contradict what the i18n docs say about where this directory should live), OR move the `docs` templates inside the `docs` app.

But I am mostly guessing here...someone who knows how this should work please correct me. 🙂

tobiasmcnulty:
I put up two PRs, #864 and #865, with what I think are the competing approaches for getting translations updated. Having gone through this exercise I think I prefer Option 2, but again, would welcome any and all comments/input!

tobiasmcnulty:
@charettes How's this look, as an alternative to removing it completely?

djangoproject/middleware.py
```diff
@@ -10,3 +14,23 @@ def __call__(self, request):
...
+
+
+class ExcludeDomainsLocaleMiddleware(LocaleMiddleware):
+    """
+    Locale middleware that lets us exclude requests on certain domains (e.g.,
+    docs.djangoproject.com) from being processed by LocaleMiddleware.
+    """
+
+    def _is_domain_excluded(self, request):
+        excluded_domains = getattr(settings, 'LOCALE_MIDDLEWARE_EXCLUDED_DOMAINS', [])
```
charettes: What about making `excluded_domain` a cached property of `ExcludeDomainsLocaleMiddleware` that turns `LOCALE_MIDDLEWARE_EXCLUDED_DOMAINS` into a frozen set. Since it's only used for containment checks and doesn't depend on ordering there's no reason that a list should be used over a set.

This might make testing a bit harder but given this is going to be executed for every docs request I think it's worth it.

djangoproject/tests.py
```diff
@@ -36,3 +41,51 @@ def test_conduct_changes(self):
...
+
+    @classmethod
+    def setUpTestData(cls):
+        r2 = Release.objects.create(version='2.0')
+        DocumentRelease.objects.create(lang='en', release=r2, is_default=True)
+
+    @classmethod
+    def tearDownClass(cls):
+        # cleanup URLconfs changed by django-hosts
+        set_urlconf(None)
```
charettes: Is this really required anymore? I thought https://github.com/jazzband/django-hosts/commit/ec5e2931d90167c1c7df262e725457c02c399d38 fixed it.

djangoproject/tests.py
```diff
@@ -36,3 +41,51 @@ def test_conduct_changes(self):
...
+        super().tearDownClass()
+
+    def test_docs_domain_excluded(self):
+        "We get no Content-Language or Vary headers when docs domain is excluded"
+        with self.settings(LOCALE_MIDDLEWARE_EXCLUDED_DOMAINS=[self.docs_domain]):
+            resp = self.client.get('/', HTTP_HOST=self.docs_domain)
+        self.assertEqual(resp.status_code, HTTPStatus.FOUND)
+        self.assertFalse('Content-Language' in resp)
+        self.assertFalse('Vary' in resp)
```
charettes: `assertNotIn`

djangoproject/tests.py
```diff
@@ -36,3 +41,51 @@ def test_conduct_changes(self):
...
+
+    def test_docs_domain_not_excluded(self):
+        "We still get Content-Language when docs domain is not excluded"
+        with self.settings(LOCALE_MIDDLEWARE_EXCLUDED_DOMAINS=[]):
+            resp = self.client.get('/', HTTP_HOST=self.docs_domain)
+        self.assertEqual(resp.status_code, HTTPStatus.FOUND)
+        self.assertTrue('Content-Language' in resp)
+        self.assertTrue('Vary' in resp)
```
charettes: `assertIn`

djangoproject/tests.py
```diff
@@ -36,3 +41,51 @@ def test_conduct_changes(self):
...
+
+    def test_www_domain(self):
+        "www should still use LocaleMiddleware"
+        with self.settings(LOCALE_MIDDLEWARE_EXCLUDED_DOMAINS=[self.docs_domain]):
+            resp = self.client.get('/', HTTP_HOST=self.www_domain)
+        self.assertEqual(resp.status_code, HTTPStatus.OK)
+        self.assertTrue('Content-Language' in resp)
+        self.assertTrue('Vary' in resp)
```
charettes: `assertIn`

djangoproject/middleware.py
```diff
@@ -10,3 +14,23 @@ def __call__(self, request):
...
+        return 'HTTP_HOST' in request.META and request.META['HTTP_HOST'] in excluded_domains
```
charettes: You want [`request.get_host()` here to make sure proxying headers are taken into consideration](https://docs.djangoproject.com/en/2.1/ref/request-response/#django.http.HttpRequest.get_host)

```suggestion
        return request.get_host() in self.excluded_domains
```

djangoproject/middleware.py
```diff
@@ -10,3 +14,23 @@ def __call__(self, request):
...
         return response
+
+
+class ExcludeDomainsLocaleMiddleware(LocaleMiddleware):
+    """
+    Locale middleware that lets us exclude requests on certain domains (e.g.,
+    docs.djangoproject.com) from being processed by LocaleMiddleware.
+    """
+
+    def _is_domain_excluded(self, request):
```
charettes: I'd name the method `_is_domain_included` and push down the negation handling to it; it felt a bit weird to have a _private_ method which have its only callers negate its return value. This also avoids the _not_ _excluded_ double negation burden.

tobiasmcnulty:
@charettes Thanks for the feedback! I've implemented all of these changes. I also:

- changed "domain" to "host" in all cases to remain consistent with `ALLOWED_HOSTS`, etc.
- added a test for the proxied host case

Note that `X-Forwarded-Host` will not actually be seen by the site right now, because we don't set `USE_X_FORWARDED_HOST`. That said, I already have Fastly forwarding the `Host` header as-is, and the docs seem to suggest that you shouldn't set `USE_X_FORWARDED_HOST` unless you really need it, so I think we're fine without it for now (but agreed it's better to support that in `ExcludeHostsLocaleMiddleware` in case it is needed in the future, especially since the whole point is to optimize docs.djangoproject.com to sit behind a proxy).

FWIW, using `@cached_property` didn't require any hoop jumping in the tests. I'm not sure if this is due to something the test client is already doing, or if Django is already instantiating a new middleware for each request...

djangoproject/tests.py
tobiasmcnulty: Turns out it's not. I was combatting a test that was failing only on CI and threw too many things at it at once, I guess. 🙂

tobiasmcnulty:
One more last minute change: I realized this didn't handle the case when `get_host()` returns a port, so I added tests for that & updated `_is_host_included()` to mimic the behavior of `django.http.request.validate_host()`:

https://github.com/django/djangoproject.com/blob/34efe5d7eea8637f7aa735a29ed40c5f4afebbab/djangoproject/middleware.py#L32-L40

I'm going to merge and deploy this momentarily. Thanks again for the review!

PR #866, opened 2019-02-18
Reorganized locale directories and added Transifex for managing the translation process

tobiasmcnulty:
We had some old PO files for English and French in the repo, but they were sorely out of date and it's high time we switched to something more modern for managing the translation process. I also discovered recently that we were not running `compilemessages` during deploy, so the few translations we did have weren't even showing up. That has now been fixed, so let's add and update translations! 😄 

This PR:

- Adds a new `locale` directories in the `dashboard` app and at the project level (in addition to `docs/locale/` that we already had). This gives us one `locale/` directory for each website domain and allows us to map the corresponding PO files to different places in Transifex.
- Adds `transifex-client` to the dev requirements, a new `.tx/config` to do the locale -> resource mapping, and some process docs for managing translations.
- Updates the one translation that we already had (French) with the updated translation files from Transifex
- Includes a script to update translations in the future, based on: https://github.com/django/django-docs-translations/blob/stable/2.1.x/manage_translations.py#L54-L72

Much credit due to @m-aciek and @claudep for helping to brainstorm and arrive at this solution!

**One question:** I copied the list of supported languages from the script in django-docs-translations, but I'm not sure if we need to keep those in sync or not. I guess we should at least aim to support at least those languages but potentially more, for djangoproject.com...(?)

Finally, a few things that can or will happen separately from this PR:
- I opted not to include translations of other languages in this stage to help keep the size of the PR down. But if/when this is approved and merged, we should pull and commit all the new translations shortly after.
- I also opted not to add any new strings (yet). There is a separate issue for localizing the fundraising app which could be a good first step after this (#377).
- We'll need to decide how we want language switching to work for sites other than docs. See #862.


tobiasmcnulty:
Thanks for all your help and feedback @claudep and @m-aciek.

I think this should be ready for another review.

tobiasmcnulty:
And your translation at work: https://www.dropbox.com/s/6vjkpxhmbde61dk/Screenshot%202019-02-18%2021.57.32.png?dl=0 😄

tobiasmcnulty:
@m-aciek Would you like to give this another review?

And/or @timgraham would you like to review before merging?

Thanks!

README.rst
```diff
@@ -287,3 +287,77 @@ from a copy of the production database and saved to the
...
+You'll need to be a member of the Django team in the `Django <https://www.transifex.com/django/>`_
+organization at Transifex. For information on how to join, please see the `Translations
+<https://docs.djangoproject.com/en/dev/internals/contributing/localizing/#translations>`_
+section of the documentation on contributing to and localizing Django.
+
+Since this repo hosts three separate sites, our ``.po`` files are organized by
+website domain. At the moment, we have:
+
+* ``dashboard/locale/`` contains the translation files for https://dashboard.djangoproject.com
+* ``docs/locale/`` contains the translation files for https://docs.djangoproject.com
```
m-aciek:
```suggestion
* ``docs/locale/`` contains the translation files for app on https://docs.djangoproject.com
```

README.rst
```diff
@@ -287,3 +287,77 @@ from a copy of the production database and saved to the
...
+   translations down from Transifex.)
+
+2. Push the updated source file to Transifex::
+
+     tx push -s
+
+3. Commit and push the changes to github::
+
+     git commit -m "Updated messages" locale/en/LC_MESSAGES/*
+     git push
```
m-aciek: Please note we don't store documentation source strings in django nor django-docs-translation repositories. It is a bit redundant as those PO files are constructed basing only on apps' source files.

On the other hand it gives us a bit more control of what was sent to Transifex or should be.

What do you think?

If we would be about to not store source PO file in repository, we should delete existing ones and may add locale/en/LC_MESSAGES/* to .gitignore.

dashboard/locale/fr/LC_MESSAGES/django.po
```diff
@@ -0,0 +1,26 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the PACKAGE package.
```
m-aciek: Docs' PO files have sth more similar to:
```suggestion
# Copyright (C) Django Software Foundation and contributors
# This file is distributed under the same license as the djangoproject.com source files.
```
here.

Should it be set in Transifex to work automatically?

If it is configurable, I'd also suggest to drop the lines `SOME DESCRIPTIVE TITLE.` and `FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.`.

README.rst
claudep: This is the mechanism by which we inform Transifex about the source messages. Changing that (if possible at all) would be the subject of another issue.

dashboard/locale/fr/LC_MESSAGES/django.po
claudep: I guess this is mostly controlled by Transifex, I don't know yet if that can be customized or not. We could also remove French files from this commit, and add them with other languages in a separate PR.

README.rst
tobiasmcnulty: I would prefer to have them all in the repo (source and translation PO files). Then we have a record of when/how the strings are updated. For the translation files this is also how they will get out to the server.

dashboard/locale/fr/LC_MESSAGES/django.po
tobiasmcnulty: If it can be changed in Transifex that would be great. Do you know if that's possible and where to do it if so? I'm not finding anything in the UI or their documentation...

Or for the Django docs is it handled through post-processing (like removal of the file locations)?

README.rst
```diff
@@ -287,3 +287,78 @@ from a copy of the production database and saved to the
...
+Translation
+-----------
+
+We're using Transifex to help manage the translation process. The
+``requirements/dev.txt`` file will install the Transifex client.
+
+Before using the command-line Transifex client, create ``~/.transifexrc``
+according to the instructions at https://docs.transifex.com/client/client-configuration.
+You'll need to be a member of the Django team in the `Django <https://www.transifex.com/django/>`_
+organization at Transifex. For information on how to join, please see the `Translations
```
timgraham: The line wrapping of the existing text is at 79 characters.

README.rst
```diff
@@ -287,3 +287,78 @@ from a copy of the production database and saved to the
...
+  strings from all apps other than ``dashboard`` and ``docs``)
+
+**Important:** To keep this working properly, note that any templates for the ``dashboard`` and
+``docs`` apps **must** be placed in the ``<app name>/templates/docs/`` directory for their
+respective app, **not** in the ``djangoproject/templates/`` directory.
+
+Updating messages on Transifex
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When there are changes to the messages in the code or templates someone will
```
timgraham: someone -> a member of the translations team on Transifix ?

requirements/dev.txt
```diff
@@ -1,3 +1,4 @@
 -r common.txt
 
 watchdog==0.9.0
+transifex-client==0.13.6
```
timgraham: Not sure if pinning a version adds value rather than using the latest one. I unpinned the version for people.djangoproject.com.

dashboard/locale/fr/LC_MESSAGES/django.po
tobiasmcnulty: Silly me. I just needed to change this in the source file, and then push/pull to update in the translations.

tobiasmcnulty:
This should be ready for another review.

tobiasmcnulty:
This has been deployed!

Screenshot 2019-03-11 20 42 56

Screenshot 2019-03-11 20 43 07

PR #886, opened 2019-03-16
Added TX_LANG_MAP for docs site to fetch translations properly for zh_CN and zh_TW

tobiasmcnulty:
After deploying #885 today, I noticed that strings in the `zh-hans` (`zh_CN`) translation aren't being retrieved properly:

Screenshot 2019-03-16 12 28 46

("Search 2.1 documentation" **should** be translated.)

We can work around this by activating the language code used by Transifex and web browsers (instead of those used in `settings.LANGUAGES`), similar to what we do in https://github.com/django/django-docs-translations/blob/stable/2.1.x/manage_translations.py:

```
>>> from django.utils.translation import activate, ugettext_lazy as _
>>> activate('zh-hans')
>>> _('Search %s documentation')
'Search %s documentation'
>>> activate('zh_CN')
>>> _('Search %s documentation')
'搜索 %s 文档'
```

I opted to go this way (rather than renaming the translations after pulling from Transifex) because Chrome requests `zh-CN` rather than `zh-hans`, and non-docs apps in this repo currently rely on the `Accept-Language` header for switching languages.

### DIFF
--- a/docs/views.py
+++ b/docs/views.py
@@ -9,7 +9,9 @@ from django.core.paginator import InvalidPage, Paginator
 from django.http import Http404, JsonResponse
 from django.shortcuts import redirect, render
 from django.template.response import TemplateResponse
-from django.utils.translation import activate, ugettext_lazy as _
+from django.utils.translation import (
+    activate as django_activate, ugettext_lazy as _,
+)
 from django.views import static
 from django.views.decorators.cache import cache_page
 from django_hosts.resolvers import reverse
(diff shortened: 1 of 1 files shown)


claudep:
For docs translation, I'm simply using this line:
```lang_map = zh_CN:zh_Hans, zh_TW:zh_Hant```
in the .tx/config file.

tobiasmcnulty:
@claudep Cool, I didn't know about that option. I think we would still run into the issue in my last comment, though: The browser sends `Accept-Language: zh-CN` rather than Django's `zh-hans`. So, translations will also be missing on the non-docs pages which do not have a language switcher.

I guess once we implement #883 that may be a non-issue. But, I don't really understand why Django is using a different code than Transifex and web browsers or which would be the better option...I can see pros and cons both ways...what do you think?

claudep:
AFAIK zh-CN is deprecated for a long time now, zh-Hans/zh-Hant should be used instead. I'm a bit surprised about your Chrome assertion, are you sure?
See also https://developers.google.com/custom-search/docs/ref_languages for example.

tobiasmcnulty:
Huh. I am definitely seeing those codes in the request header:

Screenshot 2019-03-16 13 08 56

Screenshot 2019-03-16 13 09 22

I think we need a language switcher at some point anyways, so I'm okay with using `lang_map` for now. I did notice one can specify that option on a per-resource basis in `.tx/config` (working around this issue for now), but it's probably overly confusing to be using two different language codes for the same language in one project.

claudep:
That's weird, as zh-CN and zh-TW are not even in the IANA language subtag registry.
https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry

Hopefully, Chrome does not become the new IE with regards to standards :-(

tobiasmcnulty:
Hopefully it's just that way in my English-language version of Chrome or something! I put up a separate PR (#887) for the `lang_map` approach.

PR #887, opened 2019-03-16
Added lang_map in .tx/config for zh_Hans and zh_Hant

tobiasmcnulty:
After deploying #885 today, I noticed that strings in the `zh-hans` (`zh_CN`) translation aren't being retrieved properly:

Screenshot 2019-03-16 12 28 46

("Search 2.1 documentation" **should** be translated.)

We can work around this by adding a `lang_map` to `.tx/config`. Thanks to @claudep for the suggestion.

```
>>> from django.utils.translation import activate, ugettext_lazy as _
>>> activate('zh-hans')
>>> _('Search %s documentation')
'搜索 %s 文档'
```

(Original issue and conversation here: #886)


tobiasmcnulty:
Fixed:
Screenshot 2019-03-16 19 25 23

PR #954, opened 2019-09-27
Add dependencies for update_docs to run

ntj:
I needed those packages to be added to the `Dockerfile` to be able to run `docker-compose exec web python manage.py update_docs` successfully. Without `rsync` it did not create the `_built` folder necessary to be able to open the documentation pages on the website.


carltongibson (approved):
OK, thanks 👍

PR #986, opened 2020-04-06
Display header link in code snippet on hover

mdamien:
The header link for the code snippet isn't displayed even when hovering.

This PR aims to fix that.

Example, in the tutorial: https://docs.djangoproject.com/en/3.0/intro/tutorial02/#id2

image

### DIFF
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -3207,6 +3207,10 @@ ul.corporate-members li {
 }
 
 .code-block-caption, .snippet {
+	&:hover .headerlink {
+		opacity: 1;
+	}
+
     .btn-clipboard {
         float: right;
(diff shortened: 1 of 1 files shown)


carltongibson (changes requested):
Hi @mdamien — thanks for this. 

I think the idea is right, but it needs to go with the existing list of overrides for this behaviour: 

https://github.com/django/djangoproject.com/blob/b4457c0c69f526ffac2be8bd065ec47555c62c21/djangoproject/scss/_style.scss#L2055-L2068

I checked the `.code-block-caption` case. Were you able to find one for the `.snippet` class too? 

Thanks.

sabderemane:
Closing in order to make a new one with the last changes

thibaudcolas:
See #1429.

PR #989, opened 2020-05-03
Fixed snippet header-text selection visibility on Firefox.

### LINKED ISSUE #987: Code snippets header selection is not visible on Firefox
When selection the text of the header of the snippet, it works well under Chrome:

image

But under Firefox, the selection is not visible.

This is a bit problematic since the user often wants to copy-paste the filename in the header.

Example of a page where the bug can be reproduced: https://docs.djangoproject.com/en/3.0/intro/tutorial02/#id2

beingbiplov:
When you select the text on code snippet header, the selection is not visible on Firefox ie. you cannot see what is selected.
This PR aims to fix that issue.
Fix #987

### DIFF
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -3237,4 +3237,7 @@ ul.corporate-members li {
             margin-right: 10px;
         }
     }
+    .caption-text::selection{
+        background: $green-light;
+    }
 }


carltongibson (commented):
Hey @beingbiplov — Thanks for this. Sorry for the slow uptake. 

I think `::selection` is available on all supported browsers. 

https://developer.mozilla.org/en-US/docs/Web/CSS/::selection

Would you be able to update for that, and we'll see if we can get it in. Thanks

beingbiplov:
Hey @carltongibson
Changed `-moz-selection` to `::selection`. 
It should now work fine on all supported browsers. Thanks.

knyghty:
Per [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/::selection#Accessibility_concerns), it's preferred to not use `::selection` to avoid exactly these issues.

We're already using it though, and personally I'm used to the nice colours and it pleases designers, so I think we should take this and fix issues rather than removing it all completely. But I'm open to changing my mind on it.

carltongibson (approved):
Thanks for this again @beingbiplov — sorry for the slow pick-up. 🎁

PR #1034, opened 2020-10-01
Added target property and rel="noopener" to external footer link

olawanlejoel:
I was going through the documentation and wanted to check django on twitter via the link on the footer, I did that and when I got to twitter I started doing other things, When I wanted to go back to what I was reading on django I now discovered I was not on a separate Tab. I now decided to fix this.

### DIFF
--- a/djangoproject/templates/includes/footer.html
+++ b/djangoproject/templates/includes/footer.html
@@ -30,10 +30,10 @@
       <div class="col follow last-child">
         <h2>Follow Us</h2>
         <ul>
-          <li><a href="https://github.com/django">GitHub</a></li>
-          <li><a href="https://twitter.com/djangoproject">Twitter</a></li>
+          <li><a href="https://github.com/django" target="_blank" rel="noopener">GitHub</a></li>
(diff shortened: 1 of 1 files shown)


MarkusH:
Opening links in a new tab or not should be up to a user and not forced upon them.

rfleschenberg:
@olawanlejoel Just in case you didn't know it: Most browsers have an "open link in new tab" feature. It can usually be triggered by ctrl-clicking the link.

rfleschenberg:
@olawanlejoel Wether to open external links in a new browser tab or not is a somewhat controversial issue. I think many non-technical people prefer them to do so, while techies usually prefer to make the choice themselves.

The Django PR reviewers are somewhat under stress right now because there are many spam PRs due to the DigitalOcean Hacktoberfest. I think your PR was well intentioned and does not fall under that category. While I agree with @MarkusH that we do not want to merge that change, I am sure the Django community will welcome further contributions from you in the future. If you like to, check https://docs.djangoproject.com/en/dev/internals/contributing/ for an introduction! Thanks, and have a good day!

PR #1069, opened 2021-02-23
Missing translatable strings in docs template.

claudep:


claudep:
Displaying docs home in Chinese helps spotting those :-)

PR #1095, opened 2021-05-24
Remove duplicate settings from settings.docker

adamchainz:
Use the `from .common import *` pattern also used in `settings.dev` to reduce the number of lines in `settings.docker`.


bhavanasam (approved):
always raise small chunks of commits into one single pull request

adamchainz:
What do you mean?

carltongibson (approved):
Thanks @adamchainz 🌟

PR #1132, opened 2021-12-17
Added German language support

S3bastian:


pauloxnet:
@S3bastian I think you have to open a PR in the https://github.com/django/django-docs-translations and not here

PR #1136, opened 2022-01-02
Updated Dockerfile

returnvalue:
I was having a hard time during the build process when running:

docker-compose build

The build would fail, maybe related to https://github.com/nigelpoulton/psweb/issues/10

This PR incorporates the fixes


Bidaya0:
Thanks, I had the same problem here.

Bidaya0 (approved):

PR #1147, opened 2022-02-27
Upgraded Trac dump to the version 41.

tobiasmcnulty:
This fixes https://github.com/django/code.djangoproject.com/issues/93

Steps to reproduce: https://github.com/django/code.djangoproject.com/tree/docker#how-to-recreate-tracsql-after-upgrading-trac

VS Code (and possibly other tools) shows a nicer diff than GitHub.

The key changes (afaict) are:
- Upgrade of `database_version` to `41`
- Addition of missing tables (like `spamfilter_report`)

### DIFF
--- a/tracdb/trac.sql
+++ b/tracdb/trac.sql
@@ -2,24 +2,43 @@
 -- PostgreSQL database dump
 --
 
-SET statement_timeout = 0;
-SET client_encoding = 'UTF8';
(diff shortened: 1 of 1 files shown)


felixxm (approved):
@tobiasmcnulty Thanks :+1: I changed the end of lines from `^M` to `\n`. Now diff is smaller.

Issue #1064, opened 2021-01-22
Dark theme 


himanshu007-creator:
Django website is simple and good. But what about the health of developers?
As of now, there is no provision for dark theme.Having a dark theme will improve user experience.I would like to work on this issue.

Pratikshya1201:
Would you mind if I work on this issue ? I promise to submit it within a week.

himanshu007-creator:
> Would you mind if I work on this issue ? I promise to submit it within a week.

we can work in collaboration!

Pratikshya1201:
> > Would you mind if I work on this issue ? I promise to submit it within a week.
> 
> we can work in collaboration!

sure :)

himanshu007-creator:
> > > Would you mind if I work on this issue ? I promise to submit it within a week.
> > 
> > 
> > we can work in collaboration!
> 
> sure :)

got linkedin?

Pratikshya1201:
> > > > Would you mind if I work on this issue ? I promise to submit it within a week.
> > > 
> > > 
> > > we can work in collaboration!
> > 
> > 
> > sure :)
> 
> got linkedin?

yes
https://www.linkedin.com/in/pratikshya-mishra-b84b6b1a9/

Asday:
How's it going?

himanshu007-creator:
> How's it going?

Hi @Asday sir, i am having exams rn so couldn't work on it

kcoder63:
can I work on this if its still open ?

Asday:
Please do, my retinas will thank you.

yash112-lang:
Hello @himanshu007-creator, is this issue is still open. I am new to open source & want to contribute.

kunals131:
I want to work on this issue, is it still open?

Asday:
Please, if anyone is reading this - just work on it.  The issue has been open for over a year and keeps dying in the water with people abandoning it without a word and disappearing in a puff of smoke.

It's simple CSS recolouring work.  If anyone was working on it, there would be a PR open in days, if at all that long.

In the meantime, I've written this CSS userscript for docs.djangoproject.com which I have applied with the browser extension Stylus.  It will make any designer recoil in horror, but I don't care because I'm not trying to please them, I'm trying to have working retinas by the time I have children.

```
html {
    filter: invert(100%);
}
```

pauloxnet:
FYI the work on this issue is going on in the PR #1213

Asday:
[...]

PR #1158, opened 2022-04-12
Refs #1064 -- Add dark mode

sabderemane:
I have done something for the issue #1064 :)  

Looks like that :
Capture d’écran 2022-08-02 à 00 46 49

UPDATE: 
- Review accessibility colors for dark theme
- Add CSS variables for each colors to make dark / light theme in CSS
- Add switch button to toggle theme (it takes system preferences by default and override it if the button is clicked)
- Change icon switch according Django admin switch based on Furo icon theme

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,7 +15,7 @@ jobs:
 
     services:
       postgres:
-        image: postgres:12-alpine
+        image: postgres:14-alpine
         env:
           POSTGRES_PASSWORD: postgres
         options: >-
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ locale/*/LC_MESSAGES/django.mo
 .sass-cache/
 .coverage
 .tox
+djangoproject/cache
 djangoproject/static/js/lib/jquery-flot/examples
 djangoproject/static/css/*.map
 djangoproject/static/css/*.css
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,22 @@
+exclude: \/migrations\/
+default_language_version:
+  python: python3
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: "v4.3.0"
+    hooks:
+      - id: debug-statements
+      - id: detect-private-key
+      - id: file-contents-sorter
+        files: ^(requirements/\w*.txt)$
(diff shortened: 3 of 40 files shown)


pauloxnet:
Thanks for working on this issue. Can you check the accessibility of colors in the dark version? It seems that some light text on a dark background has a low contrast level.

sabderemane:
Yes @pauloxnet , I didn't check all elements for the accessibility, I will do that and update the PR

davidism:
This is cool, but having more (less?) contrast is important to readability. I find it really hard to read white on black text, but much easier if they're light and dark greys instead. That said, I strongly prefer light mode for docs, so this needs to have a switch that can override whatever is detected from OS/browser preference.

sabderemane:
Thank you!
I understand, I will add button in the header for that.

dryan:
I'm happy to help with a toggle switch if needed!

sabderemane:
Hi folks, I finally finish it 😄 
I updated the PR and the description. Happy to have feedbacks.

pauloxnet:
> Hi folks, I finally finish it smile I updated the PR and the description. Happy to have feedbacks.

Thank, it seems great.
Can you add a screenshot of the documentation search result page ?

sabderemane:
Sure @pauloxnet 
Capture d’écran 2022-04-19 à 14 12 57

pauloxnet:
Thank you @sabderemane 

I have only two personal final thoughts:
- I don't know if the word "dark mode" is redundant, I would leave only "dark" and "light" or remove it completely
- I would make sure that the toggle button is always visible when viewed on mobile, therefore outside the hamburger icon menu

sabderemane:
Thank you for the feedback @pauloxnet

I see, first of all, I put the button only, I found it weird. We don't know really what is it about except by clicking on it.

May be like that it's better, what do you think ? 
Capture d’écran 2022-04-19 à 14 37 18
Capture d’écran 2022-04-19 à 14 37 05
I have to think of other options for the button and make proposal.

> I would make sure that the toggle button is always visible when viewed on mobile, therefore outside the hamburger icon menu
Yes good point, actually it's on the menu but I will see to make it outside of the navigation.
Capture d’écran 2022-04-19 à 14 31 11

matthiask:
>  I have to think of other options for the button and make proposal.

I really like the way it's done in the Furo sphinx theme: https://pradyunsg.me/furo/ (Sorry for the borderline bikeshedding)

PR #1170, opened 2022-05-31
solved : Code snippets header selection is not visible on Firefox #987

pradipbankar0097:
image

### DIFF
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -48,10 +48,10 @@ a {
 }
 
 ::selection { // Text Selection Colors
-    background: $green-very-light;
+    background: $green-medium;
 }
 ::-moz-selection {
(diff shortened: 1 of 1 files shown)


pradipbankar0097:
Solved proof screenshot : [https://drive.google.com/file/d/1ksZRUyK12Aqt4wiTIiijjPx_Zi04dKu4/view?usp=sharing](url)

CuriousLearner (commented):
The medium color shade looks too dark for text selection. I think the problem is that the header itself has a light-green theme which is causing the selection of an item not well evident to the user.

One way, I can think is to change the background color of headers.
The other way can be to only apply medium green color on selection of header elements exclusively.

I think this would require more discussion with the accessibility team.

camilonova:
I agree. Closing this for now.

PR #1172, opened 2022-06-07
use nip.io instead of xip.io

hjlarry:
`xip.io` seems no longer service anymore, and the `nip.io`  do the same thing


cgl (approved):
Thanks!

CuriousLearner (approved):
LGTM 🌮 

Thanks for your contribution @hjlarry

PR #1176, opened 2022-07-13
Update start.html

maggyero:


djangoproject/templates/start.html
```diff
@@ -182,15 +185,16 @@ <h2>Admin</h2>
...
 
 class MemberAdmin(admin.ModelAdmin):
     """Customize the look of the auto-generated admin for the Member model"""
     list_display = ('name', 'instrument')
     list_filter = ('band',)
 
-admin.site.register(Band)  # Use the default options
-admin.site.register(Member, MemberAdmin)  # Use the customized options
+admin.site.register(models.Band)  # Use the default options
+admin.site.register(models.Member, MemberAdmin)  # Use the customized options
```
CuriousLearner, line 197: For the sake of clarity, I won't use relative imports here. Since bands is the app, it is ok to do `from bands.models import Band, Member` and then use them.

djangoproject/templates/start.html
```diff
@@ -108,6 +109,8 @@ <h2>URLs and views</h2>
       {% pygment 'python' %}
 from django.shortcuts import render
 
+from . import models
+
```
CuriousLearner, line 113: Please change this relative import to absolute import indicating the package/module name.

camilonova:
Fixed in https://github.com/django/djangoproject.com/commit/384cc161fe015c47d4c72ac13ac220cf91ffd772

PR #1201, opened 2022-09-04
Fixed #1200 -- Add more languages search support

### LINKED ISSUE #1200: Add more languages search support
Original multi-language search support was based on PostgreSQL 10, but after the P[ostgreSQL 12 update](https://github.com/django/djangoproject.com/pull/1151) we can add new supported language in the search configuration.

pauloxnet:
Part of #1031

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -23,7 +23,7 @@ services:
     depends_on:
       - db
   db:
-    image: postgres:11.2-alpine
+    image: postgres:12-alpine
     environment:
       - POSTGRES_USER=djangoproject
       - POSTGRES_PASSWORD=secret
--- a/docs/search.py
+++ b/docs/search.py
@@ -2,25 +2,32 @@ from django.contrib.postgres.search import SearchVector
 from django.db.models import F
 from django.db.models.fields.json import KeyTextTransform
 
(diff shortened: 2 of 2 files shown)


docker-compose.yml
```diff
@@ -23,7 +23,7 @@ services:
     depends_on:
       - db
   db:
-    image: postgres:11.2-alpine
+    image: postgres:12-alpine
```
CuriousLearner, line 26: Ah, good catch!

Seems like the container for Github Actions [was updated here](https://github.com/django/djangoproject.com/pull/1151), but not the docker-compose file itself.

CuriousLearner (commented):
Overall, looks good to me. Though I'm wondering if we need to test these config languages?

It seems to be [used here](https://github.com/django/djangoproject.com/blob/d0dba2554accefebcc4fed3f9c7645d8405a79e8/docs/models.py#L174), though I'm not sure if an explicit test is needed.

pauloxnet:
There is already a specific test for multilingual search using this config languages.
https://github.com/django/djangoproject.com/blob/main/docs/tests.py#L490
But I don't think it would be an added value to have a specific test for each new language we add here because it is supported by PostgreSQL.

CuriousLearner (approved):
Sounds good. Let's get this in! :)

PR #1213, opened 2022-09-24
Dark mode update

sabderemane:
Duplicate of #1158 but with clean history  
cc @knyghty


knyghty:
A few things I've found:

The theme is stored per subdomain, so you can have a different theme on the docs site, dashboard and the main site.

We don't have to fix every contrast issue here, that's for #974 / #998 but there are a few that are quite bad:

Captions for code blocks:

Screenshot 2022-09-25 at 11 21 06

Active platform switcher is more visible that the inactive one:

Screenshot 2022-09-25 at 11 25 38

Warning admonition:

Screenshot 2022-09-25 at 11 36 52

Fun example from querysets reference:

Screenshot 2022-09-25 at 11 42 40

sabderemane:
@knyghty thanks for the review, I will try with the cookie for the subdomains issue as we said, and make the changes requested

sabderemane:
@knyghty I have made changes, tell me what you think when you have some time 😄

pauloxnet:
I tried locally your branch, it seems working.
I found only 2 issues that I'll write below.

pauloxnet:
Changing between page in light mode it show for few milliseconds the dark mode, I attach the screencast.
[dark-mode-glitch.webm](https://user-images.githubusercontent.com/521097/193240287-0b0029fc-c0fe-472b-9704-c63e1b0c7dd8.webm)

pauloxnet:
I can't see the green button in the home page if in dark mode, but it works in light mode.

dark-mode

light-mode

sabderemane:
> Changing between page in light mode it show for few milliseconds the dark mode, I attach the screencast.
> 
> [dark-mode-glitch webm](https://user-images.githubusercontent.com/521097/193240287-0b0029fc-c0fe-472b-9704-c63e1b0c7dd8.webm)
> 

Are you with dark OS preference? Because when you force in light mode with dark OS preference or the reverse, there is a little time for the update because we are taking the os preference and if there is a theme preference it's loaded after.

I have to see if I can improve that with css loading but I'm not sure.

pauloxnet:
> Are you with dark OS preference? Because when you force in light mode with dark OS preference or the reverse, there is a little time for the update because we are taking the os preference and if there is a theme preference it's loaded after.
> 
> I have to see if I can improve that with css loading but I'm not sure.

Yes, in fact in my case my OS was in dark mode and I set it to light mode on the website.

With the same OS settings I see that other open source documentation themes (es: material for mkdocs, furo for sphinx)  do not have that problem, could we take inspiration from them?

- https://squidfunk.github.io/mkdocs-material/getting-started/
- https://pradyunsg.me/furo/quickstart/

sabderemane:
I already take the inspiration from furo, but may be linked to our style because the most annoying is the background by default. 

I will double check that.

pauloxnet:
@sabderemane I've seen the new commits and trying them locally it seems to me that everything works fine. Thanks for your work.

The only thing that's not working for me locally is the release schedule image which is illegible.

release-schedule-broken-min

I saw in the code that in the dark version of the site the image would be inverted, I experimented it by hand locally but the colors are too different from the original for me.

release-schedule-invert-min

I would rather generate the image from the SVG file with a white background, without applying the inverse filter.

release-schedule-white-min

P.S. For the future we could generate a special image for the dark theme or use an SVG in which to apply CSS styles directly.

knyghty:
@pauloxnet we discussed these options at the sprints with Carlton and couldn't find any better solution immediately.

Adding another image adds burden to the person updating the image, which happens somewhat regularly.

Styling the SVG means inlining it, making it harder to change but also a bit of a mess in the code as it's quite large.

I would like a better solution as well but I'm not sure if there's one readily available, we might need to think how we can generate the chart in a different way...

pauloxnet:
> @pauloxnet we discussed these options at the sprints with Carlton and couldn't find any better solution immediately.

Thanks for the clarification.

But in local on my pc the "inverse" filter on the image is still not working, but I don't know if it's a bug or if I'm doing something wrong
release-schedule-broken-min

The simplest solution seems to me to always have the image with a white background, and this can be achieved by specifying the background color of the image in the CSS
release-schedule-white-min

However if the inverse filter is working for you and you already decided that this is the best solution we can merge
release-schedule-invert-min

sabderemane:
@pauloxnet thank you for your review!  It's not supposed to happen for the image, but I may have missed a case. 
Did you run `make compile-scss`? Do you have the steps to reproduce it on my side ?

pauloxnet:
> @pauloxnet thank you for your review! It's not supposed to happen for the image, but I may have missed a case. Did you run `make compile-scss`? Do you have the steps to reproduce it on my side ?

My steps are:
0. `git pull`
1. `make compile-scss`
2. `make collectstatic`
3. `make run`

I also tried `make compile-scss-debug` .

But if you say it works for you in local it's enough for me

camilonova (approved):

camilonova:
@sabderemane Thank you, great work!

pauloxnet:
@sabderemane thanks for your work here 🙏

PR #1218, opened 2022-10-01
Edit sharpen skills section

aaditkamat:
I'm not sure if you all agree with this change, but I think that the word `more` was missing at the end of the aside wrapped within the round brackets.


carltongibson (commented):
I think "and then some" is a standard idiom. (The more is implicit)

Issue #1240, opened 2022-10-14
Text too light to read on the documentation pages' search bar in Light Mode.


NerdvanaNC:
### The Issue:

Using Light mode on the docs page makes the search bar's text color extremely light and difficult to read.

### Steps to recreate:

Visit https://docs.djangoproject.com/ and try typing anything into the search bar.

### Potential Fix:

The cause is setting the search bar's ```color``` property to ```color: var(--hairline-color);``` - changing this to ```color: var(--body-fg);``` instead will fix the issue.

pauloxnet:
Thanks for the report, can you open a PR with your potential fix and attach o it a screenshot ?

NerdvanaNC:
@pauloxnet I've created a pull request fixing this issue. :)

pauloxnet:
> @pauloxnet I've created a pull request fixing this issue. :)

Thanks. I reopened this issue because is not her fixed, please leave open until a PR will fix it.

Can you attach two screenshots that show your fix in action with light and dark mode?

NerdvanaNC:
Oh, sorry - I didn't realize I shouldn't mark this as fixed. I've attached a total of 4 screenshots here - each of them shows the problem in both Light/Dark mode and the fix in both Light/Dark mode.

_The issue only exists in Light Mode, but I've attached Dark mode screenshots too showing that my fix works in both modes._

![djangoDarkModeAfterFix](https://user-images.githubusercontent.com/13850325/195976284-6907080e-71f0-40ef-a8cc-c462226c5866.jpg)
![djangoDarkModeBeforeFix](https://user-images.githubusercontent.com/13850325/195976285-a265055e-cb90-4992-9977-f486eabad807.jpg)
![djangoLightModeAfterFix](https://user-images.githubusercontent.com/13850325/195976286-67588187-e1c1-4bc9-bd6c-b96c25be18d3.jpg)
![djangoLightModeBeforeFix](https://user-images.githubusercontent.com/13850325/195976287-d556c4ce-a196-4be5-bcb7-6f4bea6b90f9.jpg)

pauloxnet:
Thanks. I'm going to merge your PR because right now the input search is not usable and we want to fix that.
But maybe in the future we can reconsider colors, fonts and design of the site.

sabderemane:
Can be closed now I think

PR #1247, opened 2022-10-15
Refs #1240 -- Fixed search bar text visibility in light mode

sabderemane:
Related to #1240 

This is a duplicate of PR #1241 with some improvements
cc @camilonova


camilonova:
Thank you!

PR #1249, opened 2022-10-16
Add Open Graph tags (#1144)

marksweb:
This adds Open Graph tags to address #1144.

The approach is to keep it fairly simple, with blocks where possible in the base template to allow overrides where necessary in other templates.


marksweb:
:wave: @pauloxnet @carltongibson 

Thought I'd get OG tags integrated & before going too far, I thought it'd be worth showing you guys the general approach I was taking incase you had your own thoughts on the matter.

So far I've gone with the basic few tags for `og` and `twitter`, using `block` tags to allow specific pages to overwrite content, like the download page using the image of the roadmap which I think would be useful.

nemecek-filip:
Hi!

Twitter also has support for OG tags so that is possibly something to consider - could be worth it - considering you then need to maintain fewer tags - https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/markup

Also the image one seem to need full URL otherwise they often fail to display.

marksweb:
@nemecek-filip thanks I didn't realise that about the twitter tags.

And as I understand it, the project uses an absolute url for the static tag in production. At least that's what it looks like comparing the production source to the production static url setting. It's on a sub domain of static.djangoproject.com

nemecek-filip:
@marksweb oh I see. Apologies for the false alarm! Did not consider that the paths could be absolute by default.

marksweb:
@nemecek-filip not to worry, it's rare to see static setup this way. The vast majority of projects stick to `/static` like the default sets us up for

marksweb:
@pauloxnet No need to apologise. 

Understood 👍

djangoproject/templates/overview.html
```diff
@@ -1,7 +1,8 @@
 {% extends 'base.html' %}
 
-{% block title %}Django overview{% endblock %}
 {% block layout_class %}full-width{% endblock %}
+{% block title %}Django overview{% endblock %}
+{% block og_title %}Django overview{% endblock %}
```
pauloxnet, line 5: Moving existing code as little as possible helps reviewers
```suggestion
{% block title %}Django overview{% endblock %}
{% block layout_class %}full-width{% endblock %}
{% block og_title %}Django overview{% endblock %}
```

.gitignore
```diff
@@ -1,3 +1,7 @@
+# If you need to exclude files such as those generated by an IDE, use
```
pauloxnet, line 1: This change seems out of scope for this PR.

djangoproject/templates/releases/download.html
```diff
@@ -2,8 +2,13 @@
...
 
 {% block sectionid %}download{% endblock %}
-{% block title %}Download Django{% endblock %}
 {% block layout_class %}sidebar-right{% endblock %}
+{% block title %}Download Django{% endblock %}
+
+{% block og_title %}Download Django{% endblock %}
+{% block og_image %}{% static "img/release-roadmap.png" %}{% endblock %}
+{% block og_image_alt %}Django's release roadmap{% endblock %}
+{% block og_description %}The latest official version is {{ current.version }}{% if current.is_lts %} (LTS){% endif %}{% endblock %}
```
pauloxnet, line 11: See above
```suggestion
{% block title %}Download Django{% endblock %}
{% block layout_class %}sidebar-right{% endblock %}
{% block og_title %}Download Django{% endblock %}
{% block og_image %}{% static "img/release-roadmap.png" %}{% endblock %}
{% block og_image_alt %}Django's release roadmap{% endblock %}
{% block og_description %}The latest official version is {{ current.version }}{% if current.is_lts %} (LTS){% endif %}{% endblock %}
```

.gitignore
marksweb: While you're absolutely right, this is one of those things that would help new contributors & otherwise likely wouldn't get added.

pauloxnet (approved):

pauloxnet:
@marksweb your PR is ok for me. One last question: have you check locally with some Open Graph reader or checker (e.g. [share preview](https://github.com/rafaelmardojai/share-preview/)) ?

pauloxnet:
I'm going to verify locally all the pages to check if all OG tags are ok.

marksweb:
@pauloxnet I did some testing of the markup in another project to verify. However I'm not aware of a tool for testing tags locally on macOS.

pauloxnet (changes requested):
I've tried all the pages locally with a Open Graph viewer and I've added some change suggestions.

marksweb:
Thanks @pauloxnet - all changes committed.

pauloxnet (approved):

pauloxnet:
> Thanks @pauloxnet - all changes committed.

That was fast. Thanks @marksweb for your workj and patience. I've just merged.

PR #1250, opened 2022-10-16
Feature/fixed textarea styles in dark mode

nemecek-filip:
After the recent dark mode introduction some `textarea` elements with `.codedump` class have black text making it super hard to read.


pauloxnet:
@nemecek-filip thanks for this PR, can you provide screenshots before and after your fix in light and dark mode ?

camilonova:
Thank you

PR #1254, opened 2022-10-18
Fix #1253 : Side-Scrollbar on the website breaks the focus while reading

### LINKED ISSUE #1253: Side-Scrollbar on the website breaks the focus while reading
### Issue

The broad sidebar on the website looks extremely annoying , grabbing the user's attention and breaking the focus while reading the website contents. The sidebar should either be removed completely or minimized so that it doesent grab the user's attention and dosent break the focus ,  while mantaining the ability to scroll.As of now , the issue exists only on chrome and edge browsers.

### Screenshots
Chrome : 

image

Edge :

image

### Proposed Fix

So , the major problem with hiding side-scrollbars is to retain the ability to scroll. For hiding the scrollbars , `overflow: hidden` can be added to the css but that disables the scrollability of the webpage.The better approach would be to use `-webkit-scrollbar` pseudo selector for webkit browsers like chrome and edge.Just adding the `-webkit-scrollbar` pseudo selector in `djangoproject\scss\_style.scss` would fix the issue.

### Screenshots ( After the Fix )

Chrome:

image

Edge:

image

RyanWalker277:
Hello mantainers , this is a fix to the issue #1253. I have added the `-webkit-scrollbar` pseudo selectors to hide the side scrollbar in webkit based browsers.

### DIFF
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -21,6 +21,10 @@ body {
     -webkit-font-smoothing: antialiased;
     -moz-osx-font-smoothing: grayscale;
 
+    &::-webkit-scrollbar {
+        display: none;// for hiding the scrollbar on chromium
+    }
+
     .layout-secondary,
     [role="complementary"] {
         @include sans-serif; // reverse font style on sidebar and secondary areas


logankilpatrick:
Pending feedback in the original issue linked above. The question raised is if we want to hide the scrollbar on chrome.

pauloxnet:
@RyanWalker277 can you also provide a screenshot of your PR ?

RyanWalker277:
I have attached the screenshots in the issue itself, after adding the pseudo selector , the bulgy sidebar isn't visible anymore which looks cleaner than current sidebar. Should I add the screenshots with the PR?

pauloxnet:
> I have attached the screenshots in the issue itself, after adding the pseudo selector , the bulgy sidebar isn't visible anymore which looks cleaner than current sidebar. Should I add the screenshots with the PR?
No thanks the screenshot in the issue are enough

sabderemane:
I'm closing it, because the scrollbar can be hidden in the operating system settings, as this modification should not be made for accessibility reasons.

PR #1256, opened 2022-10-20
Add a link to setting up a virtual python env

logankilpatrick:
This a small update in the hopes of making the getting started docs more accessible. Python Env's are hard for me so I hope adding this link will make people's lives easlier.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -9,7 +9,7 @@ djangoproject.com source code
 
 To run locally, do the usual:
 
-#. Create a Python 3.8+ virtualenv and activate it
+#. Create a `Python 3.8+ virtualenv and activate it`<https://docs.python.org/3/library/venv.html>_
 
 #. Install dependencies::
 


README.rst
```diff
@@ -9,7 +9,7 @@ djangoproject.com source code
 
 To run locally, do the usual:
 
-#. Create a Python 3.8+ virtualenv and activate it
+#. Create a `Python 3.8+ virtualenv and activate it`<https://docs.python.org/3/library/venv.html>_
```
CuriousLearner, line 12: This looks good! Maybe we can also add the commands to create/activate venv like:

```
$ python3 -m venv ~/.virtualenvs/djangoprojectenv
$ source ~/.virtualenvs/djangoprojectenv/bin/activate
```

I think this would make it easier for people who are new to Python. Let me know what are your thoughts.

CuriousLearner (approved):

README.rst
logankilpatrick: I am happy to do so but I also think it is sufficient to provide a link for folks to use. Either way works.

pauloxnet: @CuriousLearner I personally use the exact command you proposed but I would prefer to add only the link and let people to use their ways to use the venv/virtualenv as @logankilpatrick proposed

pauloxnet (approved):

PR #1262, opened 2022-10-27
Fix N+1 Query on Django Community page

saadmk11:
closes https://github.com/django/djangoproject.com/issues/1261


pauloxnet:
It can be improved to move all the logic in the database but for now it closes the N+1 bug.

pauloxnet (approved):

PR #1264, opened 2022-10-29
Fix N+1 Query on Feed List Page

saadmk11:
closes https://github.com/django/djangoproject.com/issues/1263


carltongibson (commented):
Can we add a test case using assertNumQueries to make sure we don't regress here?

pauloxnet (approved):

PR #1268, opened 2022-11-13
Added note to gitignore to explain exclude configuration.

marksweb:
This adds the same note that exists in the django project itself, to explain to users how to make use of exclude functionality in git.


CuriousLearner (approved):
Thanks!

PR #1277, opened 2022-11-26
Fix: Add overflow-wrap: break-word to [role='main']

### LINKED ISSUE #729: Responsive break in https://docs.djangoproject.com/en/1.10/intro/tutorial05/
The design breaks in mobile view for the below page. The bottom language & documentation version causes horizontal scroll in this page. I used Google chrome on my Redmi note 3 device.

screenshot_2016-12-22-21-20-17-084_com android chrome

You can see the pagelink here : ​https://docs.djangoproject.com/en/1.10/intro/tutorial05/

ghost:
### Type of change

SCSS file change

### Specific Code Changes
Added the overflow-wrap property to `[role='main']` with the value of `'break-word'`. 

### Objective of Change
This prevents any text inside a `[role='main']` div from causing the page to overflow on the x-axis.

Resolves #729

### Media
Before:
image

After:
image

### DIFF
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -337,6 +337,7 @@ blockquote {
 [role="complementary"] {
     //Sidebar in column layouts, gray section in fullwidth layout
     padding: 0 10px 20px;
+    overflow-wrap: break-word;
     @include font-size(14);
     @include respond-min(768px) {
         float: right;


djangoproject/scss/_style.scss
```diff
@@ -256,6 +256,7 @@ blockquote {
     //Main column. Left aligned by default. See /styleguide for explanation.
     background: var(--white-color);
     padding: 1px 10px 40px;
+    overflow-wrap: break-word;
```
sabderemane, line 259: It's also needed for `[role="complementary"]`, if you go [here](https://docs.djangoproject.com/en/3.0/ref/templates/api/) you will be able to scroll horizontally and still have the space.

PR #1305, opened 2023-01-23
issue#835 github login issue. common.py updated

aviralshastri:
github login issue. common.py updated also there is no variable for github secrets


logicwithharshal:
I noticed PR #1305 was closed without merging because it didn’t solve the problem. I’d like to work on a proper fix using {% url %} instead of hardcoded URLs and test login across subdomains.

PR #1320, opened 2023-03-02
Updated global navigation bar.

AnuragD123:
Set the position of global navigation bar to 'sticky' for a more appealing user experience.


thibaudcolas:
Hi @AnuragD123, I don’t see how this is more appealing, it just means there’s less visible page in the scrollable area.

PR #1321, opened 2023-03-04
Update tox.ini to use allowlist_externals

cgl:
Prior to this change, we were using the whitelist_externals setting.

This change updates the tox.ini to use allowlist_externals instead of whitelist_externals, as whitelist_externals is now deprecated[0].

[0]: https://tox.readthedocs.io/en/latest/config.html


CuriousLearner (approved):
LGTM 🌮 

Thank you for your contribution!

PR #1348, opened 2023-04-16
Add permalinks to the headers in teams page

cgl:
Prior to this change, we couldn't link sub-headers in the teams page.

This change adds permalink urls to each h3 header in the page. So that we can link each team section separately.

I used the `plink` class from https://github.com/django/djangoproject.com/blob/main/djangoproject/templates/conduct/faq.html#L18.


pauloxnet (approved):
LGTM

PR #1353, opened 2023-04-28
Fixed Overflow Issue

Tushar-sahani:
Hello maintainer, this is Fixed for overflow issue of text in the Documentation.


Tushar-sahani:
I have attached the screenshots and video in the issue itself, after adding the overflow-wrap property . Should I add the screenshots with the PR ?

sabderemane:
> I have attached the screenshots and video in the issue itself, after adding the overflow-wrap property . Should I add the screenshots with the PR ?

Yes please, better to have them in the PR related, I will close the issue in order to keep the older one as the current issue linked.

Tushar-sahani:
In most of the Documentation Have this problem where ```span.pre``` and ```tt``` these two tags are used.

## Before



https://github.com/django/djangoproject.com/assets/104075614/63560f93-f373-4cc9-b3c0-55472dad1f7f

## After

fixed1
Screenshot (9)

sabderemane (approved):

djangoproject/scss/_style.scss
```diff
@@ -20,6 +20,7 @@ body {
     line-height: 1.6;
     -webkit-font-smoothing: antialiased;
     -moz-osx-font-smoothing: grayscale;
+    overflow-wrap: break-word;
```
sabderemane, line 23: After some discussion, I think it would be better to have the `overflow-wrap` only on the bloc needed to avoid problems in places we didn't expect.

sabderemane (changes requested):

sabderemane (approved):
I've done a small fix to your PR, thank you for your contribution @Tushar-sahani 🎉 

Welcome on board ⛵

PR #1357, opened 2023-05-03
Add djhtml to pre-commit config

cgl:
Prior to this change, we didn't have a html autoformatter in
pre-commit hooks for django templates.

This change adds djhtml to the precommit hooks as a formatter for html
files.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -42,3 +42,8 @@ repos:
       rev: "6.0.0"
       hooks:
           - id: flake8
+    - repo: https://github.com/rtts/djhtml
+      rev: "3.0.6"
+      hooks:
+          - id: djhtml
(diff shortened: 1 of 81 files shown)


.pre-commit-config.yaml
```diff
@@ -26,7 +26,7 @@ repos:
           - id: django-upgrade
             args: [--target-version, "3.2"]
     - repo: https://github.com/psf/black
-      rev: 23.1.0
+      rev: 23.3.0
       hooks:
           - id: black
     - repo: https://github.com/pre-commit/mirrors-prettier
```
cgl, line 32: I didn't upgrade prettier as 3.0.0 is still alpha and maybe it is best for us to stay with stable releases.

knyghty (approved):

sabderemane (approved):

PR #1376, opened 2023-06-19
Refs #1377 -- Added diamond and platinum members to the download page.

nessita:
Part of #1377.

### DIFF
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -3511,6 +3511,33 @@ ul.corporate-members li {
 	padding: 0 0 10px 10px;
 }
 
+/* Corporate membership download page, this mimics .fundraising-sidebar */
+
+.corporate-members {
+	.member-logo {
+		margin-top: 20px;
+		width: 20%;
+		float: left;
+		margin-bottom: 20px;
+
+		img {
+			width: 100%;
+			max-width: 75px;
+		}
+	}
+
+	.small-cta {
+		width: 70%;
+		float: left;
+		vertical-align: top;
+		margin-left: 5%;
+		margin-right: 5%;
+		margin-bottom: 20px;
+	}
+
+	clear: both;
+}
+
 /* Messages */
 
 .messages {
--- a/djangoproject/templates/releases/download.html
+++ b/djangoproject/templates/releases/download.html
@@ -253,6 +253,29 @@
   <h1 class="visuallyhidden">Additional information</h1>
   <div role="complementary">
 
+    {% if corporate_members %}
+      <div class="corporate-members">
+        <h2>Diamond and Platinum Members</h2>
+        {% for obj in corporate_members %}
+          <div class="clearfix">
+            <div class="member-logo">
+              <a href="{{ obj.url }}" title="{{ obj.display_name }}">
+                <img src="{{ obj.thumbnail.url }}" alt="{{ obj.display_name }}" />
+              </a>
+            </div>
+            <div class="small-cta">
+              <ul class="list-links-small">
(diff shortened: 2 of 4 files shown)


nessita:
Screenshot of proposed changes:
image

sabderemane:
Hi @nessita, can you show how does it look with the dark mode ?

nessita:
> Hi @nessita, can you show how does it look with the dark mode ?

Yes, sorry! Here it is (in all honesty, I did nothing special for the dark more... shall I?):

image

sabderemane:
> Yes, sorry! Here it is (in all honesty, I did nothing special for the dark more... shall I?):

No worries, no seems to be ok, it's just to make sure it's not broken or not readable 🙂

nessita:
I created a longer list of (fake) sponsors and noticed an issue that I have now sorted out. This is how it looks:
image

.pre-commit-config.yaml
```diff
@@ -29,6 +29,7 @@ repos:
       rev: 23.3.0
       hooks:
           - id: black
+            args: ["--line-length", "88"]
```
sabderemane, line 32: Why did you change this ?

djangoproject/scss/_style.scss
```diff
@@ -3511,6 +3511,33 @@ ul.corporate-members li {
...
+	}
+
+	.small-cta {
+		width: 70%;
+		float: left;
+		vertical-align: top;
+		margin-left: 5%;
+		margin-right: 5%;
+		margin-bottom: 20px;
+	}
```
sabderemane, line 3536: This might look small in mobile, do you have a screenshot to see if it's not the case?

sabderemane (commented):
LGTM, just few questions 😄

.pre-commit-config.yaml
nessita: So, I was having resolution conflicts between black and isort. The former would wrap imports at 79 cols and the latter would expand them. So I was stuck in a loop looking like this, and the commit operation failed every time:

```
black....................................................................Failed
- hook id: black
- files were modified by this hook

reformatted releases/tests.py

All done! ✨ 🍰 ✨
1 file reformatted.

isort....................................................................Failed
- hook id: isort
- files were modified by this hook

Fixing /home/nessita/fellowship/djangoproject.com/releases/tests.py
```

The proposed solution reconciles isort and black regarding line length. If there is a better solution for this, I'm happy to change this! I could also prepare a separated PR if you prefer. What do you think?

djangoproject/scss/_style.scss
nessita: Good point, Let me get that screenshot (I'll upload it ASAP). Though this class is the same as the `small-cta` used next to the "donation heart in the sidebar", so if we need to change this, we may want to change them both.

nessita:
Attaching mobile screenshot:
image

.pre-commit-config.yaml
sabderemane: I understand but I think it's better to have it on a separated PR and ask review from djangoproject maintainers to make sure we all agree on this change 🙂

djangoproject/scss/_style.scss
sabderemane: Hmm, I think it's fine for now, we'll see how to improve it later since there's already an ongoing project to redesign the site.

.pre-commit-config.yaml
nessita: Makes sense. I have reverted and I was able to commit, so before I must've been mistaken or something else was going one :shrug: 

Thanks for your time!

sabderemane (approved):

Issue #1366, opened 2023-05-23
Make right sidebar independent from main content


adiletto64:
## Feature request

Make right sidebar scrollable independently or static like [this docs](https://vuejs.org/guide/introduction.html)
![image](https://github.com/django/djangoproject.com/assets/72970300/93f7a776-aea3-4700-b7a0-8ea9e18a1770)

Jashwanthkanderi:
Hey,
Can I Work on this

pauloxnet:
We just talked about this issue with @knyghty and @sabderemane at DjangoCon Europe 2023.
A side menu, such as those found in other Python project documentation, may be implementable.
- https://www.psycopg.org/psycopg3/docs/
- https://fastapi.tiangolo.com/lo/
- https://v2.picocss.com/docs/nav#vertical-stacking

vidit-od:
Hello, fellow community members!

I am Vidit from India and I have been using Django for quite some time now. I absolutely love this framework and I'm eager to start my open-source journey. However, I'm a bit confused about how to begin, even though it feels like something I have done before. I am eager to contribute to this issue as this seems something i have done before.

omgupta2202:
Hello @pauloxnet  , I would love to work on this! Can I be assigned? Thanks!

pauloxnet:
@omgupta2202 there's no need to assign the issue. If you want to work on it you can open a PR and request a review.

omgupta2202:
Hi @pauloxnet , After Running Locally with Docker, I encountered an issue- On clicking the documentation button. When accessing the URL "http://docs.localhost:8000/en/4.2/", I'm getting a "Page not found (404)" error.

<img width="1437" alt="Screenshot 2023-06-14 at 6 15 57 PM" src="https://github.com/django/djangoproject.com/assets/66220328/8675b63e-b115-4e8c-9a0b-6bf0bc7db154">
<img width="641" alt="Screenshot 2023-06-14 at 6 15 24 PM" src="https://github.com/django/djangoproject.com/assets/66220328/82310723-0967-4154-a6ad-652d5e75e206">

Could you please investigate this issue further and provide guidance on resolving it? Let me know if you need any additional information or if there are any steps I can take to help debug the problem.

omgupta2202:
@pauloxnet i have followed all readme steps to setup with docker, yet i am getting this error.
Should i open the issue for the same?

pauloxnet:
Yes, please.

niranjank2022:
[...]

PR #1379, opened 2023-06-19
Feature #1366 : make right sidebar independent from main content

implicitdefcncdragneel:
Issue : #1366 
issue

Hello maintainers, I have updated the scrollable functionality to **sidebar-right** .

### DIFF
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -373,6 +373,8 @@ blockquote {
 		.sidebar-right & {
 			margin-left: 3%;
 			margin-right: 0;
+			overflow-y: auto;
+			height: 100vh;
 		}
 	}
 
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -373,8 +373,6 @@ blockquote {
 		.sidebar-right & {
 			margin-left: 3%;
 			margin-right: 0;
-			overflow-y: auto;
-			height: 100vh;
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


djangoproject/scss/_style.scss
```diff
@@ -436,13 +436,22 @@ blockquote {
...
 		}
+
+		div[role="main"],
+		div[role="complementary"] {
+			max-height: 100vh;
+			overflow-y: scroll;
+		}
+		& ::-webkit-scrollbar {
+			width: 0;
+		}
```
sabderemane, line 454: Why did you remove the width of the scrollbar here ?

implicitdefcncdragneel: Hey @sabderemane,
Setting the width to 0 hides the default scrollbar, as the UI is not that good for the default one.

djangoproject/scss/_style.scss
```diff
@@ -436,13 +436,22 @@ blockquote {
 	@include respond-min(768px) {
 		margin: 0 auto;
 		max-width: 1400px;
-		padding: 0 (32/768) * 100%;
+		padding: 0 calc(32 / 768 * 100%);
```
thibaudcolas, line 439: This produces the same result with Sass and vanilla CSS computation so I don’t understand this change?

djangoproject/scss/_style.scss
thibaudcolas: We should never hide the scrollbar. That’s how users can tell the area is scrollable to start with, and lots of people interact with the scrollbar to scroll.

thibaudcolas (changes requested):
This doesn’t look quite right to me. The pattern we want here has different characteristics:

- We’d want two scrollable areas – the sidebar, and the rest of the page. This has 3 areas as-is: sidebar, main area, rest of the page.
- We’d also want the sidebar to be sticky I believe.
- And I’m not convinced this is something it makes sense to implement site-wide (at least with the current site design & content model). For documentation sites where the sidebar is only navigation, it makes sense. In the case of Django we also sometimes have regular content in the sidebar.

All in all I think this is the type of large design change it’d make more sense to consider as part of the djangoproject.com redesign (for reference: [survey form](https://docs.google.com/forms/d/e/1FAIpQLSffct3GloikvqzDPDc-IR3phuVMfVl12LoDyoJorIzZGOSV7Q/viewform), I don’t have a better ref).

Issue #1234, opened 2022-10-11
Menu button is not a button


xi:
The menu button for the mobile navigation is not accessible by keyboard because it is a `<div>` instead of a `<button>`

pauloxnet:
How can your reporting of yours be verified in concrete? How would you change the button?

RyanWalker277:
@pauloxnet The menu button is indeed a div , yet it works perfectly , I faced no issue as the button not being clickable .I tested the website on multiple mobile browsers but didn't find any such problem.Attaching screenshot for refrence.


![image](https://user-images.githubusercontent.com/32684077/196188942-39737369-6442-4898-9cac-9790a3e6875d.png)

xi:
Let me be more specific about the steps to reproduce the issue:

Imagine you are a user with bad sight and bad hand-eye coordination. So you zoom in 400% and use the keyboard to navigate.

Expected experience: I can use the tab key to get to the menu button, and use the space key to open the menu.

What happens instead: The menu button is skipped. Instead the focus moves directly to the (invisible) menu.

amit9838:
Fixed,  see PR  #1306

aviralshastri:
![● mobile-menu js - djangoproject com - Visual Studio Code 24-01-2023 11_47_43](https://user-images.githubusercontent.com/104085688/214224842-3869865e-40c3-45a2-89b8-87ee2b71a1af.jpg)

amit9838:
Ok so, do you want me to cancel my PR? right?

Charan-Deep:
Please Assign me this issue.

YashGodwal:
![Screenshot (75)](https://user-images.githubusercontent.com/96742430/229338314-3e18ab13-63d1-443c-964f-6a973f68187e.png)
A "menu button" can refer to a specific type of button that is used to access a navigation menu on a website. This type of button is typically represented by an icon with three horizontal lines, often referred to as a "hamburger menu" icon.

So, in this context, the statement "menu button is button in website" would be correct. A menu button on a website is a type of button that is used to access a navigation menu, and it is an important element of website design and user experience.

kaushal-malpure:
@pauloxnet Can I also proceed with this issue?

pauloxnet:
@kaushal-malpure feel free to work on this, no need to be assigned to work on an issue :)

kaushal-malpure:
Hey @pauloxnet , 
As per issue raised menu button div tag changed to button tag (PR#1402).
[...]

PR #1400, opened 2023-08-05
Fixed Issue Menu button is not a button #1234

faisal1025:
The issue has been fixed the menu button is now focusable on pressing `tab` from the keyboard.
The menu button is focusable from the keyboard by pressing the Tab button. 
Screenshot (1)
We can open the options from the keyboard by pressing Enter button
Screenshot (2)

### DIFF
--- a/djangoproject/static/js/mod/mobile-menu.js
+++ b/djangoproject/static/js/mod/mobile-menu.js
@@ -12,7 +12,7 @@ define([
         init: function(){
             var self = this;
             self.menu.addClass('nav-menu-on');
-            self.button = $('<div class="menu-button"><i class="icon icon-reorder"></i><span>Menu</span></div>');
+            self.button = $('<div class="menu-button"><button class="icon icon-reorder" style="background-color: transparent; color: white; border: none;"></button><span>Menu</span></div>');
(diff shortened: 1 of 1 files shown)


djangoproject/static/js/mod/mobile-menu.js
```diff
@@ -12,7 +12,7 @@ define([
         init: function(){
             var self = this;
             self.menu.addClass('nav-menu-on');
-            self.button = $('<div class="menu-button"><i class="icon icon-reorder"></i><span>Menu</span></div>');
+            self.button = $('<div class="menu-button"><button class="icon icon-reorder" style="background-color: transparent; color: white; border: none;"></button><span>Menu</span></div>');
```
sabderemane, line 15: style should be on the css file corresponding to the element (`_style.scss`) and not inline like this

sabderemane (changes requested):
Hi @faisal1025 , thank you for your contribution! ✨ 
Can you also provide screenshots before and after your changes ? 
I feel like the style you applied it's not taken in consideration as there are a border around the menu with is not the case currently.

sabderemane:
Closed by #1418

PR #1408, opened 2023-09-21
fix: Install g++ to fix libsass build issue

### LINKED ISSUE #1407: libsass issue building docker setup
Just trying to get setup via docker and I'm getting an issue building `libsass`

```
22.01       gcc: fatal error: cannot execute ‘cc1plus’: execvp: No such file or directory
22.01       compilation terminated.
22.01       error: command 'gcc' failed with exit status 1
22.01       [end of output]
22.01   
22.01   note: This error originates from a subprocess, and is likely not a problem with pip.
22.01 error: legacy-install-failure
22.01 
22.01 × Encountered error while trying to install package.
22.01 ╰─> libsass
22.01 
22.01 note: This is an issue with the package mentioned above, not pip.
22.01 hint: See above for output from the failure.
```

My host is macOS 13.5.2

Docker 24.0.6
Compose 2.21.0

marksweb:
This fixes #1407 and with g++ libsass can be installed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,6 +24,7 @@ RUN apt-get update \
 COPY ./requirements ./requirements
 RUN apt-get update \
     && apt-get install --assume-yes --no-install-recommends \
+        g++ \
         gcc \
         libc6-dev \
         libpq-dev \


sabderemane (approved):
Thank you so much @marksweb for the fix 🌟

PR #1417, opened 2023-10-18
Fixed libsass link in readme

alexgmin:
At some point in this year (based on the snapshots in web.archive.org) the http://libsass.org domain stopped rediredirecting to https://sass-lang.com/libsass and became some type of gambling related website.


sabderemane (approved):
Thank you @alexgmin for the patch! Welcome on board ⛵

PR #1367, opened 2023-06-01
Adds a "Skip to main content" link.

devanubis:
- Adds link at top of page with styles to appear when focused
- Changes role="main" to a <main> and adds an id to link to
- Moves 'copy-banner' and 'billboard' inside <main> so they are not bypassed when skipping the nav

PR #1418, opened 2023-10-20
Fix #1409 -- Add Keyboard Accessibility To Hamburger Menu

### LINKED ISSUE #1409: BLOCKER - The "hamburger" menu toggle is impossible to use with a keyboard
> Blocker: Critical issue. Certain users cannot perceive or operate this, with no workaround available.

This makes it impossible for keyboard and screen reader users to access menu items and navigate the site in this way. (https://github.com/django/djangoproject.com/issues/1234, [18:18](https://www.youtube.com/watch?v=dAiMPTOMh1U&t=1098s), [SC 2.1.1 Keyboard](https://www.w3.org/TR/WCAG22/#keyboard)).

ontowhee:
Fix for #1409 

This changes the menu button from a `div` to a `button` element so the keyboard is able to tab to it. The border is set to 0 to override the default button border style.

Before - It wasn't able to tab to the menu button
Screenshot from 2023-10-20 06-33-11

After - It is able to tab to the menu button with the red border around it
Screenshot from 2023-10-20 08-11-08

### Other notes
- [ ] ~When you tab to the menu it should be obvious you're on it (look how the dark mode button works), probably using css :focus selector~ It already shows a red outline when you tab to it.
- [ ] ~When you hit space it should activate, again like the dark mode toggle~ It already does open/close the menu when you hit SPACE or ENTER.
- [ ] When the menu is open, the first time we press the tab key, it should bring the focus to the first menu item on the list. Right now it jumps to the light/dark theme button.
- [ ] Should the menu button be in a nav element?
- [ ] Should there be aria-expanded and aria-label attributes on the menu button?
- [ ] When the menu is closed and you hit tab, should the focus go first to the light/dark theme button, and then to the menu button?

### DIFF
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -526,7 +526,7 @@ blockquote {
     .menu-button {
         @include font-size(20);
         background: $green-dark;
-        border-radius: 23px;
+        border: 0;
         color: var(--menu);
         cursor: pointer;
         display: block;
--- a/djangoproject/static/js/mod/mobile-menu.js
+++ b/djangoproject/static/js/mod/mobile-menu.js
@@ -12,7 +12,7 @@ define([
         init: function(){
             var self = this;
             self.menu.addClass('nav-menu-on');
-            self.button = $('<div class="menu-button"><i class="icon icon-reorder"></i><span>Menu</span></div>');
+            self.button = $('<button class="menu-button"><i class="icon icon-reorder"></i><span>Menu</span></button>');
             self.button.insertBefore(self.menuBtn);
             self.button.on( 'click', function(){
                 self.menu.toggleClass('active');


knyghty:
@thibaudcolas what do you think?

This doesn't fix the issue that you can tab through the links without expanding the nav but this still seems like an improvement.

thibaudcolas (approved):
Working great! Thank you @ontowhee

CuriousLearner (approved):

ontowhee:
@thibaudcolas @CuriousLearner Thank you for the approval! Do we want to create an issue for the "Other Notes" mentioned in the description above?

thibaudcolas:
>  When the menu is open, the first time we press the tab key, it should bring the focus to the first menu item on the list. Right now it jumps to the light/dark theme button.

Yes – please open an issue

>  Should the menu button be in a nav element?

I don’t think it matters too much

> Should there be aria-expanded and aria-label attributes on the menu button?

`aria-expanded` definitely

> When the menu is closed and you hit tab, should the focus go first to the light/dark theme button, and then to the menu button?

I’d expect the focus to go first to a "Skip link" on the page (#1367), then to the Django logo, then theme toggle, then menu toggle.

PR #1419, opened 2023-10-20
Fixed 1411 : h1_changed_to_h2

specialorange:
…ingle h1 per page

### DIFF
--- a/djangoproject/templates/homepage.html
+++ b/djangoproject/templates/homepage.html
@@ -17,7 +17,7 @@
 
 {% block content %}
   <div class="section">
-    <h1>Meet Django</h1>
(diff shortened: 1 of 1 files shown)


sabderemane:
Hi @specialorange, can you please also post a screenshot of the site after your changes?

thibaudcolas:
👋 we want this change but this needs quite a bit more work. If we do this for the homepage only, we might as well do this for all `<h1>` on the page to not introduce inconsistencies between the sections. Possibly change the styles so we only change the semantics and not the aesthetics.

As-is this creates an inconsistency which is more problematic than the heading level semantics so I’ll close now.

Issue #1347, opened 2023-04-15
Improve 404 page


adamchainz:
Take a broken docs link like: https://docs.djangoproject.com/en/4.2/ref/databasess/ (two "s" chars at the end instead of one).

The 404 page could at least show the docs search bar to make it easy to search for the relevant page.

For bonus points, it could even show search results for the keywords in the URL ("databasess") or show close matches from existing URL's (using e.g. [difflib.get_close_matches](https://docs.python.org/3/library/difflib.html#difflib.get_close_matches)).

gautam8404:
like this?

![image](https://user-images.githubusercontent.com/93448885/235138444-ade638cb-49ed-47f9-b2d5-39cbdf80a6e2.png)

however i don't thinks what i did is a proper way of doing it, i used same search as in documentation page by making some changes in `templatetags/docs.py`

````@register.inclusion_tag("docs/search_form.html", takes_context=True)
def search_form(context):
    request = context["request"]
    version = context["version"] if "version" in context else context["DJANGO_VERSION"]
    lang = context["lang"] if "lang" in context else context["LANGUAGE_CODE"]
    release = DocumentRelease.objects.get_by_version_and_lang(
        version,
        lang,
    )
    return {
        "form": DocSearchForm(request.GET, release=release),
        "version": version,
        "lang": lang,
    }
```

and in `404.html` 

```

{% extends 'base_error.html' %}
{% load docs %}

{% block title %}Page not found{% endblock %}
{% block header-classes %}
container--flex container--flex--wrap--mobile
{% endblock %}
{% block header %}
  <h1>404</h1>

  {% search_form %}

{% endblock %}

{% block content %}
<h2>Page not found</h2>

<p>Looks like you followed a bad link. If you think it's our fault, please <a href="https://code.djangoproject.com/">let us know</a>.</p>

<p>Here's a link to the <a href="{% url 'homepage' %}">homepage</a>. You know, just in case.</p>

{% endblock %}
```

A more proper way would be writing a handler for 404? but maybe i should ask permission for it before

Om15102003:
@adamchainz I would like to contribute on this issue.

SurajSanap:
fixed by https://github.com/django/djangoproject.com/pull/1454

bengali-creates:
hi i ma new to this project i want to contribute in this project for this problem . Can anyone pls tell me where can i find the 404.html file i am having a hard time finding it

bmispelon:
[...]

PR #1454, opened 2023-12-20
Improve 404 Page: Enhanced Search Functionality and Close Matches for Broken Docs Links

SurajSanap:
- Increased the prominence of the "404" header using a mixture of Black (#2b3137) and Green (#2dba4e) colors.
- Styled the search bar to resemble GitHub's design with clear input and button elements.
- Removed the default value from the search box for a cleaner user experience.

Closes: #1347

### DIFF
--- a/djangoproject/templates/404.html
+++ b/djangoproject/templates/404.html
@@ -2,13 +2,36 @@
 
 {% block title %}Page not found{% endblock %}
 
-{% block header %}<h1>404</h1>{% endblock %}
+{% block header %}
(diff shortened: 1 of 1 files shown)


thibaudcolas:
@SurajSanap thank you for giving this a go but this isn’t right. We don’t use inline styles. The site has an existing search component, there’s no need to use GitHub’s designs. And the code you wrote clearly won’t work, the form points to a search endpoint which doesn’t exist.

PR #1455, opened 2023-12-20
Allow using formatted text for ongoing business in DSF meeting minutes

thibaudcolas:
`body_html` contains the markup, `body` is only the plain-text source. This matches how `new_business` renders a few lines below.


thibaudcolas:
Thanks @felixxm! I’ve updated the content now that this is deployed.

PR #1457, opened 2023-12-21
Update commands in download.html

hvmanITas:
This patch changes the Django installation command.

Installing Django without selecting pip as a Python module is not a good idea for certain users, specially Linux users.

### DIFF
--- a/djangoproject/templates/releases/download.html
+++ b/djangoproject/templates/releases/download.html
@@ -30,7 +30,7 @@
   <p>The latest official version is {{ current.version }}{% if current.is_lts %} (LTS){% endif %}. Read the
     {% release_notes current.version show_version=True %}, then install it with
     <a href="https://pip.pypa.io/en/latest/">pip</a>:</p>
-  <p>Linux / MacOs:</p>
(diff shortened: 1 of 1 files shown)


djangoproject/templates/releases/download.html
```diff
@@ -30,7 +30,10 @@ <h2>Option {% cycle '1' '2' '3' as options %}: Get the latest official version</
   <p>The latest official version is {{ current.version }}{% if current.is_lts %} (LTS){% endif %}. Read the
     {% release_notes current.version show_version=True %}, then install it with
     <a href="https://pip.pypa.io/en/latest/">pip</a>:</p>
-  <pre class="literal-block"><code>pip install Django=={{ current.version }}</code></pre>
+  <p>Linux / MacOs:</p>
```
ubernostrum, line 33: Should probably be "macOS" for consistent with how Apple capitalizes it.

ubernostrum:
I think this is a good idea -- the `python -m pip` usage has been strongly preferred and recommended by the packaging maintainers for quite a while now.

camilonova (approved):

PR #1461, opened 2024-01-04
Add a link to Corporate membership in the site footer

thibaudcolas:
This [Corporate membership](https://www.djangoproject.com/foundation/corporate-membership/) page is hard to find currently even if you know it exists. From Catherine Holmes on the DSF Slack:

> The only way I know from the [Fundraising](https://www.djangoproject.com/fundraising/) page to get to the corporate members is to scroll down to the corporate members and click on "corporate members" 

I can’t find the link to this page even with Catherine’s instructions, so it feels like a footer link would definitely help.


sabderemane (approved):
LGTM, thank you @thibaudcolas 🚀

PR #1468, opened 2024-01-16
Fix table style for foundation flatpage

sabderemane:
Prior to this change, the foundation page have 2 tables with previous boards which are not responsive and hard to distinguish rows in mobile.

This change add two CSS classes which fixes this issue.

### BEFORE
Capture d’écran 2024-01-16 à 01 52 27
Capture d’écran 2024-01-16 à 01 52 45

### AFTER
**Dark Mode**
Capture d’écran 2024-01-16 à 01 41 21
Capture d’écran 2024-01-16 à 01 41 28
Capture d’écran 2024-01-16 à 01 41 42

**Light Mode**
Capture d’écran 2024-01-16 à 01 44 52
Capture d’écran 2024-01-16 à 01 46 03

### DIFF
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -2608,6 +2608,19 @@ dl.data {
     }
 }
 
+.table-wrapper {
+    // This class is used to avoid the table exceeds the size of the content
+    // and add a scrollbar to see the rest of the content in mobile
+    display: grid;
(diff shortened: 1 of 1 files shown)


thibaudcolas:
I was able to answer my own question – and have updated the flatpages’ HTML already. So this should be live once the site is deployed again.

PR #1478, opened 2024-02-09
Added ticket resolution states to trac.sql

bmispelon:
This makes it easier to run Trac locally, as it removes an error message about missing resolutions when first starting it.

I copied the values from the admin panel in production.


bmispelon:
Here's the error message that this commit fixes:

Screenshot 2024-02-09 at 10-25-54 Configuration Error – Django

bmispelon:
Thanks! 🚀

PR #1489, opened 2024-03-02
Improve docs version-switcher mobile styles (django#1488)

circus2271:


circus2271:
I had troubles with a local database, so i managed up only two versions of language.. the same is with documentation versions (in example above there is only 2 version of documentation and 2 version of language, which may seem confusing..

tim-schilling:
@circus2271 If the bug here is that using the mobile version of the site and clicking the buttons multiple times results in the extra utilities inconsistently showing the other options, that's still happening even with the change.

tim-schilling:
@circus2271 I think this is a byproduct of Browser Developer Tool's "Touch Simulation" feature and then clicking quickly. I see two events in the console when logging them on this click handler when this problem is triggered. One has a `pointerType: "mouse"` and the other `pointerType: "touch"`

adamzap:
Thanks for looking into this, @tim-schilling! @circus2271 Do you think we can close this?

circus2271:
Hi, @tim-schilling 

The bug I tried to fix is that menu remains opened when I double tap on its button. This behaviour is due to css styles.
 
This menu opens due to css styles and due to js at the same time. If I hover this button on desktop, it opens. When I click on it, it remains opened. The same thing happens on mobile version: second tap on button makes it hidden through js, but hover styles are still applied, so this menu remains visible. The fix is to remove those hover styles from mobile version as they are redundant


Here is some short video to illustrate this idea:

https://github.com/user-attachments/assets/47ff8302-fb63-4ef2-8433-c45c87fd0b82

circus2271:
@adamzap 

Maybe it needs some more investigation, as there was no clear description here

tim-schilling:
@circus2271 Can you include a series of steps to reproduce this that isn't just a video please? I'm seeing what I would consider weird behavior on my phone with the production site, but I'm not sure it's the same as what you're trying to fix.

circus2271:
Ok, @tim-schilling 

Open https://docs.djangoproject.com/en/5.2/, tap on language button, second tap on it doesn't close the menu

tim-schilling:
@circus2271 thank you for that clarification! I was getting tripped up earlier by the term "double-click". I assumed that mean two very quick clicks in succession. Not two separate clicks.

circus2271:
no problem @tim-schilling 

thank you for your attention

tim-schilling (approved):
I was able to reproduce this and confirm this change should work. Thanks @circus2271 for helping me through that.

ulgens:
@circus2271 It looks like the branch will need to be rebased over the latest main.

adamzap:
Closing in favor of #2472.

PR #1497, opened 2024-03-26
Make it possible to use a different database user

czue:
In development it's convenient to not have to create separate database users for every project. this allows it to be configurable without changing any of the defaults.


bmispelon:
Thanks! 🚀

PR #1507, opened 2024-03-26
Make it possible to force build from update_docs_and_index

czue:
If you run `./manage.py update_docs_and_index --force` the `--force` flag gets mapped to the built-in `--force-color` attribute from `BaseCommand`, and then does not get passed through to the `update_docs` command that gets called.

This allows passing `--force` and passing it through to `update_docs`.

### DIFF
--- a/docs/management/commands/update_docs_and_index.py
+++ b/docs/management/commands/update_docs_and_index.py
@@ -7,5 +7,16 @@ class Command(BaseCommand):
     Update the docs then reindex them in the search vector field.
     """
 
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--force",
+            action="store_true",
+            default=False,
+            help=(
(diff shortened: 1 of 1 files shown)


bmispelon:
Hi and thanks again for the contribution!

I was going to push a commit that adds a small test for this feature when I realized that maybe we don't need this command anymore.
Apparently I was the one who introduced it in 4465952646b630ec5e858c6c912bf5c4cf684eb6 (back when we used elasticsearch). Since then the `update_docs` command added support for the `--update-index` flag (8a8c33a89b664b59089f3de221f7c2971a437319) which I think makes this dedicated command a bit redundant (I would even say useless).

So I made a PR to simply remove it instead: #1519.

What do you think?

czue:
Oops, sorry this notification landed in the wrong tab. Removing it seems like a much better option! Will close this PR.

PR #1508, opened 2024-03-26
add og:title to the docs

czue:
This uses the same og:title as the page title.

The hooks are [already in the base template](https://github.com/django/djangoproject.com/blob/main/djangoproject/templates/base.html#L23) so this is very straightforward.

image

### DIFF
--- a/docs/templates/docs/doc.html
+++ b/docs/templates/docs/doc.html
@@ -2,7 +2,7 @@
 {% load i18n fundraising_extras docs %}
 
 {% block title %}{{ doc.title|striptags|safe }} | {% trans "Django documentation" %}{% endblock %}
-
+{% block og_title %}{{ doc.title|striptags|safe }} | {% trans "Django documentation" %}{% endblock %}
 {% block doc_url %}{% url 'document-index' lang=lang version=version host 'docs' %}{% endblock %}
 
 {% block link_rel_tags %}
--- a/docs/templates/docs/doc.html
+++ b/docs/templates/docs/doc.html
@@ -2,7 +2,8 @@
 {% load i18n fundraising_extras docs %}
 
 {% block title %}{{ doc.title|striptags|safe }} | {% trans "Django documentation" %}{% endblock %}
-{% block og_title %}{{ doc.title|striptags|safe }} | {% trans "Django documentation" %}{% endblock %}
+{% block og_title %}{{ doc.title|striptags|escape|safe }} | {% trans "Django documentation" %}{% endblock %}
+
 {% block doc_url %}{% url 'document-index' lang=lang version=version host 'docs' %}{% endblock %}
 
 {% block link_rel_tags %}
--- a/docs/templates/docs/doc.html
+++ b/docs/templates/docs/doc.html
@@ -2,7 +2,7 @@
 {% load i18n fundraising_extras docs %}
 
 {% block title %}{{ doc.title|striptags|safe }} | {% trans "Django documentation" %}{% endblock %}
-{% block og_title %}{{ doc.title|striptags|escape|safe }} | {% trans "Django documentation" %}{% endblock %}
+{% block og_title %}{{ doc.title|striptags|escape }} | {% trans "Django documentation" %}{% endblock %}
 
 {% block doc_url %}{% url 'document-index' lang=lang version=version host 'docs' %}{% endblock %}
 
--- a/docs/templates/docs/doc.html
+++ b/docs/templates/docs/doc.html
@@ -2,7 +2,7 @@
 {% load i18n fundraising_extras docs %}
 
 {% block title %}{{ doc.title|striptags|safe }} | {% trans "Django documentation" %}{% endblock %}
-{% block og_title %}{{ doc.title|striptags|escape }} | {% trans "Django documentation" %}{% endblock %}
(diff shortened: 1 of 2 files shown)


thibaudcolas (commented):
Wouldn’t we need an [escape](https://docs.djangoproject.com/en/5.0/ref/templates/builtins/#escape) as well to protect from titles with double quotes? Considering the `og:title` value is an attribute while `<title>` is a tag.

czue:
Oh, good point. Will check and update.

docs/templates/docs/doc.html
```diff
@@ -2,7 +2,7 @@
 {% load i18n fundraising_extras docs %}
 
 {% block title %}{{ doc.title|striptags|safe }} | {% trans "Django documentation" %}{% endblock %}
-
+{% block og_title %}{{ doc.title|striptags|safe }} | {% trans "Django documentation" %}{% endblock %}
```
knyghty, line 5: I think we could wrap with `{% with %}` to reduce the duplication here

czue: good idea, though it doesn't seem to work. According to ChatGPT this is because "the scope of the variables in Django's `{% with %}` statement is limited to the block in which they are defined". So since the variables are used inside separate blocks it doesn't work. We could probably pass a variable in the view if you think it's worth looking into that.

czue: Also, given that we also now want to escape the one in the meta tag but not the one in the title, I don't think there's any easy way to avoid the duplication.

czue:
@thibaudcolas I added escaping and confirmed it does the right thing
image

docs/templates/docs/doc.html
```diff
@@ -2,6 +2,7 @@
 {% load i18n fundraising_extras docs %}
 
 {% block title %}{{ doc.title|striptags|safe }} | {% trans "Django documentation" %}{% endblock %}
+{% block og_title %}{{ doc.title|striptags|escape|safe }} | {% trans "Django documentation" %}{% endblock %}
```
bmispelon, line 5: I don't think the `|safe` is needed here, since `|escape` marks its output as safe as far as I know

bmispelon (changes requested):
I put together a test for the new behavior (add it to `docs/tests.py`):
```
from django.test import RequestFactory
from django.template.loader import render_to_string

class TemplateTestCase(TestCase):
    def _assertOGTitleEqual(self, doc, expected):
        output = render_to_string(
            "docs/doc.html",
            {"doc": doc, "lang": "en", "version": "5.0"},
            request=RequestFactory().get("/")
        )

        self.assertInHTML(f'<meta property="og:title" content="{expected}" />', output)

    def test_opengraph_title(self):
        doc = Document.objects.create(
            release=DocumentRelease.objects.create(
                lang='en',
                release=Release.objects.create(version="5.0"),
            ),
        )
        doc.body = "test body"  # avoids trying to load the underlying physical file

        for title, expected in [
            ("test title", "test title"),
            ("test & title", "test &amp; title"),
            ("test <strong>title</strong>", "test title"),
        ]:
            doc.title = title
            with self.subTest(title=title):
                self._assertOGTitleEqual(doc, f"{expected} | Django documentation")
```

docs/templates/docs/doc.html
czue: good call: https://github.com/django/djangoproject.com/pull/1508/commits/7545e3ccb0a9a5f20f7119f6e76fb0069b8b633c

Also, based on the test even the `|escape` wasn't necessary (maybe because autoescape is on?). But I left it since it seemed more explicit. But let me know if you think worth removing

czue (commented):
@beingbiplov thank you for the test code! made the suggested changes.

bmispelon:
> Also, based on the test even the `|escape` wasn't necessary (maybe because autoescape is on?). But I left it since it seemed more explicit. But let me know if you think worth removing

Oh good point, I didn't think about `autoescape`. It doesn't make much difference in the end, but I think I'd personally prefer to remove the `|escape` if it's not actually doing anything. As long as we have a test, it should break and let us know in the future if we somehow turn off autoescape.

I don't seem to have the right permissions to allow the github actions to run on this PR. Can someone from @django/djangoproject-com-maintainters approve it?

czue:
@bmispelon I removed the redundant escape

docs/tests.py
```diff
@@ -592,3 +593,32 @@ def test_search_update(self):
...
+                lang="en",
+                release=Release.objects.create(version="5.0"),
+            ),
+        )
+        doc.body = "test body"  # avoids trying to load the underlying physical file
+
+        for title, expected in [
+            ("test title", "test title"),
+            ("test & title", "test &amp; title"),
+            ('test "title"', "test &quot;title&quot;"),
```
bmispelon, line 619: Good thinking on this one too 👍🏻

bmispelon (approved):
The change looks good to me, thanks again for taking the time ✨

czue:
Just confirming there are no other steps/blockers on my side before this is merged? (First time contributing so don't know the process).

BTW I had a few other suggestions based on setting up my dev environment (#1507 and #1497). Not sure if there is a process for getting those looked at.

knyghty:
> Just confirming there are no other steps/blockers on my side before this is merged? (First time contributing so don't know the process).

I don't think so, just a bit of patience, the website team are all volunteers and have other, more paid things to do.

camilonova (approved):

PR #1515, opened 2024-04-08
Enable profiling of transactions in Sentry

tobiasmcnulty:
https://docs.sentry.io/platforms/python/profiling/#enable-profiling

### DIFF
--- a/djangoproject/settings/prod.py
+++ b/djangoproject/settings/prod.py
@@ -83,6 +83,7 @@ if "sentry_dsn" in SECRETS and not DEBUG:
             RedisIntegration(),
         ],
         traces_sample_rate=0.1,
+        profiles_sample_rate=0.1,
     )
 
 # RECAPTCHA KEYS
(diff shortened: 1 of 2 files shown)


bmispelon:
The document you linked states that:

> The profiles_sample_rate setting is relative to the traces_sample_rate setting.

So I my math is right, that means only 1% of transactions will be profiled. Was that the intent?

tobiasmcnulty:
@bmispelon Excellent catch! I missed that. I suppose I intended to set `profiles_sample_rate=1.0`, but I am open to any suggestions.

bmispelon:
It's unclear to me what kind of overhead the profiling has.
I think you'll be the one living with the consequences, as I'm not on the ops team (yet? 🙃) so it's fair if you get to decide. 10% seems as good a percent as any. 👍🏻

bmispelon (approved):

tobiasmcnulty:
Thanks for the review! This has been deployed.

PR #1517, opened 2024-04-10
Added index on `aggregator_feeditem (date_modified DESC)`

tobiasmcnulty:
For pages early in the [aggregator/community feeds](https://www.djangoproject.com/community/blogs/), there is a [tremendous speedup](https://www.postgresql.org/docs/current/indexes-ordering.html) when an index is added on `aggregator_feeditem (date_modified DESC)` (i.e., in the order in which the FeedItems are returned). In my tests, it reduced the query time from ~170ms to ~6ms. The change will help reduce load on the database server and free up uWSGI processes for other requests.


bmispelon (changes requested):
This looks great, especially if the speedups you observed locally hold up in production also.

One nitpick I had would be to change the name of the migration file to something slightly more readable. Maybe `0005_feeditem_add_index_date_modified.py`. What do you think?

tobiasmcnulty:
Thanks for the review!

> One nitpick I had would be to change the name of the migration file to something slightly more readable. Maybe `0005_feeditem_add_index_date_modified.py`. What do you think?

That is a favorite nitpick of mine, I am surprised I let it slip, lol. Updated!

bmispelon (approved):

PR #1523, opened 2024-04-24
Used flex instead of floats for header layout

### LINKED ISSUE #1435: When the "hamburger" menu is open, pressing the tab key for the first time should bring the focus to the first item in the menu list
When the menu is open, the first time we press the tab key, it should bring the focus to the first menu item on the list. Currently, it jumps to the light/dark theme button.

bmispelon:
Fixed #1435

### DIFF
--- a/djangoproject/templates/includes/header.html
+++ b/djangoproject/templates/includes/header.html
@@ -1,14 +1,14 @@
 <div role="banner" id="top">
-  <div class="container">
+  <div class="container container--flex--wrap--mobile">
     <a class="logo" href="{% url 'homepage' %}">Django</a>
     <p class="meta">The web framework for perfectionists with deadlines.</p>
+    <div class="mobile-toggle">
+      {% include "includes/toggle_theme.html" %}
+    </div>
     <button class="menu-button">
       <i class="icon icon-reorder"></i>
       <span class="visuallyhidden">Menu</span>
     </button>
-    <div class="mobile-toggle">
-      {% include "includes/toggle_theme.html" %}
-    </div>
     <div role="navigation">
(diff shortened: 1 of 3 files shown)


djangoproject/templates/includes/header.html
```diff
@@ -1,14 +1,15 @@
 <div role="banner" id="top">
-  <div class="container">
+  <div class="container container--flex--wrap--mobile">
     <a class="logo" href="{% url 'homepage' %}">Django</a>
     <p class="meta">The web framework for perfectionists with deadlines.</p>
+    <div class="spacer"></div>
```
bmispelon, line 5: I don't like having this in the HTML but I couldn't figure out how to achieve the same result with CSS only. Suggestions welcome!

sabderemane: Did you try to use `justify-content: space-around` or `justify-content: space-between`? (I didn't try, might not work but I can have a look later)

sabderemane (commented):
Looks nice! ✨

djangoproject/templates/includes/header.html
matthiask: Another thing which might work is `margin-left: auto;` on `.mobile-toggle`. Margins sometimes work surprisingly well inside flex containers.

bmispelon: Finally got around to trying the `margin-left` approach and it's working perfectly. Thanks for the tip! 👍🏻

matthiask: Nice!

bmispelon:
Thanks everyone! ✨

PR #1531, opened 2024-05-15
Update fontdeck.html DOM text reinterpreted as HTML

Shivam7-1:
By using innerText, it will avoid the risk of HTML injection, as these properties automatically escape any HTML special characters in the provided text. This helps prevent cross-site scripting (XSS) vulnerabilities by treating the input as plain text rather than interpreted HTML.


Shivam7-1:
Hi @bmispelon Could You Please Review This PR 
Thanks

bmispelon:
Hi,

We are not interested in these changes you're proposing. Please don't submit more of these search/replace pull requests, and please don't tag maintainers asking for a review after such a short time.

Have a nice day

PR #1534, opened 2024-05-17
Add new local django communities with images

Gandi24:
image

image


Gandi24:
This is an accidental Pull request, apologies.

Issue #1381, opened 2023-06-22
Accessibility issues with site design and implementation


thibaudcolas:
During DjangoCon Europe 2023, I did [accessibility testing on the website’s homepage](https://thib.me/djangocon-europe-2023-accessibility-review-of-djangoproject-com). Here is a recap of issues I spotted, split into the categories I normally use, with links to my testing video’s timestamps, WCAG success criteria, and existing GitHub issues where appropriate.

Note though I only tested the homepage, there were very few content-related issues – I believe the vast majority of issues listed here are present site-wide.

### Blockers

> Blocker: Critical issue. Certain users cannot perceive or operate this, with no workaround available.

- The "hamburger" menu toggle is impossible to use with a keyboard. This makes it impossible for keyboard and screen reader users to access menu items and navigate the site in this way. ([Menu button is not a button #1234](https://github.com/django/djangoproject.com/issues/1234), [18:18](https://www.youtube.com/watch?v=dAiMPTOMh1U&t=1098s), [SC 2.1.1 Keyboard](https://www.w3.org/TR/WCAG22/#keyboard)).
- The page’s second column content is hidden in print stylesheets. This makes it impossible for users of magnifiers to access this content ([22:00](https://www.youtube.com/watch?v=dAiMPTOMh1U&t=1320s)).

### Major

> Major: Certain users are very likely to have difficulty perceiving or operating this.

- Heading level 1 should be used for the title of the page / visually most prominent heading, ideally not for invisible content ("Django" on homepage). Ideally each page would only have a single h1. ([07:04](https://www.youtube.com/watch?v=dAiMPTOMh1U&t=424s), [SC 1.3.1 Info and Relationships](https://www.w3.org/TR/WCAG22/#info-and-relationships)).

### Confusing

> Confusing: Understanding or operating this may be difficult for some users.

- Focus style almost invisible on "Get started with Django" CTA ([02:30](https://www.youtube.com/watch?v=dAiMPTOMh1U&t=150s), [SC 1.4.11 Non-Text Contrast](https://www.w3.org/TR/WCAG22/#non-text-contrast)).
- Links should be identifiable with other means than color ([05:30](https://www.youtube.com/watch?v=dAiMPTOMh1U&t=330s), [SC 1.4.1 Use of Color](https://www.w3.org/TR/WCAG22/#use-of-color)).
[...]

PR #1545, opened 2024-06-11
Fixes focus indicators

SaptakS:
Refs #1381 

Improves the focus indicators. Based on WCAG 2.2 standards, it is a better idea to have solid outlines instead of a dotted outline for focus indicators. Also, color change for focus indication isn't great because it doesn't often work for color blinded people.

So things done in this PR:
- Make all anchor tags have a solid 2px outline of the currentColor (Sadly the currentColor itself doesn't meet WCAG color contrast requirement, so the focus indicator doesn't either, but hopefully that's a separate discussion)
- Apply the same above design to all buttons as well
- Adds an offsetted focus outline for the CTAs
- Moves padding in navigation elements from the `<a>` element to the `<li>` element.

These address the below concerns by @thibaudcolas in #1381 
> Focus style almost invisible on "Get started with Django" CTA

> Custom "dotted outline" focus styles: not visible enough and broken in main navigation

Attaching some screenshots of the changes:
image
image

### DIFF
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -590,12 +590,14 @@ blockquote {
             text-transform: uppercase;
             margin: 0 10px;
             border-top: 1px solid lighten($green-dark, 5%);
+            padding: 20px 0px;
 
             @include respond-min(768px) {
                 margin: 0;
                 border: 0;
                 float: left;
                 text-align: left;
+                padding: 20px 10px;
             }
 
             &.active a {
@@ -606,7 +608,6 @@ blockquote {
         a {
             color: var(--menu);
             display: block;
-            padding: 20px 0px;
             text-decoration: none;
 
             &:active,
@@ -615,7 +616,6 @@ blockquote {
             }
 
             @include respond-min(768px) {
-                padding: 20px 10px;
             }
         }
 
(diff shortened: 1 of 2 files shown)


djangoproject/scss/_style.scss
```diff
@@ -611,7 +620,6 @@ blockquote {
             }
 
             @include respond-min(768px) {
-                padding: 20px 10px;
             }
```
bmispelon, line 623: The whole `@include` is now empty, so it can be deleted, right?

djangoproject/scss/_style.scss
```diff
@@ -1229,11 +1237,15 @@ a.cta {
         font-style: normal;
     }
 
-    &:hover,
-    &:focus {
+    &:hover{
```
bmispelon, line 1240: Missing a space before the bracket:

```suggestion
    &:hover {
```

bmispelon (changes requested):
This seems like a good change.  I've tested it locally and it works.

I have just two minor comments which don't affect the feature itself.

Thanks for working on this! ✨

SaptakS:
Thanks @bmispelon for pointing those out. Have updated them.

bmispelon (approved):

bmispelon:
Thanks for the quick update. This looks good to me now, unless anyone objects I'll merge and deploy this over the coming weekend.

bmispelon:
And it's live, well done everybody 👏🏻

Issue #1562, opened 2024-06-27
Add a test to check that there are no pending migrations


bmispelon:
We use `django-upgrade` (with pre-commit) and its author has recommended having a test for pending migrations: https://mastodon.social/@adamchainz@fosstodon.org/112687118840974800

This is to ensure that code modifications by `django-upgrade` don't result in forgotten migrations.

Adam has a blog post about how do do that (see the link above) so this should be a good first-time contribution if anyone is tempted. I would place the test in `djangoproject/tests.py` since it's not tied to a specific app.

Lakshya272:
I am new to contributions, where should I write the code for the test to resolve the issue? should I directly go to djangoproject/tests.py and make changes there to make a pr, or write it in a comment for you to paste it onto the file?

PR #1567, opened 2024-07-01
Update djangoproject/tests.py for #1562

Lakshya272:
Added a test to check that there are no pending migrations

Resolved issue-  Add a test to check that there are no pending migrations #1562


bmispelon:
Perfect, thanks so much! 🚀

PR #1575, opened 2024-07-01
Bump pre-commit and docutils

bmispelon:
Bumping them to their last version that supports Python 3.8

Sphinx is a pretty central dependency for us (since it's the tool that builds our documentation), so I'd like to do a few more manual checks before merging:

- [ ] Make sure the documentation builds locally
- [ ] Does the `Sphinx` version need to be synced with django/django?


pauloxnet:
These changes have been accomplished in two seprated PR: #1645 and #1696
I close this PR

Issue #1620, opened 2024-09-12
dev-warning strip on Django docs development version covers the search bar when used shortcut search key


khanxmetu:
Minor issue. The floating warning strip at the top on https://docs.djangoproject.com/en/dev/ covers the search bar underneath when pressed shortcut search key (<kbd>CTRL</kbd> + <kbd>K</kbd>) to search.

PR #1624, opened 2024-09-23
fix: smooth scroll to search input with dynamic banner offset

FreakPirate:
### Overview
This pull request addresses the scrolling behavior when focusing on the search input field via the shortcut (Ctrl + K / Cmd + K). The previous implementation did not account for the height of the warning banner, resulting in a less-than-ideal user experience.

### Changes Made
- **Adjusted Scroll Behavior**: The scroll position now considers the height of the `#dev-warning` banner, ensuring that the search input is not obscured.
- **Replaced `scrollIntoView`**: The code now uses `window.scrollTo` to smoothly scroll to the search input's position while accounting for the banner's height.
- **Improved User Experience**: This change provides a seamless transition when the search input is focused, improving overall usability.

### Testing
- Tested the functionality across different browsers to ensure compatibility.
- Verified that the scrolling behaves as expected with and without the warning banner present.

---

### Issue Reference
This pull request fixes https://github.com/django/djangoproject.com/issues/1620.

### DIFF
--- a/djangoproject/static/js/mod/search-key.js
+++ b/djangoproject/static/js/mod/search-key.js
@@ -17,8 +17,14 @@ define([
 
                 $(window).keydown(function(e) {
                     if ((e.metaKey || e.ctrlKey) && e.key === 'k' && $('input:focus, textarea:focus').length === 0) {
+                        const warning_banner_height = $('#dev-warning').outerHeight() || 0;
+                        const search_form_input_top = search_form_input.offset().top;
+                        const scroll_to_position = search_form_input_top - warning_banner_height;
                         search_form_input.focus().select();
-                        search_form_input[0].scrollIntoView({ behavior: "smooth", block: "start" });
+                        window.scrollTo({
+                            top: scroll_to_position,
+                            behavior: 'smooth'
+                        });
                         return false;
                     }
                 });


djangoproject/static/js/mod/search-key.js
```diff
@@ -17,8 +17,14 @@ define([
 
                 $(window).keydown(function(e) {
                     if ((e.metaKey || e.ctrlKey) && e.key === 'k' && $('input:focus, textarea:focus').length === 0) {
+                        const warning_banner_height = $('#dev-warning').outerHeight() || 0;
```
bmispelon, line 20: There's also a banner when the version is [outdated](https://docs.djangoproject.com/en/1.8/), but it uses a different id (`#outdated-warning`). I would suggest using the class instead:
```suggestion
                        const warning_banner_height = $('.doc-floating-warning').outerHeight() || 0;
```

bmispelon (dismissed):
Hi and thanks so much for this. I've tested it locally and it seems to work very well.

One thing I noted is that the overlapping problem also shows up on outdated pages (like https://docs.djangoproject.com/en/1.8/) so it'd be nice to have it fixed there too. I've made a suggestion for how to do that but if you have a better idea I'm all ears.

Let me know what you think!

adamzap:
@FreakPirate Could you address the comments on this PR if you're still interested in working on it? Otherwise it will be closed in a week. Thanks!

FreakPirate:
Hi @adamzap . I must have missed the comment. Let me address it right away. Thanks for being considerate 🙌

djangoproject/static/js/mod/search-key.js
FreakPirate: Thanks @bmispelon for the suggestion. Apologies for responding so late, I must have missed this comment.

FreakPirate: Since search-key.js was [refactored](https://github.com/django/djangoproject.com/pull/1933) into `djangoproject.js`, I have make the suggested changes in the designated place. Please re-review. 🙌

adamzap:
@FreakPirate Thanks for this! Let me see if we are able to get #2075 in before we move forward here.

adamzap:
@FreakPirate Thanks for your patience here.

@bmispelon I think it's unfortunate that this code has to account for the floating warning element specifically. If we added another element with a fixed position at the top of the page, we'd have to account for it in this JavaScript code. I suppose another option would be to just scroll to the top of the page. Does that seem cleaner to you, or does it worsen the user experience here?

bmispelon:
> @bmispelon I think it's unfortunate that this code has to account for the floating warning element specifically. If we added another element with a fixed position at the top of the page, we'd have to account for it in this JavaScript code. I suppose another option would be to just scroll to the top of the page. Does that seem cleaner to you, or does it worsen the user experience here?

I don't have super strong opinions honestly. It's true that this code is a little brittle and would need updating if and when new fixed elements are added, but how often does that happen realistically? Scrolling simply to the top of the page seems like a good compromise since that's where the search bar currently is, but could break if we'd redesigned the site and moved it somewhere else (but that seems even less likely than adding another fixed element).

I defer to what you think is best @adamzap (if you forced me to have an opinion, I'd vote for scrolling to the top but only by a tiny margin).

adamzap:
Ok thanks!

@FreakPirate Would you be willing to update this pull request to make it scroll to the top of the page instead?

FreakPirate:
Sure thing. Just to confirm: we’ll change the behavior to scroll all the way to the top, assuming the search bar stays near the header. That still feels safe even if we refresh the layout later.
Let me know if I am aligned and I’ll update the PR.

adamzap:
@FreakPirate Yes, that sounds correct. Also, please try to follow the code style already established in the file. Thanks!

FreakPirate:
There you go @adamzap @bmispelon , let me know if this is okay. Thanks 🙏

djangoproject/static/js/djangoproject.js
```diff
@@ -97,8 +97,15 @@ window.addEventListener('keydown', function (e) {
   const el = document.querySelector('#id_q');
 
   el.select();
-  el.focus({ preventScroll: true });
-  el.scrollIntoView({ behavior: 'smooth' });
+  el.focus();
+
+  // Scroll to top instead of calculating banner offsets for simpler, more robust behavior
+  // This ensures the search input is always visible regardless of floating elements
+  // https://github.com/django/djangoproject.com/pull/1624#issuecomment-3062351229
```
adamzap: Could you remove this comment? The `git blame` will suffice.

djangoproject/static/js/djangoproject.js
```diff
@@ -97,8 +97,15 @@ window.addEventListener('keydown', function (e) {
...
+  window.scrollTo({
```
adamzap: This is a nitpick, but would you mind making this one line to match the style of similar lines in this file?

djangoproject/static/js/djangoproject.js
```diff
@@ -97,8 +97,9 @@ window.addEventListener('keydown', function (e) {
   const el = document.querySelector('#id_q');
 
   el.select();
-  el.focus({ preventScroll: true });
-  el.scrollIntoView({ behavior: 'smooth' });
+  el.focus();
+
+  window.scrollTo(0, 0);
```
adamzap: Could you use the `options` way of calling `scrollTo` so we can preserve smooth scrolling? See the second example [here](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTo#examples):

```suggestion
  window.scrollTo({ top: 0, left: 0, behavior: 'smooth' });
```

FreakPirate: I am confused. I used options way the last time, but you suggested [this](https://github.com/django/djangoproject.com/pull/1624#discussion_r2258390073) nit. And the file contains one-line references only.

adamzap: Sorry, I'm not trying to be confusing. I can't see the commit you are referring to because you squashed it away.

It seems to me that in the current state of the branch, functionality has changed from the `main` branch: we used to have smooth scrolling, but we no longer do.

adamzap:
@FreakPirate Are you still interested in working on this?

FreakPirate:
Yeah sorry. Will push an update ASAP. I usually squash changes into a single commit since the change is too small.

FreakPirate:
Thanks a lot @adamzap for overseeing, even when its such a small change.

@bmispelon probably need an approval from you as well. Thank you

marksweb (approved):
This looks rather simple after a lot of discussion, but LGTM 👍🏼

sabderemane (approved):
Thank you @FreakPirate for this improvement! Welcome on board 🎸

PR #1632, opened 2024-09-26
Specify currency choices for MoneyField fields

### LINKED ISSUE #1597: Problem updating babel dependency
Looks like the migration check breaks when updating `babel`: https://github.com/django/djangoproject.com/pull/1593

I haven't had time to investigate, but it seems that upgrading `babel` triggers some changes in some models in the `foundation` app (from the CI failure in the PR linked above):
```
 Migrations for 'foundation':
  foundation/migrations/0005_alter_approvedgrant_amount_currency_and_more.py
    - Alter field amount_currency on approvedgrant
    - Alter field treasurer_balance_currency on meeting
 ```

We have a check that triggers a failure when such changes happen, which is why the PR has a red mark.

I don't have time to look at it right now, so if anyone wants to dig into the issue, they are most welcome.

### PR DESCRIPTION
**Pull request by t-payne**
If not specified in `settings.py` or passed in when defining a `MoneyField`,  [django-money](https://github.com/django-money/django-money) will generate a default list of `currency_choices`. The default list is a large list of currency codes and names. `django-money` uses [py-moneyed](https://github.com/py-moneyed/py-moneyed) to generate this list, which in turn uses [Babel](https://github.com/python-babel/babel) for localization (for the currency name). 

This meant that when `Babel` was updated, the new version was causing differences in the generated `currency_choices` list for the two `MoneyField` fields.

This PR fixes this issue by using `USD` by default for `MoneyField` fields, and then passing in a custom list when currencies other than `USD` are needed.

Fixes #1597

### CODE REVIEW

File: foundation/migrations/0006_alter_approvedgrant_amount_and_more.py [hunk 1]
```diff
@@ -0,0 +1,30 @@
+# Generated by Django 4.2.14 on 2024-09-26 11:12
```
bmispelon, line 1: (minor) I'd rename the migration file to something like `0006_hardcode_currency_choices.py` just to be a bit more clear about what the migration does in practice.

foundation/models.py
```diff
@@ -131,6 +132,11 @@ class ApprovedGrant(models.Model):
         decimal_places=2,
         default_currency="USD",
         default=Decimal("0.0"),
+        currency_choices=[
```
bmispelon, line 135: (minor) Can we have a comment like `list of currencies extracted from current usage` to explai why we use a non-defaut list here?

bmispelon (changes requested):
Looks great, thanks for taking a stab at it! I had two minor suggestions but other than that the code looks great 👍🏻

bmispelon:
Great work, thanks again! 🚀

PR #1642, opened 2024-10-03
Fixed `isort` and `flake8` errors and ran all `tox` envs in CI

tobiasmcnulty:
The `tox` and `pre-commit` configs for `flake8` and `isort` don't match exactly (though they appear to be compatible), so we should keep the pre-commit.ci check and also run tox with all envs in GitHub Actions to avoid additional errors.


pauloxnet:
For some reason `isort` and `flake8` are missing int the requirements, I think this is the right PR to add them.

```patch
diff --git a/requirements/tests.txt b/requirements/tests.txt
index 971f973c..256e28d1 100644
--- a/requirements/tests.txt
+++ b/requirements/tests.txt
@@ -1,4 +1,5 @@
 -r dev.txt
 coverage==7.6.1
+flake8==7.1.1  
+isort==5.13.2
 requests-mock==1.12.1
 tox<5.0.0
```

tobiasmcnulty:
🤔 I just noticed that `tests.txt` and `tests.yml` install different versions of tox, though. I'll clean that up.

pauloxnet (approved):
LGTM

PR #1645, opened 2024-10-04
Improve pre-commit configuration hooks

pauloxnet:
I improve `pre-commit` configuration hooks: added new hooks, fixed a deprecated one and updated to the latest version that support Python 3.8.


fundraising/test_data/empty_payment.json
```diff
@@ -7,7 +7,6 @@
     "object": ""
   },
   "livemode": true,
-  "object": "event",
```
tobiasmcnulty, line 10: Is this change related to the `pre-commit` improvements?

tobiasmcnulty (approved):
One question, then 🚢

fundraising/test_data/empty_payment.json
pauloxnet: yes, a new JSON-related hook found this duplicated key in this JSON file.

.pre-commit-config.yaml
```diff
@@ -28,11 +40,11 @@ repos:
           - id: django-upgrade
             args: [--target-version, "4.2"]
     - repo: https://github.com/psf/black
-      rev: 24.8.0
+      rev: 24.10.0
       hooks:
           - id: black
-    - repo: https://github.com/pre-commit/mirrors-prettier
-      rev: "v4.0.0-alpha.8"
+    - repo: https://github.com/rbubley/mirrors-prettier
```
bmispelon, line 46: How much can we trust that this random person's fork will stay maintained? I understand that the official plugin was archived, but maybe we should move to a different plugin if that's the case. This new maintainer (`rbubley`) doesn't seem to have any association to the project.

.pre-commit-config.yaml
```diff
@@ -28,11 +40,11 @@ repos:
...
+      rev: "v3.3.3"
```
bmispelon, line 47: What's the reasoning behind the downgrade?

bmispelon (commented):
Thanks for that! ✨ 

I'm also curious if you could explain how you decided which new rules and plugins to add.

.pre-commit-config.yaml
pauloxnet: the previous version used was buggy and incorrectly updated so I rolled back the version to the latest stable available.

.pre-commit-config.yaml
pauloxnet: This fork is the most used since the original repository was archived. I have been using it in the last few months with success. I think it was a mistake to continue using the version archived 6 months ago, and at the moment the most painless choice is this fork. If we then wanted to decide to change formatter I would open an issue and understand which one, but I would not leave a broken and unmaintained alpha version here.

pauloxnet:
> Thanks for that! ✨
> 
> I'm also curious if you could explain how you decided which new rules and plugins to add.

I tried locally all the official [pre-commit hooks](https://github.com/pre-commit/pre-commit-hooks) and added in our precommit configuration only the ones that were applicable or fixed some issues. BTW I used most of them in my day by day job and they work.

I removed the general exclusion of migrations, and left for the moment only for black to avoid committing too many changes in this PR.  

I also added a new hook for the pyporoject.toml and the makefile.

pauloxnet:
@bmispelon I updated a couple of packages subversion that have been published in the last two weeks, I see Tobias already approved this PR, I tried to answer your questions. Do you still have some doubts about this PR?

marksweb (approved):
Happy with this 👍🏼

thibaudcolas:
woop!

.pre-commit-config.yaml
```diff
@@ -1,38 +1,50 @@
...
           - id: debug-statements
           - id: detect-private-key
           - id: end-of-file-fixer
             exclude_types: [json, sql]
           - id: file-contents-sorter
             files: ^(requirements/\w*.txt)$
             args: ["--ignore-case", "--unique"]
-          - id: fix-encoding-pragma
-            args: ["--remove"]
+          - id: fix-byte-order-marker
```
ulgens, line 26: @pauloxnet Was this added as `fix-encoding-pragma`s replacement?

PR #1656, opened 2024-10-08
[wip] Added basic API endpoint to get ticket info

bmispelon:


pauloxnet:
Can we create an issue for a new implementation like this one. Maybe with an Open API contract.

bmispelon:
> Can we create an issue for a new implementation like this one. Maybe with an Open API contract.

This a proof of concept and not ready for anything final yet. Discussion is happening over there: https://github.com/django/code.djangoproject.com/pull/198

Issue #1648, opened 2024-10-06
Proposal to add translation support for non-docs pages of djangoproject.com


jacklinke:
While Django’s documentation is translated into several languages, the main entry-point to the framework — the djangoproject.com website itself — remains mostly untranslated. Ensuring djangoproject.com is accessible to non-English speakers is important for demonstrating Django’s inclusivity and supporting a broad user base.

- *Please let me know if this proposal misses any existing discussions or considerations.*
- *If this should be addressed to the [`Internals - Internationalization` forum](https://forum.djangoproject.com/c/internals/i18n/14), please let me know, but the technical portion of this proposal (adding translation templatetags) at least seemed more appropriate to post here.*

This issue did not seem to be identified as a problem in the recent User research report ([1496](https://github.com/django/djangoproject.com/issues/1496))

## Discussion

It feels a bid odd to see that the docs are translated into a variety of languages, but the main entry-point to those docs - djangoproject.com itself - is only available to English speakers.

Django strongly promotes internationalization, with clear documentation on using trans and blocktrans tags, and even shows an example of using `{% blocktrans %}` and `{% trans %}` templatetags on the [Start page](https://www.djangoproject.com/start/), but the page itself is only available in English. ([ref](https://github.com/django/djangoproject.com/blob/main/djangoproject/templates/start.html))

One issue that we will run into is the content of djangoproject.com is a mix of text in templates and text that comes from models. While the models-based text could be a problem for another day, there are concrete steps we could take today to work toward translation of the template-based text.

## Questions

- Was it a conscious decision not to translate the (non-docs) pages of djangoproject.com? If so, does the reason behind that decision still stand?
- Would it be beneficial to start work on this?
- If we move forward with this, would we prefer using a service like Transifex for translating non-documentation pages, similar to how documentation is currently handled, or would it be more appropriate to stick to the existing .po files?

## Proposed Work

- [x] Identify which templates already have translation tags and which do not
- [ ] Add translation tags to the remainder of templates and submit PR
- [ ] Begin working with translators to translate 

## Templates with at least some `trans` or `blocktrans` tags

Approximate percentage of tags present on each template in the repo compared to potential `trans` or `blocktrans` opportunities.

| Template                                                                                                   | Completion |
| ---------------------------------------------------------------------------------------------------------- | ---------- |
| /dashboard/templates/dashboard/index.html                                                                  | 100%       |
| /djangoproject/templates/members/corporate_members_join_thanks.html                                        | 100%       |
[...]

PR #1661, opened 2024-10-13
feat: Translation preparation for accounts app

marksweb:
This begins the implementation of translation tags as suggested by [#1648](https://github.com/django/djangoproject.com/issues/1648)

Translations tags are added to the accounts templates and gettext added to the placeholders for `ProfileForm`.

### DIFF
--- a/djangoproject/templates/accounts/user_profile.html
+++ b/djangoproject/templates/accounts/user_profile.html
@@ -1,40 +1,40 @@
 {% extends "base_community.html" %}
-{% load humanize %}
+{% load humanize i18n %}
 
 {% block title %}{% firstof user_obj.profile.name user_obj.username %}{% endblock %}
 
 {% block content-related %}
-    <h1 class="visuallyhidden">Additional Information</h1>
+    <h1 class="visuallyhidden">{% translate "Additional Information" %}</h1>
     <div role="complementary">
         <h2>
             {% if user_obj == user %}
-                This is you!
+                {% translate "This is you!" %}
             {% else %}
-                Is this you?
+                {% translate "Is this you?" %}
             {% endif %}
         </h2>
 
-        <p>Need to edit something? Here's how:</p>
+        <p>{% translate "Need to edit something? Here's how:" %}</p>
         <ul>
-            <li><a href="{% url 'edit_profile' %}">Edit your name and email here.</a></li>
-            <li>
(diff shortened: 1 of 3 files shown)


djangoproject/templates/accounts/user_profile.html
```diff
@@ -1,39 +1,39 @@
...
                 at <a href="https://en.gravatar.com/">Gravatar</a>. If you see a robot, that's
                 because you don't have a Gravatar yet.
-                (Robots provided by <a href="https://robohash.org/">Robohash</a>.)
+                (Robots provided by <a href="https://robohash.org/">Robohash</a>.){% endblocktranslate %}
             </li>
-            <li>
+            <li>{% blocktranslate trimmed %}
                 The rest of the data is read-only for the time being. If you see
-                outrageous errors, please email <kbd>jacob</kbd> @ this domain.
+                outrageous errors, please email <kbd>jacob</kbd> @ this domain.{% endblocktranslate %}
```
bmispelon, line 29: This is out of scope for this PR, but we should probably change that, I don't think Jacob wants to personally receive emails like these anymore 😅 

(I'll file an issue later for this, unless you beat me to the punch)

bmispelon (approved):
Looks great, thanks for taking this on 💪🏻 

Just so I get this right, these are all the changes needed for the `accounts` app, is that right? If so, then my plan is to squash all the commits and merge this whole PR as one. Does that make sense to you?

djangoproject/templates/accounts/user_profile.html
marksweb: Ah I missed this though I did see another mention of him being lazy 😂

I'm away from my laptop for a while so you might beat me to it. Feel free to assign to me if you raise it first 👍

bmispelon: I've pinged him on mastodon to see how he feels about it: https://mastodon.social/deck/@bmispelon/113305480588627305

marksweb:
@bmispelon excellent, I'll keep going through the apps. I've got a PR for agregator open as well. So that matches my thoughts that we end up with a squashed commit for each app that makes their user facing strings ready for languages to be added in.

After I've done each app I'll do Django settings, i18n url patterns and the frontend language selector as a final PR.

I'm not working this week so I imagine I'll get through the lot 🤞

bmispelon:
This needs fixing after #1676 got merged

bmispelon (approved):

PR #1662, opened 2024-10-13
feat: Translation preparation for aggregator app

marksweb:
This begins the implementation of translation tags as suggested by [#1648](https://github.com/django/djangoproject.com/issues/1648)

Translations tags & gettext are added to the aggregator app with this change.

### DIFF
--- a/aggregator/views.py
+++ b/aggregator/views.py
@@ -2,6 +2,7 @@ from django.contrib import messages
 from django.contrib.auth.decorators import login_required
 from django.shortcuts import get_object_or_404, redirect, render
 from django.views.generic.list import ListView
+from django.utils.translation import gettext_lazy as _
 
 from .forms import FeedModelForm
 from .models import APPROVED_FEED, Feed, FeedItem, FeedType, LocalDjangoCommunity
@@ -70,8 +71,8 @@ def add_feed(request, feed_type_slug):
         messages.add_message(
             request,
             messages.INFO,
-            "Your feed has entered moderation. "
-            "Please allow up to 1 week for processing.",
+            _("Your feed has entered moderation. "
+            "Please allow up to 1 week for processing."),
         )
         return redirect("community-index")
 
--- a/aggregator/views.py
+++ b/aggregator/views.py
@@ -1,7 +1,7 @@
 from django.contrib import messages
 from django.contrib.auth.decorators import login_required
 from django.shortcuts import get_object_or_404, redirect, render
-from django.utils.translation import gettext_lazy as _
+from django.utils.translation import gettext as _
 from django.views.generic.list import ListView
 
 from .forms import FeedModelForm
--- a/aggregator/views.py
+++ b/aggregator/views.py
@@ -1,8 +1,8 @@
 from django.contrib import messages
 from django.contrib.auth.decorators import login_required
 from django.shortcuts import get_object_or_404, redirect, render
-from django.views.generic.list import ListView
 from django.utils.translation import gettext_lazy as _
+from django.views.generic.list import ListView
 
 from .forms import FeedModelForm
 from .models import APPROVED_FEED, Feed, FeedItem, FeedType, LocalDjangoCommunity
@@ -71,8 +71,10 @@ def add_feed(request, feed_type_slug):
         messages.add_message(
             request,
             messages.INFO,
-            _("Your feed has entered moderation. "
-            "Please allow up to 1 week for processing."),
+            _(
+                "Your feed has entered moderation. "
+                "Please allow up to 1 week for processing."
+            ),
         )
         return redirect("community-index")
 
--- a/djangoproject/templates/aggregator/feeditem_list.html
+++ b/djangoproject/templates/aggregator/feeditem_list.html
@@ -1,43 +1,47 @@
 {% extends "base_community.html" %}
+{% load i18n %}
 
 {% block content %}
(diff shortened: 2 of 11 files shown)


djangoproject/templates/aggregator/feeditem_list.html
```diff
@@ -1,22 +1,26 @@
...
+    This page, updated regularly, aggregates {{ name }}
+    from the Django community.{% endblocktranslate %}
   </h2>
 
   <ul class="list-news">
     {% for item in object_list %}
       <li>
         <h2><a href="{{ item.link }}">{{ item.title }}</a></h2>
-        <span class='meta'>Posted on {{ item.date_modified|date:"F j, Y" }} at {{ item.date_modified|date:"g:i A" }} by <a href="{{ item.feed.public_url }}">{{ item.feed.title }}</a> <a class="rss" href="{{ item.feed.feed_url }}">RSS</a></span>
+        <span class='meta'>{% blocktranslate trimmed with date_modified=item.date_modified|date:"F j, Y" time_modified=item.date_modified|date:"g:i A" public_url=item.feed.public_url title=item.feed.title feed_url=item.feed.feed_url %}
```
bmispelon, line 17: The particular formats (`F j, Y` and `g:i A`) used here are very english/US-centric and should vary based on locale. Is there an easy way to do that?

Just spitballing here, but what about a new `localized_date_modified` property on the `FeedItem` model? That would also help make this line a bit shorter.

bmispelon (changes requested):
Again, impressive work ✨ 

I found a few things that need fixing, some that can be improved, and an open question about date/time localization.

Feel free to either push new commits or amend the existing history, whatever works for you 👍🏻

aggregator/views.py
```diff
@@ -1,6 +1,7 @@
 from django.contrib import messages
 from django.contrib.auth.decorators import login_required
 from django.shortcuts import get_object_or_404, redirect, render
+from django.utils.translation import gettext_lazy as _
```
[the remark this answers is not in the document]
marksweb: Ah yes - bad habbit brought on by too much copy/paste

djangoproject/templates/aggregator/feeditem_list.html
marksweb: You can use the localised date formats that come from a `formats.py` file to allow customisation of these, if I recall.

`|date:"DATE_FORMAT"` will output Oct. 14, 2024 (think I'll use this)

`|date:"SHORT_DATE_FORMAT"` will output 10/14/2024 

Some of the named formats you can use can be found here; https://github.com/django/django/blob/main/django/utils/formats.py#L31

marksweb:
@bmispelon I've addressed your points, adding the localised `DATE_FORMAT` so that will get picked up by the translation system. And I added an extra translation from the feed iteration which gives the date and "by" feed to allow that to be translated also.

djangoproject/templates/aggregator/index.html
```diff
@@ -65,26 +66,30 @@ <h2 class="bullet-icon"><i class="icon icon-rss blue"></i> {{ feedtype.name }}</
...
             {% endfor %}
           </dl>
           <p class="meta">
             {% if latest_feeds %}
-              <a href="{% url 'community-feed-list' feedtype.slug %}">View more</a>
+              <a href="{% url 'community-feed-list' feedtype.slug %}">{% translate "View more" %}</a>
             {% endif %}
             {% if latest_feeds and feedtype.can_self_add %}
-              or
+              {% translate "or" %}
```
bmispelon, line 81: This kind of piecemeal translation might not very well for some languages I can't come up with a concrete example, but I could imagine a language where the word "or" might need to be translated differently based on context.

A solution might be a refactor into a less clever if/elif/else structure where full sentences are translated.
If that's too complicated, then at the minimum we should probably use a [contextual marker](https://docs.djangoproject.com/en/5.1/topics/i18n/translation/#contextual-markers):
```suggestion
              {% translate "or" context "view more OR add your feed" %}
```

bmispelon (changes requested):
Hi,

I think we should use readymade date/time formats everywhere when doing localization. I also spotted a potential issue with context-unaware translatable strings that I must have missed in my first round of review, sorry.

It's starting to look good though 👍🏻

djangoproject/templates/aggregator/index.html
```diff
@@ -65,26 +66,30 @@ <h2 class="bullet-icon"><i class="icon icon-rss blue"></i> {{ feedtype.name }}</
           <dl class="list-links">
             {% for item in latest_feeds %}
               <dt><a href="{{ item.link }}">{{ item.title }}</a></dt>
-              <dd>{{ item.date_modified|date:"N jS, Y \a\t P" }} by <a href="{{ item.feed.public_url }}">{{ item.feed.title }}</a></dd>
+              <dd>
+                {% blocktranslate trimmed with modified_date=item.date_modified|date:"N jS, Y \a\t P" public_url=item.feed.public_url title=item.feed.title %}
```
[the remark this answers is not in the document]
marksweb: I could add a custom format here using the `formats.py` I've added to #1669

bmispelon:
This needs fixing after #1676 got merged

marksweb:
@bmispelon Ok I rebased main & believe this is now in a good state.

bmispelon (approved):
👏🏻

PR #1669, opened 2024-10-15
feat: Translation preparation for blog app

marksweb:
This adds translations tags to the blog app as suggested in [#1648](https://github.com/django/djangoproject.com/issues/1648)

### DIFF
--- a/djangoproject/templates/blog/blog_pagination.html
+++ b/djangoproject/templates/blog/blog_pagination.html
@@ -6,7 +6,7 @@
       {% if page_obj.has_previous %}
         <li><a class="previous" href="?page={{ page_obj.previous_page_number }}">
           <i class="icon icon-chevron-left"></i>
-          <span class="visuallyhidden">{% translate "Previous" %}</span>
+          <span class="visuallyhidden">{% translate "Previous" context "pagination" %}</span>
         </a></li>
       {% endif %}
       <span class="page-current">
@@ -17,7 +17,7 @@
       {% if page_obj.has_next %}
         <li><a class="next" href="?page={{ page_obj.next_page_number }}">
           <i class="icon icon-chevron-right"></i>
-          <span class="visuallyhidden">{% translate "Next" %}</span>
+          <span class="visuallyhidden">{% translate "Next" context "pagination" %}</span>
         </a></li>
       {% endif %}
     </ul>
--- a/djangoproject/templates/blog/blog_pagination.html
+++ b/djangoproject/templates/blog/blog_pagination.html
@@ -1,19 +1,23 @@
+{% load i18n %}
+
 {% if is_paginated %}
   <div class="pagination">
     <ul class="nav-pagination" role="navigation">
       {% if page_obj.has_previous %}
         <li><a class="previous" href="?page={{ page_obj.previous_page_number }}">
           <i class="icon icon-chevron-left"></i>
-          <span class="visuallyhidden">Previous</span>
+          <span class="visuallyhidden">{% translate "Previous" %}</span>
         </a></li>
       {% endif %}
       <span class="page-current">
-        Page {{ page_obj.number }} of {{ page_obj.paginator.num_pages }}
+        {% blocktranslate trimmed with page_num=page_obj.number pages=page_obj.paginator.num_pages %}
+          Page {{ page_num }} of {{ pages }}
+        {% endblocktranslate %}
       </span>
       {% if page_obj.has_next %}
         <li><a class="next" href="?page={{ page_obj.next_page_number }}">
           <i class="icon icon-chevron-right"></i>
-          <span class="visuallyhidden">Next</span>
+          <span class="visuallyhidden">{% translate "Next" %}</span>
         </a></li>
       {% endif %}
     </ul>
--- a/djangoproject/templates/blog/entry_archive_day.html
+++ b/djangoproject/templates/blog/entry_archive_day.html
@@ -1,12 +1,13 @@
 {% extends "base_weblog.html" %}
+{% load i18n %}
 
-{% block title %}{{ day|date:"F j" }} | Weblog{% endblock %}
+{% block title %}{{ day|date:"F j" }} | {% translate "Weblog" %}{% endblock %}
 
-{% block og_title %}Django news: {{ day|date:"F j" }} archive{% endblock %}
+{% block og_title %}{% blocktranslate %}Django news: {{ day|date:"F j" }} archive{% endblocktranslate %}{% endblock %}
 
 {% block content %}
 
-  <h1>{{ day|date:"F j" }} archive</h1>
+  <h1>{% blocktranslate %}{{ day|date:"F j" }} archive{% endblocktranslate %}</h1>
 
   <ul class="list-news">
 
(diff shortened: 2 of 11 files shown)


djangoproject/templates/blog/blog_pagination.html
```diff
@@ -1,19 +1,23 @@
+{% load i18n %}
+
 {% if is_paginated %}
   <div class="pagination">
     <ul class="nav-pagination" role="navigation">
       {% if page_obj.has_previous %}
         <li><a class="previous" href="?page={{ page_obj.previous_page_number }}">
           <i class="icon icon-chevron-left"></i>
-          <span class="visuallyhidden">Previous</span>
+          <span class="visuallyhidden">{% translate "Previous" %}</span>
```
bmispelon, line 9: I would add a contextual marker here:
```suggestion
          <span class="visuallyhidden">{% translate "Previous" context "pagination" %}</span>
```

djangoproject/templates/blog/blog_pagination.html
```diff
@@ -1,19 +1,23 @@
...
-        Page {{ page_obj.number }} of {{ page_obj.paginator.num_pages }}
+        {% blocktranslate trimmed with page_num=page_obj.number pages=page_obj.paginator.num_pages %}
+          Page {{ page_num }} of {{ pages }}
+        {% endblocktranslate %}
       </span>
       {% if page_obj.has_next %}
         <li><a class="next" href="?page={{ page_obj.next_page_number }}">
           <i class="icon icon-chevron-right"></i>
-          <span class="visuallyhidden">Next</span>
+          <span class="visuallyhidden">{% translate "Next" %}</span>
```
bmispelon, line 20:
```suggestion
          <span class="visuallyhidden">{% translate "Next" context "pagination" %}</span>
```

djangoproject/templates/blog/entry_archive_day.html
```diff
@@ -1,12 +1,13 @@
 {% extends "base_weblog.html" %}
+{% load i18n %}
 
-{% block title %}{{ day|date:"F j" }} | Weblog{% endblock %}
+{% block title %}{{ day|date:"F j" }} | {% translate "Weblog" %}{% endblock %}
```
bmispelon, line 4: As I mentionned in another PR, I think we should move to using named formats rather than hardcoded ones for localized strings:
```suggestion
{% block title %}{{ day|date:"MONTH_DAY_FORMAT" }} | {% translate "Weblog" %}{% endblock %}
```
(there's several instances of this in this file that need to be fixed too)

djangoproject/templates/blog/entry_archive_month.html
```diff
@@ -1,12 +1,13 @@
 {% extends "base_weblog.html" %}
+{% load i18n %}
 
-{% block title %}{{ month|date:"F" }} | Weblog{% endblock %}
+{% block title %}{{ month|date:"F" }} | {% translate "Weblog" %}{% endblock %}
```
bmispelon, line 4: I'd use `YEAR_MONTH_FORMAT` consistently in this file:
```suggestion
{% block title %}{{ month|date:"YEAR_MONTH_FORMAT" }} | {% translate "Weblog" %}{% endblock %}
```

djangoproject/templates/blog/entry_detail.html
```diff
@@ -1,14 +1,18 @@
 {% extends "base_weblog.html" %}
+{% load i18n l10n %}
 
 {% block title %}{{ object.headline|escape }} | Weblog{% endblock %}
 
 {% block og_title %}{{ object.headline|safe }}{% endblock %}
-{% block og_description %}Posted by {{ object.author }} on {{ object.pub_date|date:"F j, Y" }}{% endblock %}
+{% block og_description %}{% blocktranslate trimmed with author=object.author pub_date=object.pub_date %}
+    Posted by {{ author }} on {{ pub_date|date:"F j, Y" }}
```
bmispelon, line 8: The filtering needs to move to the `blocktranslate` declaration (and we should use a named format)

djangoproject/templates/blog/entry_detail.html
```diff
@@ -1,14 +1,18 @@
...
 {% block content %}
     <h2>{{ object.headline|safe }}</h2>
 
-    <span class="meta">Posted by <strong>{{ object.author }}</strong> on {{ object.pub_date|date:"F j, Y" }} {% comment %}in <a href="#">Release Announcements</a>{% endcomment %}</span>
-
-    {{ object.body_html|safe }}
+    <span class="meta">{% blocktranslate trimmed with author=object.author pub_date=object.pub_date|date:"DATE_FORMAT" %}
+        Posted by <strong>{{ author }}</strong> on {{ pub_date }} </span>
+    {% endblocktranslate %}
+        {{ object.body_html|safe }}
```
bmispelon, line 17: I think this line got indented by mistake.

Also - and this is definitely out of scope for this PR but I was wondering if you'd thought of that already - data that comes from the database such as `body_html` here should be marked with the correct `lang` HTML attribute (and we should probably store which language an entry was written in on the model).

bmispelon (changes requested):

djangoproject/templates/blog/blog_pagination.html
marksweb: @bmispelon Yeah, good idea - I don't use context enough with translation!!

djangoproject/templates/blog/entry_detail.html
marksweb: Ah missed this one 🤦🏼

djangoproject/templates/blog/entry_detail.html
marksweb: Well translating database objects is a decision that needs making as to a third party package to utilise I think.

An issue with this though is you - at least in my experience - don't get content exported to your `.po` files from the usual packages so you need to do the translations in admin. But maybe if we're storing html/markdown then files can be shared with translators and then entered into the system by someone with access.

_(I work with someone who wrote a management command that extracts database data and writes it to html files so that it can be picked up by regular translation methods)_

marksweb:
@bmispelon Bit of scope creep at the 11th hour here, but I've added a formats file to allow us to customise the datetime formats.

With this, it'd be good to get this merged before others so that I can then start using named date formats on the other apps.

marksweb:
> I don't think adding custom formats works

Oh boy. Here's my first 🤦🏼 moment on this project.

So I guess that means you can only customise the format of pre-defined formats. I'll just select the most suitable format in each instance then.

bmispelon:
Here's hoping to many more 🤦🏻 moments, that just means we're doing exciting things 😁 

I didn't know before reviewing your PRs that you could use named formats, so I'm happy to be learning new things.

djangoproject/templates/blog/entry_archive_year.html
```diff
@@ -1,12 +1,13 @@
 {% extends "base_weblog.html" %}
+{% load i18n %}
 
-{% block title %}{{ year|date:"Y" }} | Weblog{% endblock %}
+{% block title %}{{ year|date:"YEAR_MONTH_FORMAT" }} | {% translate "Weblog" %}{% endblock %}
```
bmispelon, line 4: I don't think that's the right named format to use here, I'd just keep using the year:
```suggestion
{% block title %}{{ year|date:"Y" }} | {% translate "Weblog" %}{% endblock %}
```

(also in the other two instances later in the file)

bmispelon (changes requested):
Almost there, I think we're done to one final template! 🏁

bmispelon (approved):
Thanks! 🎸

PR #1670, opened 2024-10-15
feat: Translation preparation for the contact app 

marksweb:
This adds translations tags to the contact app as suggested in [#1648](https://github.com/django/djangoproject.com/issues/1648).

### DIFF
--- a/djangoproject/templates/contact/coc_form.html
+++ b/djangoproject/templates/contact/coc_form.html
@@ -1,22 +1,23 @@
+{% load i18n %}
 <form action="{% url "contact_coc" %}" method="post" accept-charset="utf-8" class="wide">
   {% csrf_token %}
 
 {# hardcoding inputs because they'll be included on the CoC page itself #}
   <p>
-    <label for="id_name">Your name (optional):</label>
+    <label for="id_name">{% translate "Your name (optional):" %}</label>
     {% if form.name.errors %}<p class="errors">{{ form.name.errors.as_text }}</p>{% endif %}
     <input id="id_name" class="required" type="text" name="name">
   </p>
   <p>
-    <label for="id_email">Your email address (optional):</label>
+    <label for="id_email">{% translate "Your email address (optional):" %}</label>
     {% if form.email.errors %}<p class="errors">{{ form.email.errors.as_text }}</p>{% endif %}
     <input id="id_email" class="required" type="text" name="email">
   </p>
   <p>
-    <label for="id_body">Your message:</label>
+    <label for="id_body">{% translate "Your message:" %}</label>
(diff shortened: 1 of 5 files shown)


djangoproject/templates/contact/coc_form.html
```diff
@@ -1,22 +1,23 @@
...
     <input id="id_email" class="required" type="text" name="email">
   </p>
   <p>
-    <label for="id_body">Your message:</label>
+    <label for="id_body">{% translate "Your message:" %}</label>
     {% if form.body.errors %}<p class="errors">{{ form.body.errors.as_text }}</p>{% endif %}
     <textarea id="id_body" class="required" rows="10" name="body" cols="40"></textarea>
   </p>
-  <p class="submit"><input type="submit" value="Send &rarr;"></p>
+  <p class="submit"><input type="submit" value="{% translate 'Send' %} &rarr;"></p>
```
bmispelon, line 21: I'm not super well versed in that aspect of localization, but I think the `&rarr;` should be part of the translatable string so that locales that don't use left-to-right writing systems could adapt.

djangoproject/templates/contact/foundation.html
```diff
@@ -31,6 +48,6 @@ <h1>Contact the Django Software Foundation</h1>
       {{ form.body }}
     </p>
     <p>{{ form.captcha }}</p>
-    <p class="submit"><input type="submit" class="cta" value="Send &rarr;"></p>
+    <p class="submit"><input type="submit" class="cta" value="{% translate 'Send' %} &rarr;"></p>
```
bmispelon, line 51: Same not about `&rarr;` being part of the translatable string.

bmispelon (commented):
I feel I'm repeating myself a bit, but I'll say it again: top-notch work Mark 👍🏻 

I had one minor question about the inclusion of `&rarr;` in the string or not, otherwise this looks great.

djangoproject/templates/contact/foundation.html
marksweb: Ah yes - this should be in the translation because the arrow might not apply/make sense to all locales - good catch. That's just me seeing a work and translating it in a hurry

marksweb: Screenshot 2024-10-20 at 02 22 38

Here's an example of a left arrow instead 🎉

marksweb:
> I feel I'm repeating myself a bit, but I'll say it again: top-notch work Mark 👍🏻 
> 
> I had one minor question about the inclusion of `&rarr;` in the string or not, otherwise this looks great.

Sorry thought I replied to this. I've included the entity ref incase it's appropriate for other language to change it or position it before the word. Thinking mostly RTL languages with that.

bmispelon (approved):
🚀

PR #1671, opened 2024-10-15
feat: Translation preparation for the dashboard app

marksweb:
This adds translations tags to the dashboard app as suggested in [#1648](https://github.com/django/djangoproject.com/issues/1648)

### DIFF
--- a/dashboard/views.py
+++ b/dashboard/views.py
@@ -5,6 +5,7 @@ from django.core.cache import cache
 from django.forms.models import model_to_dict
 from django.http.response import Http404, JsonResponse
 from django.shortcuts import render
+from django.utils.translation import gettext as _
 
 from .models import Metric
 from .utils import generation_key
@@ -69,4 +70,4 @@ def _find_metric_or_404(slug):
             return MC.objects.get(slug=slug)
         except MC.DoesNotExist:
             continue
-    raise Http404(f"Could not find metric with slug {slug}")
+    raise Http404(_(f"Could not find metric with slug {slug}"))
--- a/dashboard/models.py
+++ b/dashboard/models.py
@@ -7,6 +7,7 @@ from django.conf import settings
 from django.contrib.contenttypes.fields import GenericForeignKey, GenericRelation
 from django.contrib.contenttypes.models import ContentType
 from django.db import connections, models
+from django.utils.translation import gettext_lazy as _
 from django_hosts.resolvers import reverse
 
 from tracdb.models import Ticket
@@ -15,9 +16,9 @@ METRIC_PERIOD_INSTANT = "instant"
 METRIC_PERIOD_DAILY = "daily"
 METRIC_PERIOD_WEEKLY = "weekly"
 METRIC_PERIOD_CHOICES = (
-    (METRIC_PERIOD_INSTANT, "Instant"),
-    (METRIC_PERIOD_DAILY, "Daily"),
-    (METRIC_PERIOD_WEEKLY, "Weekly"),
+    (METRIC_PERIOD_INSTANT, _("Instant")),
+    (METRIC_PERIOD_DAILY, _("Daily")),
+    (METRIC_PERIOD_WEEKLY, _("Weekly")),
 )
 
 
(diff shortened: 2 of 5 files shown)


dashboard/views.py
```diff
@@ -69,4 +70,4 @@ def _find_metric_or_404(slug):
             return MC.objects.get(slug=slug)
         except MC.DoesNotExist:
             continue
-    raise Http404(f"Could not find metric with slug {slug}")
+    raise Http404(_(f"Could not find metric with slug {slug}"))
```
bmispelon, line 73: I don't think this will work, since the f-string will be evaluated before `gettext()` is called. I don't think we can use f-strings here, we need to use the old `%` or `str.format`:
```suggestion
    raise Http404(_("Could not find metric with slug %s") % slug)
```

Come to think of it, this kind of thing should have been caught by our linter, do you know if there might be some additional rules/plugins we could install that would have caught this?

bmispelon (changes requested):
🙇🏻

dashboard/views.py
marksweb: Yes good catch - since I did this I've actually read up on f strings and gettext. I'm not sure about linters catching this, but that would be nice to have.

bmispelon:
I don't really understand why, but the Python 3.12 CI is failing with this error:
```

Run coveralls --service=github
Submitting coverage to coveralls.io...
Error running coveralls: Could not submit coverage: 422 Client Error: Unprocessable Entity for url: https://coveralls.io/api/v1/jobs
Received 422 submitting job via Github Actions. By default, coveralls-python uses the "github" service name, which requires you to set the $GITHUB_TOKEN environment variable. If you want to use a COVERALLS_REPO_TOKEN instead, please manually override $COVERALLS_SERVICE_NAME to "github-actions". For more info, see https://coveralls-python.readthedocs.io/en/latest/usage/configuration.html#github-actions-support
  File "/opt/hostedtoolcache/Python/3.12.7/x64/lib/python3.12/site-packages/coveralls/api.py", line 298, in submit_report
    response.raise_for_status()
  File "/opt/hostedtoolcache/Python/3.12.7/x64/lib/python3.12/site-packages/requests/models.py", line 1024, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 422 Client Error: Unprocessable Entity for url: https://coveralls.io/api/v1/jobs

The above exception was the direct cause of the following exception:

  File "/opt/hostedtoolcache/Python/3.12.7/x64/lib/python3.12/site-packages/coveralls/cli.py", line 98, in main
    result = coverallz.wear()
             ^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.12.7/x64/lib/python3.12/site-packages/coveralls/api.py", line 275, in wear
    return self.submit_report(json_string)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.12.7/x64/lib/python3.12/site-packages/coveralls/api.py", line 301, in submit_report
    raise CoverallsException(
coveralls.exception.CoverallsException: Could not submit coverage: 422 Client Error: Unprocessable Entity for url: https://coveralls.io/api/v1/jobs
Error: Process completed with exit code 1.
```
(The 3.8 is running fine and uploading to coveralls with no apparent issue)

marksweb:
@bmispelon Hmm. Mysterious - with 3.8 working it makes me think that maybe it's a 3.12 bug. But I'd expect it to be easier to find with a search. At first I wondered if it was one of those network issues that goes away with a re-try - but it hasn't done.

I have followed some issues through to an example "good config" here; https://github.com/zhmcclient/python-zhmcclient/blob/1b91074d0be3fe7a7843a24af8abfad2cc8c1bb3/.github/workflows/test.yml#L205

To get there I started [here](https://github.com/lemurheavy/coveralls-public/issues/1709#issuecomment-1529082196) which then linked through to [here](https://github.com/lemurheavy/coveralls-public/issues/1710).

bmispelon:
I did some force-pushing to test out the config changes for coveralls, I hope I didn't break anything (the branch should be back where you left it) 🤞🏻

bmispelon (approved):

PR #1672, opened 2024-10-16
feat: Translation preparation for the project templates

marksweb:
This adds translations tags to the project level templates (those not corresponding to an app name) as suggested in [#1648](https://github.com/django/djangoproject.com/issues/1648)


bmispelon (changes requested):
Thanks again for this herculean work (and not sisyphean I hope 😁 ).

I spotted a few issues, can you also make sure the branch is rebased on the latest `main`, to make sure we've picked up all latest changes?

Thanks! ✨

djangoproject/templates/registration/password_reset_email.html
```diff
@@ -1,15 +1,16 @@
-{% load i18n %}{% autoescape off %}
-    {% blocktrans %}You're receiving this e-mail because you requested a password reset
-        for your user account at {{ site_name }}{% endblocktrans %}.
+{% load i18n %}
+{% autoescape off %}
```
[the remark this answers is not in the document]
marksweb: No, I pondered that, but didn't stop long enough to work it out.

djangoproject/templates/donate_thanks.html
```diff
@@ -1,18 +1,40 @@
...
+    {% blocktranslate trimmed %}
+      Thank you for supporting Django. If you like, you can use one of the badges below to show your support and help us bring in other donors.
+    {% endblocktranslate %}
+  </h2>
 
-  <p>Copy and paste the source below each image to link to our donation page.</p>
+  <p>{% translate "Copy and paste the source below each image to link to our donation page." %}</p>
 
-  <h3><a href="https://www.djangoproject.com/foundation/donate/"><img src="https://www.djangoproject.com/s/img/badges/djangodonated126x54.gif" border="0" alt="I donated to Django" width="126" height="54" /></a></h3>
+  <h3><a href="https://www.djangoproject.com/foundation/donate/"><img src="https://www.djangoproject.com/s/img/badges/djangodonated126x54.gif" border="0" alt="{% translate "I donated to Django" %}" width="126" height="54" /></a></h3>
```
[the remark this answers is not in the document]
marksweb: I think so. Alt text is to help people understand an image so that's what I was going for here.

djangoproject/templates/diversity/changes.html
```diff
@@ -1,39 +1,51 @@
...
+      A complete list of changes can always be found
+      <a href="https://github.com/django/djangoproject.com/commits/main/djangoproject/templates/diversity">on GitHub</a>;
+      major changes and releases are summarized below.
+    {% endblocktranslate %}
+  </p>
+
+  <h2>{% translate "Changelog" %}</h2>
 
   <dl>
     <dt>Jun 16, 2015</dt>
```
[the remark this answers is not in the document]
marksweb: Good catch. And yes I understand your point that there might be things we miss through this phase, but if we do it's easy to pick things up as they're noticed.

There's also a lot of dates like this in the releases app. I wondered the same thing but don't have a solution to localising these yet beyond translating the month. Months are already translated and I think people can make sense of the year and day if given.

djangoproject/templates/base_community.html
```diff
@@ -1,55 +1,63 @@
...
 {% block layout_class %}sidebar-right{% endblock %}
-{% block title %}Django Community{% endblock %}
+{% block title %}{% translate "Django Community" %}{% endblock %}
 
 {% block header %}
-  <h1 class="visuallyhidden">Community</h1>
-  <p>Building the <em>Django</em> Community{% if community_stats %}
-    {% if community_stats.age %} for <em>{{ community_stats.age }}</em>{% endif %}{% endif %}. Come join us!</p>
+  <h1 class="visuallyhidden">{% translate "Community" %}</h1>
+  {% spaceless %}
```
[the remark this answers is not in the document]
marksweb: Think I'll remove that. Don't think I intended for it to get committed.

djangoproject/templates/diversity/changes.html
```diff
@@ -1,39 +1,51 @@
...
+      <a href="https://github.com/django/djangoproject.com/commits/main/djangoproject/templates/diversity">on GitHub</a>;
+      major changes and releases are summarized below.
+    {% endblocktranslate %}
+  </p>
+
+  <h2>{% translate "Changelog" %}</h2>
 
   <dl>
-    <dt>Jun 16, 2015</dt>
+    <dt>{% translate "June" %} 16, 2015</dt>
```
bmispelon, line 46: This won't work either unfortunately (in french for example, the day comes before the month name).

I haven't tested it, but I think a custom tag like this should work:
```
from datetime import date

from django import template
from django.template.defaultfilters import date as datefilter

register = template.Library()


@register.simple_tag
def isodate(datestr, dateformat="DATE_FORMAT"):
    """
    Convert the given string to a date object (ISO) then format it using
    the given format (using Django's |date template filter)
    """
    d = date.fromisoformat(datestr)
    return datefilter(d, dateformat)
```
Then you can use it like so:
```suggestion
    <dt>{% isodate "2015-06-16" %}</dt>
```

bmispelon (changes requested):

djangoproject/templates/diversity/changes.html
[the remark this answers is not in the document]
bmispelon: I've made a suggestion in another thread for a custom tag solution that I think could work. Let me know what you think there.

djangoproject/templates/diversity/changes.html
marksweb: Ooo clever. I'll try that out. I was going for the "people will make sense of this" approach rather than thinking of a solution!

I'll try this out.

releases/templatetags/date_format.py
```diff
@@ -0,0 +1,16 @@
+from datetime import date
+
+from django import template
+from django.template.defaultfilters import date as datefilter
+
+register = template.Library()
+
+
+@register.simple_tag
+def isodate(datestr, dateformat="DATE_FORMAT"):
```
bmispelon, line 10: Needs some tests:
```
    def test_isodate(self):
        self.assertEqual(isodate("2005-07-21"), "July 21, 2005")

    def test_isodate_explicit_format(self):
        self.assertEqual(isodate("2005-07-21", "Ymd"), "20050721")

    @override_settings(LANGUAGE_CODE="nn")
    def test_isodate_translated(self):
        self.assertEqual(isodate("2005-07-21"), "21. juli 2005")
```
(in `releases/tests.py::TestTemplateTags` for example)

bmispelon (changes requested):
Some tests for the new template tag would be nice

releases/templatetags/date_format.py
marksweb: Yes, fair point 👍🏼  Thanks for writing some more of this 😆

releases/tests.py
```diff
@@ -65,6 +66,17 @@ def test_release_notes_1_10(self):
...
+    def test_isodate(self):
+        self.assertEqual(isodate("2005-07-21"), "July 21, 2005")
+
+    def test_isodate_explicit_format(self):
+        self.assertEqual(isodate("2005-07-21", "Ymd"), "20050721")
+        self.assertEqual(isodate("2005-07-21", "d/m/Y"), "21/07/2005")
+
+    @override_settings(LANGUAGE_CODE="nn")
+    def test_isodate_translated(self):
+        self.assertEqual(isodate("2005-07-21"), "juli 21, 2005")
```
bmispelon, line 78: Oh this is interesting. On my machine this test fails:
```
FAIL: test_isodate_translated (releases.tests.TestTemplateTags.test_isodate_translated)
  File "/home/tisba/Programming/djangoproject.com/venv/lib/python3.12/site-packages/django/test/utils.py", line 443, in inner
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/tisba/Programming/djangoproject.com/releases/tests.py", line 78, in test_isodate_translated
    self.assertEqual(isodate("2005-07-21"), "juli 21, 2005")
AssertionError: '21. juli 2005' != 'juli 21, 2005'
- 21. juli 2005
+ juli 21, 2005


```

As far as I know, the correct version should be "21. juli 2005" (see the source code for the nn format in Django: https://github.com/django/django/blob/main/django/conf/locale/nn/formats.py#L5). That means something funny is going on with the way the CI is setting up the locale for the tests.

Does that test work as-is on your own machine?

marksweb: Yeah the example you provided didn't work for me so I put in what the test compared to.

Why is there a difference though?! 😱😬

marksweb: The only thing I could think of was that the django upgrade had merged and I'd not ran this against the latest, but just rebased against `main` and cleaned out my tox dir and I'm still seeing tests pass. :needmytinfoilhatemojihere:

bmispelon: There must be a difference in the system configuration between tox/you on one side, and me on the other (I suspect I have some `nn`-related locale stuff installed on my system which might play a role here). I'm going to dig a little bit more ⛏️

marksweb: Yeah I certainly dont have anything setup for languages on my machine. It's very much vanilla macOS for an English speaking user. 

(Our education system did a terrible job of trying to teach me French and Spanish for a few years 😂)

marksweb:
@bmispelon Thanks - I've rebased and corrected an "effect" vs "affect" issue.

bmispelon (approved):
Excellent work 👍🏻

PR #1677, opened 2024-10-17
Remove expired PyCharm offer banner from homepage

jawadd:
This PR removes the PyCharm banner from the homepage as the offer expired on October 6, 2024.


bmispelon:
Hi,

Thanks for submitting this, but keeping the banner up was a deliberate decision by the fundraising people (even though as you point out, the end date has come and gone). We're waiting for their green light to turn the banner off.

Have a nice day ✨

PR #1679, opened 2024-10-17
Add community section to the homepage

thibaudcolas:
Content taken as-is from the Community page – trying to add more relevant options to stay in the loop, than the disused mailing lists. And balance the left and right columns.

Screenshot:

djangoproject homepage community section with two items


sabderemane (approved):

thibaudcolas:
Well that was fast!

marksweb:
More for me to translate! 😆

bmispelon:
Hi folks 👋🏻 
I wanted to test this locally before deploying and noticed that it looks very different for me than in the given screenshot:

Screenshot 2024-10-18 at 21-34-13 The web framework for perfectionists with deadlines Django

(the links are not side-by-side, and there's no light green background)

I would also like to see the `svg` tags moved to an included template, so we avoid duplicating the code (and so that it's easier to replace later on if needed).

You mind if I revert this so we can do a second attempt?

PR #1687, opened 2024-10-23
feat: Translation preparation for fundraising app

marksweb:
This adds translations tags to the fundraising app as suggested in [#1648](https://github.com/django/djangoproject.com/issues/1648)


djangoproject/templates/fundraising/index.html
```diff
@@ -25,64 +29,111 @@ <h1>{{ message }}</h1>
...
   <ul>
-    <li><a href="https://django.threadless.com/" target="_blank">Official merchandise store</a> -
-      Buy official t-shirts, accessories, and more to support Django.</li>
-    <li><a href="https://github.com/sponsors/django" target="_blank">Sponsor Django via GitHub Sponsors</a>.</li>
-    <li><a href="/foundation/donate/#benevity-giving">Benevity Workplace Giving
-      Program</a> - If your employer participates, you can make donations to the
-      DSF via payroll deduction.</li>
+    <li>
+      {% blocktranslate trimmed %}
+        <a href="https://django.threadless.com/" target="_blank">Official merchandise store</a> -
```
bmispelon, line 40: Out of scope for this PR, but I wonder if this is a problem with a known solution:

with the current translation infrastructure, translators have the power to do some significant  (and potentially malicious) modifications to the HTML like changing the `href` or even injecting a `<script>` tag.
Are there tools that can help with that, or is that something that needs to be verified by hand for each language/string?

djangoproject/templates/fundraising/manage-donations.html
```diff
@@ -14,27 +14,31 @@
...
         data-update-card-url="{% url 'fundraising:update-card' %}">
-    <h2>Manage your participation in the fundraising campaigns</h2>
-    <p>Information entered below will be visible on all of your donations to the Django Project.</p>
+    <h2>{% translate "Manage your participation in the fundraising campaigns" %}</h2>
+    <p>{% translate "Information entered below will be visible on all of your donations to the Django Project." %}</p>
     {% csrf_token %}
     {% include 'fundraising/includes/_form.html' with form=hero_form %}
     <p class="submit">
-      <input type="submit" class="cta" value="Save &rarr;">
+      <input type="submit" class="cta" value="{% translate "Save &rarr;" %}">
```
bmispelon, line 34: We should have a comment for translators or maybe a context marker to indicate what is being saved since that might influence the form of the verb used.

fundraising/forms.py
```diff
@@ -11,7 +12,7 @@ class DjangoHeroForm(forms.ModelForm):
     hero_type = forms.ChoiceField(
         required=False,
         widget=forms.RadioSelect,
-        label="I am donating as an",
+        label=_("I am donating as an"),
```
bmispelon, line 15: This probably won't work in every language (even in English, it's only correct now because both choices start with a vowel). But I don't have a suggestion for how to fix it, so it's probably good enough for now

fundraising/forms.py
```diff
@@ -129,7 +131,7 @@ class DonateForm(forms.Form):
         (1000, "US $1,000"),
         (1250, "US $1,250"),
         (2500, "US $2,500"),
-        ("custom", "Other amount"),
+        ("custom", _("Other amount")),
```
bmispelon, line 134: The other choice need to be marked as translatable as well

djangoproject/templates/fundraising/email/thank-you.html
```diff
@@ -1,11 +1,19 @@
-Hello!
+{% load i18n %}
```
bmispelon, line 1: This file has an `.html` extension but is clearly plaintext, no? If so then the same remarks about `trimmed` and preserving newlines apply.

bmispelon (changes requested):
The plaintext files need to be handled a bit differently since spaces and newlines there are significant (unlike in HTML).

Localization is hard work, isn't it? 😁

djangoproject/templates/fundraising/email/payment_failed.txt
```diff
@@ -1,10 +1,21 @@
-Hi {{ donation.donor.name_with_fallback }},
+{% load i18n %}
+{% blocktranslate trimmed with name=donation.donor.name_with_fallback %}
+Hi {{ name }},
+{% endblocktranslate %}
 
-This is to let you know that payment for your recurring donation to the Django Software Foundation has failed. This might be because your payment card has expired.
+{% blocktranslate trimmed %}
```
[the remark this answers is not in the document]
marksweb: Yes good point - that's my `blocktranslate` habbits showing because trimming the HTML makes for nicer `.po` files.

fundraising/forms.py
marksweb: Yeah I think it's one to address when someone has a solution

marksweb:
@bmispelon Ok, I _think_ I've addressed all the issues here - thanks for another review!

bmispelon (changes requested):
There's still an issue with plaintext templates: the `{ %blocktranslate %}` are inserting extra new lines.

Here's the code I used to test this:
```
from django.template import loader
print(loader.get_template("fundraising/email/payment_failed.txt").render({"donation":{"donor": {"pk": 1234, "name_with_fallback": "TEST"}}}))
```

If you run this code on `main` vs. `marksweb:feat/i18n-fundraising/1648` you'll see extra blank lines between paragraphs. Ideally, the output should be identical.

marksweb:
@bmispelon Ah yes, apologies - didn't realise that `blocktranslate` created space like this. 🤦🏼 

I've corrected the 3 email templates - though we are adding a blank line at the end of files so there is that still.




```
>>> from django.template import loader
>>> print(loader.get_template("fundraising/email/payment_failed.txt").render({"donation":{"donor": {"pk": 1234, "name_with_fallback": "TEST"}}}))
Hi TEST,

This is to let you know that payment for your recurring donation to the Django Software Foundation has failed.
This might be because your payment card has expired.

Please go to http://www.djangoproject.localhost:8000/fundraising/manage-donations/1234/ to add a new card. We won't be able to take any payments in the mean time.

Thanks very much for your support.

Regards,
Django Software Foundation

>>> print(loader.get_template("fundraising/email/subscription_cancelled.txt").render({"donation":{"donor": {"pk": 1234, "name_with_fallback": "TEST"}}}))
Hi TEST,

This is to let you know that your recurring payment to the Django Software
Foundation has been cancelled. No further payments will be made.

If you didn't intend to do this, please go back to the fundraising page and
recreate your payment subscription:
http://www.djangoproject.localhost:8000/fundraising/
Thanks very much for your support.

Regards,
Django Software Foundation

>>> print(loader.get_template("fundraising/email/thank-you.html").render({"donation":{"donor": {"pk": 1234,}}}))
Hello!

Thank you for making a donation to the Django Software Foundation.

Your support is invaluable to continue the rapid development of Django and helps the Django Fellowship program in particular.

If you want to change information about your donation, you can do so by accessing the link below:

http://www.djangoproject.localhost:8000/fundraising/manage-donations/1234/

Best regards,
Django Software Foundation
```

djangoproject/templates/fundraising/email/payment_failed.txt
```diff
@@ -1,10 +1,15 @@
-Hi {{ donation.donor.name_with_fallback }},
+{% load i18n %}{% spaceless %}{% url 'fundraising:manage-donations' donation.donor.pk as manage_url %}
```
bmispelon, line 1: Clever use of `spaceless` 👍🏻

bmispelon (approved):

PR #1688, opened 2024-10-23
feat: Translation preparation for releases app

marksweb:
This adds translations tags to the releases app as suggested in [#1648](https://github.com/django/djangoproject.com/issues/1648)

### DIFF
--- a/djangoproject/templates/releases/download.html
+++ b/djangoproject/templates/releases/download.html
@@ -30,7 +30,8 @@
   </p>
 
   {% release_notes current.version show_version=True as current_release_version %}
-  <h2>{% translate "Option" %} {% cycle '1' '2' '3' as options %}: {% translate "Get the latest official version" %}</h2>
+  {% cycle 1 2 3 as options silent %}
+  <h2>{% blocktranslate %}Option {{ options }}: Get the latest official version{% endblocktranslate %}</h2>
   <p>
     {% blocktranslate trimmed with version=current.version %}
       The latest official version is {{ current.version }}{% endblocktranslate %}{% if current.is_lts %} (LTS){% endif %}.
@@ -46,8 +47,11 @@
 
   {% if preview %}
     {% release_notes preview.version show_version=True as previous_release %}
+    {% cycle options %}
     {% with preview.version|slice:":3" as major_version %}
-      <h2>{% translate "Option" %} {% cycle options %}: Get the {{ preview.get_status_display }} for {{ major_version }}</h2>
+      <h2>{% blocktranslate trimmed with version_display=preview.get_status_display %}
+        Option {{ options }}: Get the latest official version {{ version_display }} for {{ major_version }}
+      {% endblocktranslate %}}</h2>
       <p>
         {% blocktranslate trimmed with preview_version=preview.version preview_status=preview.get_status_display %}
           As part of the
@@ -63,7 +67,8 @@
     {% endwith %}
   {% endif %}
 
-  <h2>{% translate "Option" %} {% cycle options %}: {% translate "Get the latest development version" %}</h2>
+  {% cycle options %}
+  <h2>{% blocktranslate %}Option {{ options }}: Get the latest development version{% endblocktranslate %}</h2>
   <p>
     {% blocktranslate trimmed %}
       The latest and greatest Django version is the one that’s in our Git repository (our revision-control system).
@@ -339,7 +344,7 @@
 
     {% donation_snippet %}
 
-    <h2>{% translate "For the impatient:" %}</h2>
+    <h3>{% translate "For the impatient:" %}</h3>
     <ul>
       <li>{% translate "Latest release:" %}
         {% include "releases/_download_links.html" with release=current %}
@@ -364,8 +369,8 @@
       {% endblocktranslate %}
     </p>
 
-    <p>{% translate "If you're just looking for a stable deployment target and don’t mind waiting for the next release, you’ll want to stick with the latest official release (which will always include detailed notes on any changes you’ll need to make while upgrading)." %}</p>
-    <h2>{% translate "Previous releases" %}</h2>
+    <p>{% translate "If you're just looking for a stable deployment target and don't mind waiting for the next release, you'll want to stick with the latest official release (which will always include detailed notes on any changes you'll need to make while upgrading)." %}</p>
+    <h3>{% translate "Previous releases" %}</h3>
     <ul>
       <li>Django {{ previous.version }}{% if previous.is_lts %} (LTS){% endif %}:
         {% include "releases/_download_links.html" with release=previous %}
--- a/djangoproject/templates/releases/download.html
+++ b/djangoproject/templates/releases/download.html
@@ -8,7 +8,7 @@
 {% block og_title %}{% translate "Download Django" %}{% endblock %}
 {% block og_image %}{% static "img/release-roadmap.svg" %}{% endblock %}
 {% block og_image_alt %}{% translate "Django's release roadmap" %}{% endblock %}
-{% block og_description %}{% blocktranslate trimmed with version=current.version %}The latest official version is {{ version }}{% endblocktranslate %}{% if current.is_lts %} (LTS){% endif %}{% endblock %}
+{% block og_description %}{% blocktranslate trimmed with version=current.version_with_lts %}The latest official version is {{ version }}{% endblocktranslate %}{% endblock %}
 {% block og_image_width %}1030{% endblock %}
 {% block og_image_height %}480{% endblock %}
 {% block og_image_type %}image/svg+xml{% endblock %}
@@ -19,6 +19,7 @@
 {% block content %}
   <h1>{% translate "How to get Django" %}</h1>
   {% url 'document-detail' lang='en' version='dev' url='faq/install' host 'docs' as faq_install_url %}
+  {% comment %}Translators: The FAQ link is in English, and the hash after the URL should not be translated{% endcomment %}
   <p>{% blocktranslate trimmed %}
     Django is available open-source under the
     <a href="https://github.com/django/django/blob/main/LICENSE">BSD license</a>.
@@ -33,10 +34,11 @@
   {% cycle 1 2 3 as options silent %}
   <h2>{% blocktranslate %}Option {{ options }}: Get the latest official version{% endblocktranslate %}</h2>
   <p>
-    {% blocktranslate trimmed with version=current.version %}
-      The latest official version is {{ current.version }}{% endblocktranslate %}{% if current.is_lts %} (LTS){% endif %}.
-    {% blocktranslate trimmed %}Read the
-      {{ current_release_version }}, then install it with
+    {% blocktranslate trimmed with version=current.version_with_lts %}
+      The latest official version is {{ version }}.
+    {% endblocktranslate %}
+    {% blocktranslate trimmed %}
+      Read the {{ current_release_version }}, then install it with
       <a href="https://pip.pypa.io/en/latest/">pip</a>:
     {% endblocktranslate %}
   </p>
@@ -86,10 +88,10 @@
   </p>
 
   <h2>{% translate "After you get it" %}</h2>
-  {% url 'document-detail' lang='en' version='stable' url='intro/install' host 'docs' as stable_url %}
+  {% url 'document-detail' lang='en' version='stable' url='intro/install' host 'docs' as install_url %}
   <p>
     {% blocktranslate trimmed %}
-      See the <a href="{{ stable_url }}">installation guide</a> for further instructions. Make sure you read the
+      See the <a href="{{ install_url }}">installation guide</a> for further instructions. Make sure you read the
       documentation that corresponds to the version of Django you’ve just installed.
     {% endblocktranslate %}
   </p>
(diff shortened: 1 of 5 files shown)


djangoproject/templates/releases/download.html
```diff
@@ -1,270 +1,319 @@
...
+      <th>{% translate "Latest Release" %}</th>
+      <th>{% translate "End of mainstream support" %}<sup><a href="#ft1">1</a></sup></th>
+      <th>{% translate "End of extended support" %}<sup><a href="#ft2">2</a></sup></th>
     </tr>
     <tr>
       <td>5.1</td>
       <td>{% get_latest_micro_release '5.1' %}</td>
-      <td>April 2025</td>
-      <td>December 2025</td>
+      <td>{% translate "April" %} 2025</td>
```
marksweb, line 137: There's lots of my poor effort at making date roughly recognisable to people.

After we're happy with #1672 then I can rebase & use the template tag in here

djangoproject/templates/releases/download.html
```diff
@@ -1,306 +1,371 @@
...
+    Here's how to get it:
+  {% endblocktranslate %}
+  </p>
 
-  <h2>Option {% cycle '1' '2' '3' as options %}: Get the latest official version</h2>
-  <p>The latest official version is {{ current.version }}{% if current.is_lts %} (LTS){% endif %}. Read the
-    {% release_notes current.version show_version=True %}, then install it with
-    <a href="https://pip.pypa.io/en/latest/">pip</a>:</p>
+  {% release_notes current.version show_version=True as current_release_version %}
+  <h2>{% translate "Option" %} {% cycle '1' '2' '3' as options %}: {% translate "Get the latest official version" %}</h2>
```
bmispelon: Ideally we should translate the word "option" and its respective number together as "option X", but the `{% cycle %}` is making it really hard.

I believe the following should work (simplified):
```
{% cycle 1 2 3 as options silent %}
<h2>{% blocktranslate %}Option {{ options }}: {% endblocktranslate %}{% translate "..." %}</h2>
...
{% if ... %}
  {% cycle options %}{# advances the option index #}
  <h2>{% blocktranslate %}Option {{ options }}: {% endblocktranslate %}{% translate "..." %}</h2>
{% endif %}
...
{% cycle options %}{# advances the option index again #}
<h2>{% blocktranslate %}Option {{ options }}: {% endblocktranslate %}{% translate "..." %}</h2>
```

djangoproject/templates/releases/download.html
```diff
@@ -1,306 +1,371 @@
...
-      <h2>Option {% cycle options %}: Get the {{ preview.get_status_display }} for {{ major_version }}</h2>
-      <p>As part of the
-        <a href="https://code.djangoproject.com/wiki/Version{{ major_version }}Roadmap">
-          Django {{ major_version }} development process</a>, Django
-        {{ preview.version }} is available. This release is only for users who
-        want to try the new version and help identify remaining bugs before the
-        {{ major_version }} release. Please read the
-        {% release_notes preview.version show_version=True %} before using this package.
-        <p>Install the {{ preview.get_status_display }} with <a href="https://pip.pypa.io/">pip</a>:</p>
+      <h2>{% translate "Option" %} {% cycle options %}: Get the {{ preview.get_status_display }} for {{ major_version }}</h2>
```
bmispelon: See previous comment about translating the `cycle`

djangoproject/templates/releases/download.html
```diff
@@ -1,306 +1,371 @@
...
+          {{ previous_release }} before using this package.
+          <p>Install the {{ preview_status }} with <a href="https://pip.pypa.io/">pip</a>:</p>
+        {% endblocktranslate %}
       <pre class="literal-block"><code>pip install --pre django</code></pre>
     {% endwith %}
   {% endif %}
 
-  <h2>Option {% cycle options %}: Get the latest development version</h2>
-  <p>The latest and greatest Django version is the one that’s in our Git repository (our revision-control system). This is only for experienced users who want to try incoming changes and help identify bugs before an official release. Get it using this shell command, which requires <a href="https://git-scm.com/">Git</a>:</p>
+  <h2>{% translate "Option" %} {% cycle options %}: {% translate "Get the latest development version" %}</h2>
```
bmispelon: Again, see comment about translating `cycle`

djangoproject/templates/releases/download.html
```diff
@@ -1,306 +1,371 @@
...
   <div role="complementary">
-    <h2 class="visuallyhidden" id="aside-header">Additional Information</h2>
+    <h2 class="visuallyhidden" id="aside-header">{% translate "Additional Information" %}</h2>
 
     {% top_corporate_members "diamond" "platinum" header="Diamond and Platinum Members" %}
 
     {% donation_snippet %}
 
-    <h3>For the impatient:</h3>
+    <h2>{% translate "For the impatient:" %}</h2>
```
bmispelon: Any particular reason why this `<h3>` became an `<h2>`?

djangoproject/templates/releases/download.html
```diff
@@ -1,306 +1,371 @@
...
+        If you choose to follow the development version, keep in mind that there will occasionally be
+        backwards-incompatible changes. You’ll want to pay close attention to the commits by watching
+        <a href="https://github.com/django/django">Django on GitHub</a> or subscribing to
+        <a href="https://groups.google.com/group/django-updates">django-updates</a>.
+      {% endblocktranslate %}
+    </p>
 
-    <p>If you’re just looking for a stable deployment target and don’t mind waiting for the next release, you’ll want to stick with the latest official release (which will always include detailed notes on any changes you’ll need to make while upgrading).</p>
-    <h3>Previous releases</h3>
+    <p>{% translate "If you're just looking for a stable deployment target and don’t mind waiting for the next release, you’ll want to stick with the latest official release (which will always include detailed notes on any changes you’ll need to make while upgrading)." %}</p>
```
bmispelon: Oh, a fellow curly-quote hater spotted! 😆

djangoproject/templates/releases/download.html
```diff
@@ -1,306 +1,371 @@
...
+    <h2>{% translate "Previous releases" %}</h2>
```
bmispelon: Another rogue h3 -> h2 conversion?

releases/templatetags/date_format.py
```diff
@@ -12,5 +12,8 @@ def isodate(datestr, dateformat="DATE_FORMAT"):
     Convert the given string to a date object (ISO) then format it using
     the given format (using Django's |date template filter)
     """
-    d = date.fromisoformat(datestr)
+    try:
```
bmispelon: Can you elaborate why this is needed?

bmispelon (commented):
I did a first pass: I just read through the code but didn't check the project out locally.

I also didn't double-check that the dates on the download page were correctly converted (I'll reserve that for a second pass).

I added a suggestion to improve the translation of the various download options, let me know if it works.

Also there seem to be some merge conflicts, could you fix those and re-request a review? Thanks!

(And as always: excellent work!)

djangoproject/templates/releases/download.html
marksweb: Likely an issue with conflict resolution where I've brought in the change that added the translation tag and missed the fact the header changed from 2 to 3 upstream.

releases/templatetags/date_format.py
marksweb: The try/except is to avoid templates failing to render.

Perhaps it could be considered acceptable and something that should get picked up through PR review, but thought I'd just put a safety net in.

djangoproject/templates/releases/download.html
marksweb: Ok, I _think_ I have done this, though it's a bit of a test of one's DTL capabilities to add translation to this template!

djangoproject/templates/releases/download.html
```diff
@@ -1,306 +1,376 @@
...
+{% block title %}{% translate "Download Django" %}{% endblock %}
 {% block layout_class %}sidebar-right{% endblock %}
 
-{% block og_title %}Download Django{% endblock %}
+{% block og_title %}{% translate "Download Django" %}{% endblock %}
 {% block og_image %}{% static "img/release-roadmap.svg" %}{% endblock %}
-{% block og_image_alt %}Django's release roadmap{% endblock %}
-{% block og_description %}The latest official version is {{ current.version }}{% if current.is_lts %} (LTS){% endif %}{% endblock %}
+{% block og_image_alt %}{% translate "Django's release roadmap" %}{% endblock %}
+{% block og_description %}{% blocktranslate trimmed with version=current.version %}The latest official version is {{ version }}{% endblocktranslate %}{% if current.is_lts %} (LTS){% endif %}{% endblock %}
```
bmispelon: We should try to have the ` (LTS)` bit translated, and be part of the same translatable bit. I suggest adding a method onto the `Release` model that encapsulates this logic:
```
    @property
    def version_with_lts(self):
        if self.is_lts:
            # Translators: Long Term Support version
            return gettext("{version} (LTS)").format(version=self.version)
        else:
            return self.version
```
(grepping for `git grep -F '(LTS)'` seems to show other places where this could be used too)

Then in the template we'd just have:

```suggestion
{% block og_description %}{% blocktranslate trimmed with version=current.version_with_lts %}The latest official version is {{ version }}{% endblocktranslate %}{% endblock %}
```

djangoproject/templates/releases/download.html
```diff
@@ -1,306 +1,376 @@
...
+        a gzipped tarball</a> of the development version. This archive is updated
+      every time we commit code.
+    {% endblocktranslate %}
+  </p>
 
-  <h2>After you get it</h2>
-  <p>See the <a href="{% url 'document-detail' lang='en' version='stable' url='intro/install' host 'docs' %}">installation guide</a> for further instructions. Make sure you read the documentation that corresponds to the version of Django you’ve just installed.</p>
-  <p>And be sure to sign up to the <a href="https://forum.djangoproject.com">Django Forum</a>, where other Django users and the Django developers themselves all hang out to help each other.</p>
+  <h2>{% translate "After you get it" %}</h2>
+  {% url 'document-detail' lang='en' version='stable' url='intro/install' host 'docs' as stable_url %}
```
bmispelon: I'd rename the variable to `install_url` to reflect the content of the link (similar to `release_process_url` which you used further down in the file):

```suggestion
  {% url 'document-detail' lang='en' version='stable' url='intro/install' host 'docs' as install_url %}
```

djangoproject/templates/releases/download.html
```diff
@@ -1,306 +1,376 @@
...
-  <h1>How to get Django</h1>
-  <p>Django is available open-source under the
+  <h1>{% translate "How to get Django" %}</h1>
+  {% url 'document-detail' lang='en' version='dev' url='faq/install' host 'docs' as faq_install_url %}
+  <p>{% blocktranslate trimmed %}
+    Django is available open-source under the
     <a href="https://github.com/django/django/blob/main/LICENSE">BSD license</a>.
     We recommend using the latest version of Python.
-    See <a href="{% url 'document-detail' lang='en' version='dev' url='faq/install' host 'docs' %}#what-python-version-can-i-use-with-django">
+    See <a href="{{ faq_install_url }}#what-python-version-can-i-use-with-django">
```
bmispelon, line 36: It's a bit problematic that the URL links to a resource in English rather than the active language, but I don't see a way to fix that for now. It wouldn't be enough to simply change the `lang` parameter in the `{% url 'document-detail' ... %}` above since the `dev` docs are not translated as far as I know).

In the meantime, we should have a [translator comment](https://docs.djangoproject.com/en/6.0/topics/i18n/translation/#comments-for-translators-in-templates) to indicate that:
1) The link is in English
2) the URL hash (`#what-python-version-can-i-use-with-django`) should not be translated

djangoproject/templates/releases/download.html
```diff
@@ -1,306 +1,376 @@
...
+    {% blocktranslate trimmed %}
+      Certain feature releases will be designated as <strong>long-term support
+        (LTS) releases</strong>. These releases will get security and data loss
+      fixes applied for a guaranteed period of time, typically three years.
+    {% endblocktranslate %}
+  </p>
+  <p>
+    {% url 'document-detail' lang='en' version='dev' url='internals/release-process' host 'docs' as release_process_url %}
+    {% blocktranslate trimmed %}
+      See the <a href="{{ release_process_url }}#supported-versions">
```
bmispelon, line 142: There should be a comment for translators here too (see above comment)

releases/templatetags/date_format.py
bmispelon: I think the tag should error out on bad input, rather than silently return nothing. Otherwise it's possible to mistype a date in a template and not realize it's not rendering correctly.

Without the try/except, a malformed date on the download page currently makes two of our tests fail which makes it very obvious that something is wrong.

If anything, I'd maybe **add** a test that simply renders the download page to make sure it doesn't error out. I know I just wrote we already have two test that render this page, but they have to do with sponsor logic, so it's not immediately obvious that they test the rendering of the download page as a side-effect.

bmispelon (commented):
I've now tested it locally, and read through the changes in more details.

I've made a few suggestions, let me know what you think.

Can you also resolve the merge conflicts (either via a merge from `main` into this branch, or via rebase + force push)?

Thanks!

djangoproject/templates/releases/download.html
marksweb: Yes, works well @bmispelon!

Confirmed the comment gets picked up for context

```
#. Translators: Long Term Support version
#: releases/models.py:324
#, python-brace-format
msgid "{version} (LTS)"
msgstr ""
```

releases/models.py
```diff
@@ -313,20 +318,28 @@ def is_dot_zero(self):
         """Return True if this is a final X.Y.0 release."""
         return self.status == "f" and self.micro == 0
 
+    @cached_property
+    def version_with_lts(self):
+        if self.is_lts:
+            # Translators: Long Term Support version
```
Stormheg, line 324: ❤️ you really help out translators with context like this!

PR #1691, opened 2024-10-24
Added debug toolbar in dev requirements.

pauloxnet:
Debug toolbar is already configured in the project, but the dependency in the requirements for local development was missing.


marksweb (approved):

bmispelon:
Nice catch, thank you both ✨

PR #1692, opened 2024-10-24
Activate Sphinx multi-process mode.

pauloxnet:
Activated Sphinx [multi-process mode](https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-j).

On my local machine I add a great improvement:

## Before
```console
$ time python -m manage update_docs --update-index --force
...
real	0m8,495s
user	0m2,904s
sys	0m0,499s
```

## After
```console
$ time python -m manage update_docs --update-index --force
...
real	3m10,627s
user	16m42,442s
sys	0m31,487s
```

### DIFF
--- a/docs/management/commands/update_docs.py
+++ b/docs/management/commands/update_docs.py
@@ -186,6 +186,8 @@ class Command(BaseCommand):
                         builder,
                         "-D",
                         "language=%s" % to_locale(release.lang),
+                        "-j",
+                        "auto",
                         "-Q" if self.verbosity == 0 else "-q",
                         str(source_dir),  # Source file directory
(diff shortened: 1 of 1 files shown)


bmispelon:
That's great idea, thanks 👍🏻 

Might be worth it to do the same for [django/django](https://github.com/django/django/blob/main/docs/Makefile#L24) if you have time.

pauloxnet:
> Might be worth it to do the same for [django/django](https://github.com/django/django/blob/main/docs/Makefile#L24) if you have time.

Done! https://github.com/django/django/pull/18720

Issue #1644, opened 2024-10-04
Drop support for Python 3.8


tobiasmcnulty:
Since adding Python 3.12 and dropping Python 3.8 support are being handled separately, I'm creating a new issue for dropping Python 3.8 support.

@bmispelon in #1574:
> Some of our dependencies have started dropping support for Python 3.8 (the one we're currently running in production).

> I don't think there's much to be done in this project's codebase (this is mostly an ops issue), but I wanted to have an issue where I can track which dependencies are being affected:

- [ ] `docutils` (+ `Sphinx`): #1571
- [ ] `pre-commit`: #1568
- [ ] `watchdog`: #1617

Other to dos:
- [ ] Update the `pyupgrade` repo in `.pre-commit-config.yaml` to target `--py312-plus`

pauloxnet:
What do you think it's the best way to drop Python 3.8 and start using Python 3.12 in production?

pauloxnet:
In the meantime Python 3.8 is now officially EOL 👇
https://discuss.python.org/t/python-3-8-is-now-officially-eol/66983

Issue #1634, opened 2024-09-27
Update Sphinx and docutils to the latest version


pauloxnet:
Since in the Django documentation requirements we are using the latest available version of Sphinx [1] we agreed with Baptiste that it is safe to also update the version we use on the site to the latest available in order to be aligned with the pipeline that generates the Django documentation.

[1] https://github.com/django/django/blob/main/docs/requirements.txt#L2

PR #1694, opened 2024-10-24
Updated Sphinx and related dependencies.

pauloxnet:
As noted in #1634, the latest version of Sphinx has been used in the Django repository for a long time without any problems.

I tested locally the documentation generation without any problems: `python -m manage update_docs --update-index --force`

Of course, this PR is blocked by the Python 3.8 removal issue #1644

### DIFF
--- a/requirements/common.txt
+++ b/requirements/common.txt
@@ -8,7 +8,7 @@ django-read-only==1.16.0
 django-recaptcha==4.0.0
 django-registration-redux==2.13
 Django==4.2.16
-docutils==0.20.1
+docutils==0.21.2
 feedparser==6.0.11
 Jinja2==3.1.4
 libsass==0.23.0
@@ -19,5 +19,5 @@ Pygments==2.18.0
 pykismet3==0.1.1
 requests==2.32.3
 sorl-thumbnail==12.11.0
-Sphinx==7.1.2
+Sphinx==8.1.3
 stripe==3.1.0


requirements/common.txt
```diff
@@ -8,7 +8,7 @@ django-read-only==1.16.0
 django-recaptcha==4.0.0
 django-registration-redux==2.13
 Django==4.2.16
-docutils==0.17.1
+docutils==0.21.2
```
marksweb, line 11: with the 0.21 release, python 3.8 was dropped which is what will be causing the test issues for 3.8.

marksweb (changes requested):
I think while we're supporting python 3.8 docutils needs pinning to [0.20.1](https://pypi.org/project/docutils/0.20.1/) because after that they drop 3.8 support.

requirements/common.txt
marksweb: With the on hold label, I guess this is waiting on #1644 

There is also #1575 which could be merged in the meantime.

pauloxnet: > With the on hold label, I guess this is waiting on #1644

Exactly, I also wrote it in the description.

BTW I'm running the doc update and index locally with the latest Sphinx version and Python 3.12 without any problems.

pauloxnet:
I rebased this PR and flagged ready for review because issue #1644 is solved.

marksweb (approved):
LGTM 🎉

bmispelon:
Thanks everyone ✨

Issue #1693, opened 2024-10-24
Docs return no search results after there are recent doc changes (but comes back later)


sarahboyce:
See: https://forum.djangoproject.com/t/docs-seem-to-go-missing-for-a-given-version-when-there-are-docs-changes/35861

We get periodic reports about the docs search stopping working:
- https://code.djangoproject.com/ticket/35861
- https://code.djangoproject.com/ticket/35710
- https://code.djangoproject.com/ticket/35860

bmispelon:
Here is how the `update_docs` management command is run on the server (with CRON):
`15 * * * * /bin/bash -c "source /home/www/www/venv/bin/activate && django-admin update_docs --settings=djangoproject.settings.prod --verbosity=2 --update-index --purge-cache | ts '[\%Y-\%m-\%d \%H:\%M:\%S]' >> /home/www/www/logs/cron_update_docs.log"`

sarahboyce:
Are there logs on how long this takes to run?
To me that looks like it refreshes everything every 15 minutes, rather than reacting to when there are docs changes (and then only updating those docs)

bmispelon:
> Are there logs on how long this takes to run? To me that looks like it refreshes everything every 15 minutes, rather than reacting to when there are docs changes (and then only updating those docs)

There are logs, let me see if I can get some numbers from them.

As far as I know, the CRON entry I posted above is the only way the docs get built. There's no push, it's all pull, and it happens every hour at 15 past (not every 15min).

nessita:
> Here is how the `update_docs` management command is run on the server (with CRON): `15 * * * * /bin/bash -c "source /home/www/www/venv/bin/activate && django-admin update_docs --settings=djangoproject.settings.prod --verbosity=2 --update-index --purge-cache | ts '[\%Y-\%m-\%d \%H:\%M:\%S]' >> /home/www/www/logs/cron_update_docs.log"`

@pauloxnet Does the above ring any bell for you?

bmispelon:
Here is the log I have for today: https://gist.github.com/bmispelon/b2a714f02daa300fa1e2dfef78ad7ebd (there's a lot of noise that seems to be caused by incomplete translated docs).

Note that this was before #1692 was deployed.

sarahboyce:
I was searching around the 15 minutes past mark and for a few minutes after - and that was fine
So I think having docs changes are related.
I wonder if the translation strings being updated is playing a part here

sarahboyce:
[...]

Issue #1650, opened 2024-10-07
Start using GeneratedField for Documents


pauloxnet:
The Document models have a JSON field with the data generated by Sphinx.

At the moment we use it only to read some value and populate the SearchVectorField used to perform the full-text search.

We can transform this field into a GeneratedField so that the database updates it automatically without needing to run a command.

In addition to that we can create other GeneratedFields to use them directly in the page rendering.

This issue is blocked by others:
- finally switch to Python 3.12
- update to Django 5.x

PR #1695, opened 2024-10-24
Add elapsed time in update index command

pauloxnet:
I moved the `transaction.atomic` from decorator to context manager and added the elapsed time information in the command output.

I used this code to perform this check on the documentation update in the issue #1693 

See https://github.com/django/djangoproject.com/issues/1693#issuecomment-2436254717

### DIFF
--- a/docs/management/commands/update_index.py
+++ b/docs/management/commands/update_index.py
@@ -1,3 +1,5 @@
+import time
+
 from django.core.management.base import BaseCommand
 from django.db import transaction
 
@@ -5,7 +7,7 @@ from ...models import Document
 
 
 class Command(BaseCommand):
-    @transaction.atomic
+
     def handle(self, *args, **options):
         """
         Update Document's search vector field in an atomic transaction.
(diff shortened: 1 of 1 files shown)


camilonova (approved):

bmispelon:
I'm a bit late to the party here (didn't have time to review this last night).

I don't really understand the reasoning behind moving the code from a decorator to a context manager, could you expand on that?

Since the PR had no tests (which I understand might be a little tricky to write), there's not really anything preventing someone from reverting to a decorator in the future. If this is about performance, a comment indicating that would be helpful.

pauloxnet:
> I don't really understand the reasoning behind moving the code from a decorator to a context manager, could you expand on that?

Following the Django documentation about ["Database transactions"](
https://docs.djangoproject.com/en/5.1/topics/db/transactions/#controlling-transactions-explicitly):

> _"Open transactions have a performance cost for your database server. To minimize this overhead, keep your transactions as short as possible."_

I wanted to leave in the transaction only the two lines of code for which it made sense to have it, leaving out the calculation of the elapsed time or the writing to the terminal.

> Since the PR had no tests (which I understand might be a little tricky to write), there's not really anything preventing someone from reverting to a decorator in the future. If this is about performance, a comment indicating that would be helpful.

I agree that a comment would be nice, and if you want I can add it, but the short-term goal is to convert the vector search field to a generated field (see  #1650), which will allow us to completely remove the need to perform this update.

PR #1699, opened 2024-10-26
Removed production Docker entrypoint

tobiasmcnulty:
In the deployed environment, these commands are now run by Ansible after the container starts to shorten the startup time. We should remove the entrypoint to avoid the false impression that this file is used.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -48,5 +48,5 @@ RUN npm install
 # copy project
 COPY . .
 
-# run docker-entrypoint.sh
-ENTRYPOINT ["./docker-entrypoint.prod.sh"]
+# ENTRYPOINT is specified only in the local docker-compose.yml to avoid
+# accidentally running it in deployed environments.
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,7 +7,7 @@ services:
             dockerfile: Dockerfile
             args:
                 - REQ_FILE=requirements/tests.txt
-        entrypoint: docker-entrypoint.dev.sh
+        entrypoint: ./docker-entrypoint.dev.sh
         command: python manage.py runserver 0.0.0.0:8000
         volumes:
             - .:/usr/src/app/
--- a/docker-entrypoint.prod.sh
+++ b/docker-entrypoint.prod.sh
@@ -1,8 +1,0 @@
-#!/bin/sh
-
-python -m manage migrate
-python -m manage compilemessages
-make compile-scss # must come before collectstatic
-python -m manage collectstatic --no-input --clear
-
-exec "$@"


pauloxnet:
LGTM
Can we also rename `docker-entrypoint.dev.sh` to `docker-entrypoint.sh`?

marksweb:
Ah, now this answers some questions I had 👍

I had considerations ahead of translations. Is the ansible code kept in another repo?

bmispelon:
> I had considerations ahead of translations. Is the ansible code kept in another repo?

Yes, but that repository is private (only ops has access to it)

bmispelon:
> Can we also rename `docker-entrypoint.dev.sh` to `docker-entrypoint.sh`?

What's the reason behid the rename? Just using a more standard name, or is there a benefit I'm missing? (I'm +1 on the rename, just trying to understand a bit better)

bmispelon (approved):
I agree that this changes should make it more clear (as seen by Mark's comment). Might be worth it to rename the file as pointed out by Paolo's but otherwise the changes look good 👍🏻

marksweb:
> Can we also rename `docker-entrypoint.dev.sh` to `docker-entrypoint.sh`?

On the other hand, keeping the name makes it clear from the filename that it's only for dev.

marksweb:
> Yes, but that repository is private (only ops has access to it)

No problem. I assumed as much.

pauloxnet:
> > Can we also rename `docker-entrypoint.dev.sh` to `docker-entrypoint.sh`?
> 
> What's the reason behid the rename? Just using a more standard name, or is there a benefit I'm missing? (I'm +1 on the rename, just trying to understand a bit better)

Having a file that ends in 'dev' makes it look like there is another file ending in 'prod', as it was until now, but we are deleting it.

tobiasmcnulty:
> Having a file that ends in 'dev' makes it look like there is another file ending in 'prod', as it was until now, but we are deleting it.

I thought about renaming the file while while making this change, but opted against it to make it quite clear that the file is not intended for use anywhere other than in local development. While I can see how the `.dev.sh` suffix implies there might be a `.prod.sh` suffix, having no suffix also implies a more generally useful `docker-entrypoint.sh`, which this file is not.

All that said, I'm okay with the risk, if the general consensus is still to rename it. The comment I added in the `Dockerfile` and the explicitly-cleared `entrypoint` in the Ansible scripts should help ensure this file is not accidentally run on production at some point in the future.

pauloxnet:
We can leave the file name as is, if this can help to avoid running it in production.

pauloxnet (approved):

PR #1701, opened 2024-10-28
Dropped support for Python 3.8

bmispelon:
I used `git grep -FIw 3.8` to find mentions of Python 3.8 and deleted everything that seemed relevant.

The tests now pass, not sure if I might have missed anything 🤔


pauloxnet (approved):
I also tried searching for "py38" and "38", but it seems you already fixed those occurrences. 👍🏻

marksweb (approved):
I've been running locally with 3.12 and all looks good 👍🏼

camilonova (approved):

PR #1707, opened 2024-10-30
Switched admin to dracula theme 🧛🏻 

bmispelon:
Could be fun for halloween

### DIFF
--- a/djangoproject/settings/common.py
+++ b/djangoproject/settings/common.py
@@ -75,6 +75,7 @@ INSTALLED_APPS = [
     "django_hosts",
     "sorl.thumbnail",
     "djmoney",
+    "django_admin_dracula",
     "django.contrib.sites",
     "django.contrib.auth",
     "django.contrib.admin",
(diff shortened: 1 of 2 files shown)


bmispelon:
@sjbitcode ping 😁

sjbitcode:
wooohooo! 🦇 I'm already hearing dark pipe organ music 🎶 🧛‍♂️

bmispelon:
Had to revert it, it seems it might not be compatible with the version of Django we're currently running (4.2) 😿

bmispelon:
Oh wait, I deployed the revert and now the theme seems to be here 🤔 
(looks like something is off with our deployment scripts)

Is our website haunted now? 👻

sjbitcode:
oh interesting, the theme should work with [Django 4.2 and above](https://github.com/sjbitcode/django-admin-dracula/blob/5b0233b1c85d39c17508f9ce0a1bf40ded7f6899/pyproject.toml#L37-L39) though

bmispelon:
> oh interesting, the theme should work with [Django 4.2 and above](https://github.com/sjbitcode/django-admin-dracula/blob/5b0233b1c85d39c17508f9ce0a1bf40ded7f6899/pyproject.toml#L37-L39) though

That's good to know, thanks! I think this is a bug in our deployment scripts, I'm looking into it 👀

PR #1713, opened 2024-11-03
Updated to watchdog 6.0.0

pauloxnet:
Tested `$ make watch-scss` locally.


bmispelon:
Thanks @pauloxnet !

I suspect that dependabot would have opened this PR tomorrow anyway (I've set it up to run weekly at noon in my timezone), I'm impressed you managed to be faster than it 😁

pauloxnet:
Dependabot don't test your command locally 😉

PR #1726, opened 2024-11-07
Add a link to Individual membership in the site footer

jonafato:
Finding information about Individual membership can be difficult if one does not already know where to look. The current path to this page from the homepage is via the footer at Learn More -> Django Software Foundation -> Individual Members. This change is in the same spirit as pull request #1461, which added a link to the Corporate membership page in the footer for similar discoverability reasons.

### DIFF
--- a/djangoproject/templates/includes/footer.html
+++ b/djangoproject/templates/includes/footer.html
@@ -29,6 +29,7 @@
             <li><a
               href="{% url 'document-detail' lang='en' version='dev' url="internals/security" host 'docs' %}#reporting-security-issues">Report
               a Security Issue</a></li>
+            <li><a href="{% url 'members:individual-members' %}">Individual membership</a></li>
           </ul>
         </div>
 


djangoproject/templates/includes/footer.html
```diff
@@ -29,6 +29,7 @@ <h2>Get Involved</h2>
             <li><a
               href="{% url 'document-detail' lang='en' version='dev' url="internals/security" host 'docs' %}#reporting-security-issues">Report
               a Security Issue</a></li>
+            <li><a href="/foundation/individual-members/">Individual membership</a></li>
```
bmispelon, line 32: I'd use the URL name here instead of hardcoding the link:
```suggestion
            <li><a href="{% url 'individual-members' %}">Individual membership</a></li>
```

bmispelon (changes requested):
I think this makes sense, especially in light of #1461 as you pointed out.

Just have a small suggestion, looks good otherwise.

Thanks for taking the time for this! ✨

jonafato:
@bmispelon Thanks! Suggestion committed.

bmispelon:
Oops sorry, that url tag should have been `{% url 'members:individual-members' %}` (I'd only skimmed the `members/urls.py` file and missed the `app_name`).

Glad the tests caught that 😅

jonafato:
Updated and squashed into a single commit. Hopefully third time's the charm.

bmispelon (approved):
Tested things locally and it works (tested both on the `www` and `docs` subdomains) 👍🏻

bmispelon:
Would you mind filing another PR for the same change on `django/code.djangoproject.com`? https://github.com/django/code.djangoproject.com/blob/main/trac-env/templates/site_footer.html#L24

(You'll have to hardcode the full URL there, as `{% url ... %}` won't work)

Thanks! 🚀

jonafato:
No problem, https://github.com/django/code.djangoproject.com/pull/227

bmispelon:
Both changes are now live, thanks again ✨

PR #1737, opened 2024-11-15
Added links for user-profile stats.

sarahboyce:


tracdb/testutils.py
```diff
@@ -51,6 +51,13 @@ def _create_db_table_for_model(model, schema_editor):
     """
     Use the schema editor API to create the db table for the given (unmanaged) model.
     """
+    with connections["trac"].cursor() as cursor:
+        tables = schema_editor.connection.introspection.table_names(cursor)
+
+    if model._meta.db_table in tables:
+        # Table already created.
+        return
```
sarahboyce, line 59: This is based off the following: https://github.com/django/django/blob/91c879eda595c12477bbfa6f51115e88b75ddf88/django/db/migrations/recorder.py#L63-L66

bmispelon (changes requested):
I like this change, it's nice to see the profile page getting some love after all this time 💚 

As I mentionned in one comment, can you try to rebase on to the latest main and see if that fixes the errors you experienced that lead to the fix in `testutils`?

Thanks!

tracdb/stats.py
```diff
@@ -36,42 +43,33 @@ def get_user_stats(username):
...
-@stat("Tickets closed")
-def tickets_closed(username):
-    # Raw query so that we can do COUNT(DISTINCT ticket).
-    q = """SELECT COUNT(DISTINCT ticket) FROM ticket_change
-           WHERE author = %s AND field = 'status' AND newvalue = 'closed';"""
-    return run_single_value_query(q, username)
+@stat("Tickets fixed")
+def tickets_fixed(username):
+    count = Ticket.objects.filter(owner=username, resolution="fixed").count()
+    link = get_trac_link(f"owner=~{username}&resolution=fixed")
```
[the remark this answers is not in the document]
sarahboyce: Ah nice!

accounts/tests.py
```diff
@@ -1,20 +1,132 @@
...
+            author="a-user",
+            field="owner",
+            oldvalue="b-user",
+            newvalue="a-user",
+            ticket=ticket_3,
+            _time=1731669560,
+        )
+
+    @override_settings(TRAC_URL="https://code.djangoproject.com/")
+    def test_user_profile(self):
```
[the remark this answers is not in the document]
sarahboyce: I kept the "tickets fixed" and "tickets opened" counts in the same test as it is nice to have multiple tickets impacting different counts, but split the rest :+1:

tracdb/stats.py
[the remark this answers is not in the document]
sarahboyce: It appears that Trac adds that when you use the filters but it is not required:
- https://code.djangoproject.com/query?owner=~sarahboyce
- https://code.djangoproject.com/query?owner=sarahboyce
So removed :+1:

djangoproject/templates/accounts/user_profile.html
```diff
@@ -43,10 +43,10 @@ <h1>
         </h1>
 
         {% if stats %}
-            <h2>{% translate "Lies, damned lies, and statistics:" %}</h2>
+            <h2>{% translate "Statistics on Django core contributions:" %}</h2>
```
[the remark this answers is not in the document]
sarahboyce: I think you jest :+1: 
But in case I should justify, I'm conscious we should avoid cultural references when dealing with a global audience with a wide range of ages groups

sarahboyce:
Thank you for the thorough review - I pushed up changes :+1:

djangoproject/templates/accounts/user_profile.html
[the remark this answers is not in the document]
bmispelon: I agree with you 🃏 

References can be fine, but I think this one is quite jarring when you don't know it, and definitely doesn't fit the tone of what we're going for.

accounts/tests.py
```diff
@@ -1,23 +1,134 @@
...
+        )
+        self.assertContains(
+            user2_response,
+            '<a href="https://code.djangoproject.com/query?'
+            'reporter=user2&desc=1&order=changetime">'
+            "Tickets opened: 4.</a>",
+            html=True,
+        )
+
+    def test_stat_tickets_triaged(self):
```
bmispelon, line 105: I'd split this test in 3 o make it clearer what we're testing (test 3 here was missing and refers to the comment in `new_tickets_reviewed()`)
```
test_stat_tickets_triaged():
    # possible values are from trac.ini in code.djangoproject.com
    1 ticket for user1, Unreviewed -> Accepted
    1 ticket for user1, Unreviewed -> Someday/Maybe
    1 ticket for user1, Unreviewed -> Ready for checkin
    1 ticket for user2, Unreviewed -> Accepted
    assert user1 has count 3
    # no assertion for user2 (it's there to make sure only user1 changes
    # are counted)

test_stat_tickets_triaged_unaccepted_not_counted():
    1 ticket Unreviewed -> Accepted
    1 ticket Accepted -> Unreviewed
    assert user1 has count 1

test_stat_tickets_triaged_ticket_accepted_multiple():
    1 ticket with 2 Unreviewed -> Accepted changes
    assert user1 has count 2
```

accounts/tests.py
```diff
@@ -1,23 +1,134 @@
...
+        ticket_1 = Ticket.objects.create()
+        ticket_2 = Ticket.objects.create()
+        ticket_3 = Ticket.objects.create()
+        accepts_ticket = {
```
bmispelon, line 109: That's clever to use a dict like this 👍🏻

bmispelon (changes requested):
Thanks for the quick update!

Most of my comments are nitpicks about the testing style (I find tests with multiple assertions confusing, but that's probably a personal taste).

The one thing that's actually a bug is the behavior change where stats with a zero count were previously hidden from the profile page but are now visible (that's caused by the `StatData` change and the following incorrect `if v == 0` in `views.get_user_stats()`).

Nice work! 🚀

accounts/tests.py
sarahboyce: The last test `test_stat_tickets_triaged_ticket_accepted_multiple` is currently not possible because `TicketChange` has been modeled with the ticket as a primary key with the comment `# XXX See note at the top about composite pk`

accounts/tests.py
```diff
@@ -1,23 +1,134 @@
...
+    def test_username(self):
+        user1_response = self.client.get(self.user1_url)
+        user2_response = self.client.get(self.user2_url)
+        self.assertContains(user1_response, "user1")
+        self.assertContains(user2_response, "user2")
+
+    def test_stat_commits(self):
+        Revision.objects.create(
+            author="user1",
+            rev="91c879eda595c12477bbfa6f51115e88b75ddf88",
```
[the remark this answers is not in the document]
sarahboyce: Sorry I picked a couple of random recent ones :sweat_smile:

accounts/tests.py
sarahboyce: I think some model factory thing would be nice in future

accounts/tests.py
bmispelon: My bad, I forgot about that 🤦🏻  (the actual table has a 3-column primary key `ticket, time, field` which is ... a choice)

accounts/tests.py
bmispelon: > I think some model factory thing would be nice in future

I tell myself this every time I write tests that involves a lot of models, but somehow it always seem like too big an investment for the current PR, and then I never get around to it. Classic sunk cost fallacy I guess ⛵

bmispelon (approved):
Great work, thanks! 👍🏻

PR #1742, opened 2024-11-18
feat: Translation preparation for members app

marksweb:
This adds translations tags to the members app as suggested in [#1648](https://github.com/django/djangoproject.com/issues/1648)

Lost this branch amongst all the others!

### DIFF
--- a/members/admin.py
+++ b/members/admin.py
@@ -4,6 +4,7 @@ from django.contrib import admin
 from django.templatetags.static import static
 from django.utils.formats import localize
 from django.utils.html import format_html
+from django.utils.translation import gettext_lazy as _
 
 from members.models import CorporateMember, IndividualMember, Invoice, Team
 
@@ -34,9 +35,9 @@ class StatusFilter(admin.SimpleListFilter):
 
     def lookups(self, request, model_admin):
         return (
-            (None, "Active"),
-            ("inactive", "Inactive"),
-            ("all", "All"),
+            (None, _("Active")),
+            ("inactive", _("Inactive")),
+            ("all", _("All")),
         )
 
     def choices(self, cl):
--- a/members/admin.py
+++ b/members/admin.py
@@ -30,7 +30,7 @@ class StatusFilter(admin.SimpleListFilter):
     Display only active members in the changelist page, by default.
     """
 
-    title = "Status"
+    title = _("Status")
     parameter_name = "status"
 
     def lookups(self, request, model_admin):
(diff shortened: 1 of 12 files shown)


members/admin.py
```diff
@@ -4,6 +4,7 @@
 from django.templatetags.static import static
 from django.utils.formats import localize
 from django.utils.html import format_html
+from django.utils.translation import gettext_lazy as _
```
bmispelon, line 7: (not the correct line, but I can't seem to add comments to unmodified lines)

More things need to be marked as translatable here:

- `StatusFilter.title`
- The `alt` text in `CorporateMemberAdmin.renewal_link()`

members/forms.py
```diff
@@ -1,14 +1,15 @@
 from django import forms
 from django.conf import settings
 from django.core.mail import send_mail
+from django.utils.translation import gettext_lazy as _
```
bmispelon, line 4: (again, marking this on the wrong line)

The various strings inside `save()` also need to be marked as translatable. There will also need to be some refactoring to avoid building the sentence piecemeal and instead have full sentences be marked as translatable (I think we did something similar in another pull request but I can't remember which one now).

members/views.py
```diff
@@ -62,7 +63,7 @@ def get_object(self):
             pk = signing.loads(self.kwargs["token"], max_age=2.592e6)
         except signing.BadSignature:
             raise Http404(
-                "No %(verbose_name)s found matching the query"
+                _("No %(verbose_name)s found matching the query")
```
bmispelon, line 66: This type of sentence building doesn't work well in many languages (since the `verbose_name` form might not be correct for the context of this sentence).

I suggest removing the clever `verbose_name` interpolation and hardcoding the model name in a regular translated string (this code seems like it's only used with `model=CorporateMember`)

bmispelon (changes requested):
Nice one 👍🏻 

Found a few missing strings and some that should be changed a bit. I wasn't sure about the `spaceless` trick in the text template but I tried it and it does seem to work, nice find (you should blog about that 😉 )

members/forms.py
marksweb: OK I've approached this email from and `if/else` on the renewal flag.

bmispelon (approved):
Great work (and great blog post too!) 🚀

marksweb:
> @marksweb I noticed that the last commit on the branch was made with a different commit than usual, was that on purpose?

Very well spotted! No it was not on purpose.

I've been changing my shell setup & things are still in flux. Though looking at my git config I'm not sure what put my personal email on django commits. Anyway - I've added a config entry.

Thank you - expert reviewing!

PR #1751, opened 2024-11-19
chore: Update messages to reflect current state of the translations

marksweb:
This is a simple `makemessages --all` to update the `.po` files to reflect the recently merged translation branches.

I'll run another of these when the remaining branches are merged, but I've made contact with Claude so I think it's worth having some refreshed files to see if they feed in to the current transifex project.

-------

I'm now a transifex maintainer and can see the repo is not integrated to the project there.
I think this is a worthwhile thing to do so we don't need to sync files another way. But I think that'll require someone with repo or org admin on here, because it requires settings access.

### DIFF
--- a/dashboard/locale/fr/LC_MESSAGES/django.po
+++ b/dashboard/locale/fr/LC_MESSAGES/django.po
@@ -11,19 +11,78 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2019-03-11 20:38-0500\n"
+"POT-Creation-Date: 2024-11-19 08:13-0600\n"
 "PO-Revision-Date: 2019-02-21 14:15+0000\n"
 "Last-Translator: Claude Paroz <claude@2xlibre.net>, 2019\n"
 "Language-Team: French (https://www.transifex.com/django/teams/17080/fr/)\n"
+"Language: fr\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
-"Language: fr\n"
 "Plural-Forms: nplurals=2; plural=(n > 1);\n"
 
+#: dashboard/models.py:19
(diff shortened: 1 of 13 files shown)


bmispelon:
I'm afraid nobody is going to really be reviewing a +20k -/ 500 pull request 😅 

What kind of feedback are you after?

One question I had is what the criteria is/was for the list of languages that are included: english, french, indonesian, polish, and portuguese (brazilian)?

marksweb:
@bmispelon I'm not expecting anybody to review. And I think the languages offered here are from the PR I put together in 2022. Perhaps back then these were the languages on offer in the docs.

This is just django doing django's thing. And at this point the languages offered aren't so important because we're not putting them out as options to users.

I'm essentially looking to get the ball rolling on translations in transifex. If we've got new `po` files then they should feed in to there.

Once we've got a pipeline to transifex I think then we review the languages on offer. I've already seen the transifex project and the list is vast.

bmispelon:
> This is just django doing django's thing. And at this point the languages offered aren't so important because we're not putting them out as options to users.

Our deployment script automatically runs `compilemessages`, doesn't that mean that Django will automatically start offering translated content to users based on their `Accept-Language` header since we use (a variant of) `LocaleMiddleware` [1]?

[1] https://github.com/django/djangoproject.com/blob/main/djangoproject/settings/common.py#L149

marksweb:
@bmispelon Yeah, `compilemessages` messages is the one you want to run during deployment. That gets the `.po` files and compiles the `.mo` files that are then used to serve the translated content.

What I'm trying to do with this is update those `.po` files that the deployment process would use for translation.

It's a development task rather than ops/deployment because it needs to happen as strings in the application change which is obviously what we've been doing lately.

The current deployments might `compilemessages` on the current `.po` files but those strings might no longer exist within the application because there hasn't been a `makemessages` process in a while to update the message catalog.

codingjoe:
Hello there, 👋 

@marksweb, how about future-proofing this? E.G. with a pre-commit hook or some other CI job that'll simply fail, if the gettext message files are not updated.

I also recommend `msgcheck` which does support pre-commit too (in its latest release).

Cheers!
Joe

PR #1752, opened 2024-11-20
Add rel attributes to next/prev links

sersorrel:
Marking up next/prev links with the [`rel="next"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel#next) and [`rel="prev"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel#prev) links allows the user agent and/or assistive technology to recognise the purpose of these links. For example, [Vimium](https://github.com/philc/vimium) provides keyboard shortcuts to go to the next/previous page, and relies partially on these attributes to do so.

### DIFF
--- a/djangoproject/templates/aggregator/feeditem_list.html
+++ b/djangoproject/templates/aggregator/feeditem_list.html
@@ -29,13 +29,13 @@
     <div class="pagination">
       <ul class="nav-pagination" role="navigation">
         {% if page_obj.has_previous %}
-          <li><a class="previous" href="?page={{ page_obj.previous_page_number }}">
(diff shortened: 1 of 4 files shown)


sersorrel:
Searching for e.g. `class="next"`, it looks like there are also spots in the `blog` and `aggregator` templates where these might be applicable – now added.

bmispelon:
I've deployed the changes, they should be live now.

Thanks again for your contribution, if you see anything else don't hesitate to open another PR (or an issue if it's a bigger topic).

🎸

PR #1753, opened 2024-11-20
funding.json (see https://floss.fund)

jacobian:


djangoproject/static/.well-known/funding-manifest-urls
```diff
@@ -0,0 +1 @@
+https://djangoproject.com/funding.json
```
bmispelon, line 1: Missing newline at the end of the file

fundraising/views.py
```diff
@@ -338,3 +340,129 @@ def checkout_session_completed(self):
...
+            "type": "organisation",
+            "role": "owner",
+            "name": "Django Software Foundation",
+            "email": "foundation@djangoproject.com",
+            "description": (
+                "Development of Django is supported by the Django Software Foundation, "
+                "a 501(c)(3) non-profit."
+            ),
+            "webpageUrl": {
+                "url": "http://djangoproject.com/foundation",
```
bmispelon, line 361:
```suggestion
                "url": "https://www.djangoproject.com/foundation",
```

fundraising/views.py
```diff
@@ -338,3 +340,129 @@ def checkout_session_completed(self):
...
+        "projects": [
+            {
+                "guid": "django",
+                "name": "Django",
+                "description": (
+                    "Django is a Python web framework that makes it possible to build "
+                    "better web apps quickly and with less code"
+                ),
+                "webpageUrl": {
+                    "url": "http://djangoproject.com/",
```
bmispelon, line 373:
```suggestion
                    "url": "https://www.djangoproject.com/",
```

fundraising/views.py
```diff
@@ -338,3 +340,129 @@ def checkout_session_completed(self):
...
+                },
+                "repositoryUrl": {
+                    "url": "http://github.com/django/django",
```
bmispelon, line 376:
```suggestion
                    "url": "https://github.com/django/django",
```

fundraising/views.py
```diff
@@ -338,3 +340,129 @@ def checkout_session_completed(self):
...
+                    "wellKnown": (
+                        "http://github.com/django/django/blob/main/static/"
```
bmispelon, line 378:
```suggestion
                        "https://github.com/django/django/blob/main/static/"
```

fundraising/views.py
```diff
@@ -338,3 +340,129 @@ def checkout_session_completed(self):
...
+                "name": f"Corporate Sponsorship - {level.title()}",
+                "description": f"Corporate Sponsoreship - {level.title()}",
+                "amount": amount,
+                "currency": "USD",
+                "frequency": "yearly",
+                "channels": [c["guid"] for c in funding["funding"]["channels"]],
+            }
+        )
+
+    # TODO: it'd be nice to do history. We do have the data, but not in the dp.com app.
```
bmispelon, line 466: That does seem like a nice feature, but a ticket would be a better place to track this (where is that history located by the way?).

djangoproject/static/.well-known/funding-manifest-urls
bmispelon, line 1: Is the point of this file to be served also by the webserver? If so, I assume it should be served under the `www` subdomain and not the `static` one, right?

bmispelon (changes requested):
Nice work, thanks! 🎸 

Some basic tests might be nice to have. I'm thinking a spot-check that the view is returning a well formed json, maybe also testing a few strategic keys of the response.

djangoproject/static/.well-known/funding-manifest-urls
jacobian: Ah yes you're right, I am missing the `www`, thanks.

jacobian:
@bmispelon aw man you're gonna make me write _tests_!?

(yeah you're right, will do)

marksweb (commented):
Besides adding some tests as Baptiste says, this is great to see! 👏🏼

adamzap:
@jacobian Are you still interested in getting this merged? If so, do you want to add tests? I could probably help if you don't have the bandwidth.

carltongibson (commented):
@jacobian this looks great. 

I just wonder, and this is your call, whether it would be worth defining attrs classes (or dataclasses?) for the JSON structures here and then using cattrs to serialise that? If this is never going to need to be updated likely not, but manually interpolating into dicts gives me the shudders. 

(I'll totally accept "not worth it" here!)

ulgens:
Is this still relevant?

PR #1757, opened 2024-11-20
Only return published blog entries.

### LINKED ISSUE #1755: Unpublished Blog Entries show up for staff in list pages
When you are logged in with staff access on the website, our draft Blog Entries show up whether they are published or not. I think we should either make it more obvious that these are DRAFT entries, or ideally we should not show them at all. The color is so subtle that it's hard to visually tell that one is grayed out and whereas one is very lightly greened out. (I can't share an example without spoiling who won an award, for example.) 

This also can create an issue if someone is screen sharing and pulls up the website, which seems to be a common pattern during the DSF Office Hours. 

**Staff should still be able to see draft posted when previewed from the admin.** So I would not remove this feature from the `BlogDateDetailView` view. 

Please see: https://github.com/django/djangoproject.com/blob/main/blog/views.py#L19-L23

marksweb:
This fixes #1755 to only return published blog entries, even for staff users.

Unpublished entries can still be previewed via the admin.

### DIFF
--- a/blog/tests.py
+++ b/blog/tests.py
@@ -187,13 +187,6 @@ class EventTestCase(DateTimeMixin, TestCase):
 
 
 class ViewsTestCase(DateTimeMixin, TestCase):
-    """
-    TODO:
-        * staff users without write permission on BlogEntry can't see unpublished
-        entries at all
-        * staff users with write permission on BlogEntry can't see unpublished
-        entries in the list, but can view the detail page
-    """
 
     def test_staff_with_write_permission_can_see_unpublished_detail_view(self):
         """
@@ -204,6 +197,15 @@ class ViewsTestCase(DateTimeMixin, TestCase):
             pub_date=self.yesterday, is_active=False, headline="inactive", slug="a"
         )
         user = User.objects.create(username="staff", is_staff=True)
+        # Add blog entry change permission
+        from django.contrib.auth.models import Permission
+        from django.contrib.contenttypes.models import ContentType
+
+        content_type = ContentType.objects.get_for_model(Entry)
+        change_permission = Permission.objects.get(
+            content_type=content_type, codename="change_entry"
+        )
+        user.user_permissions.add(change_permission)
         self.client.force_login(user)
         self.assertEqual(Entry.objects.all().count(), 1)
         response = self.client.get(reverse("weblog:index"))
@@ -222,7 +224,37 @@ class ViewsTestCase(DateTimeMixin, TestCase):
         )
         request = response.context["request"]
         self.assertTrue(request.user.is_staff)
+        self.assertTrue(request.user.has_perm("blog.change_entry"))
         self.assertEqual(response.status_code, 200)
+
+    def test_staff_without_write_permission_cannot_see_unpublished_detail_view(self):
+        """
+        staff users without write permission on BlogEntry can't see unpublished entries
+        """
+        e1 = Entry.objects.create(
+            pub_date=self.yesterday, is_active=False, headline="inactive", slug="a"
+        )
+        user = User.objects.create(username="staff-no-perm", is_staff=True)
+        # No permissions added
+        self.client.force_login(user)
+        self.assertEqual(Entry.objects.all().count(), 1)
+
+        # Test detail view for unpublished entry - should return 404
+        response = self.client.get(
+            reverse(
+                "weblog:entry",
+                kwargs={
+                    "year": e1.pub_date.year,
+                    "month": e1.pub_date.strftime("%b").lower(),
+                    "day": e1.pub_date.day,
+                    "slug": e1.slug,
+                },
+            )
+        )
+        request = response.context["request"]
+        self.assertTrue(request.user.is_staff)
+        self.assertFalse(request.user.has_perm("blog.change_entry"))
+        self.assertEqual(response.status_code, 404)
 
     def test_no_past_upcoming_events(self):
         """
--- a/blog/tests.py
+++ b/blog/tests.py
@@ -271,6 +271,58 @@ class ViewsTestCase(DateTimeMixin, TestCase):
                 self.assertEqual(response.status_code, 200)
                 self.assertQuerySetEqual(response.context["events"], [])
 
+    def test_anonymous_user_cannot_see_unpublished_entries(self):
+        """
+        Anonymous users can't see unpublished entries at all (list or detail view)
+        """
+        # Create a published entry to ensure the list view works
+        published_entry = Entry.objects.create(
+            pub_date=self.yesterday,
+            is_active=True,
+            headline="published",
+            slug="published",
+        )
+
+        # Create an unpublished entry
+        unpublished_entry = Entry.objects.create(
+            pub_date=self.tomorrow,
+            is_active=True,
+            headline="unpublished",
+            slug="unpublished",
+        )
+
+        # Test list view - should return 200 but not include the unpublished entry
+        response = self.client.get(reverse("weblog:index"))
+        self.assertEqual(response.status_code, 200)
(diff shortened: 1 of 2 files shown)


marksweb:
On a side note, my test suite is failing me and I'm not sure why.

```
ERROR: blog.tests (unittest.loader._FailedTest.blog.tests)
ImportError: Failed to import test module: blog.tests
  File "/Users/mark/.local/share/uv/python/cpython-3.12.7-macos-aarch64-none/lib/python3.12/unittest/loader.py", line 396, in _find_test_path
    module = self._get_module_from_name(name)
  File "/Users/mark/.local/share/uv/python/cpython-3.12.7-macos-aarch64-none/lib/python3.12/unittest/loader.py", line 339, in _get_module_from_name
    __import__(name)
  File "/Users/mark/dev/djangoproject.com/blog/tests.py", line 2, in <module>
    from test.support import captured_stderr
ModuleNotFoundError: No module named 'test'
```

blog/views.py
```diff
@@ -17,12 +17,10 @@ def get_allow_future(self):
         return self.request.user.is_staff
 
     def get_queryset(self):
-        if self.request.user.is_staff:
-            return Entry.objects.all()
-        else:
-            return Entry.objects.published()
+        return Entry.objects.published()
```
jefftriplett, line 20: This change is good, but I would move the deleted `get_queryset()` into the `BlogDateDetailView` view so we can still preview the Entry on the website. Otherwise, I don't think we'll be able to proofread the rst and markdown rendering since the admin doesn't have a true preview feature.

blog/views.py
```diff
@@ -17,12 +17,10 @@ def get_allow_future(self):
...
 
     def get_context_data(self, **kwargs):
+        """ """
```
pauloxnet, line 23: I would remove or add some text.

marksweb: 🤦🏼‍♂️ I was about to, then decided we all know what it does and it's not that complicated. Can't believe I missed that.

blog/views.py
marksweb: Ah, fair point - the way I read your issue I assumed that preview was handled separately.

marksweb: Nothing's ever as simple as it looks 😆

bmispelon (changes requested):
Thanks for the quick response on this, nice work 👍🏻 

If you allow me to change/extend the scope of this PR a little bit, I think it's not enough to test for `user.is_staff` but we should also make sure that the user has write permissions on the `BlogEntry` model (since this feature is meant to be used for previewing).

And there should also be some tests added (I'm suprised there weren't already, or that if there were they were still passing after this change):
- anon users can't see unpublished entries at all (list or detail)
- logged in users (non-staff) can't see unpublished entries at all
- staff users without write permission on `BlogEntry` can't see unpublished entries at all
- staff users with write permission on `BlogEntry` can't see unpublished entries in the list, but can view the detail page

jefftriplett:
@marksweb this is great. Thank you for jumping in on this.

marksweb:
> And there should also be some tests added (I'm suprised there weren't already

Yes I was surprised & expecting you to suggest this 😄

marksweb:
> Thanks for the quick response on this, nice work 👍🏻
> 
> If you allow me to change/extend the scope of this PR a little bit, I think it's not enough to test for `user.is_staff` but we should also make sure that the user has write permissions on the `BlogEntry` model (since this feature is meant to be used for previewing).
> 
> And there should also be some tests added (I'm suprised there weren't already, or that if there were they were still passing after this change):
> 
>     * anon users can't see unpublished entries at all (list or detail)
> 
>     * logged in users (non-staff) can't see unpublished entries at all
> 
>     * staff users without write permission on `BlogEntry` can't see unpublished entries at all
> 
>     * staff users with write permission on `BlogEntry` can't see unpublished entries in the list, but can view the detail page

Ok, it took longer than I expected, but I've added tests and enforced permissions.

blog/tests.py
```diff
@@ -187,6 +187,75 @@ def test_past_future_ordering(self):
 
 
 class ViewsTestCase(DateTimeMixin, TestCase):
+
```
adamzap, line 190: Could we drop this blank line to stay consistent with the rest of the file?

blog/tests.py
```diff
@@ -187,6 +187,75 @@ def test_past_future_ordering(self):
...
+    def test_staff_with_write_permission_can_see_unpublished_detail_view(self):
```
adamzap, line 191: Do you think it would be a bit better if we changed "write" to "change" in the test name and the docstring? Maybe elsewhere in this PR too? That way it would match the name of the permission. I don't feel strongly about this.

blog/tests.py
```diff
@@ -187,6 +187,75 @@ def test_past_future_ordering(self):
...
+        """
+        staff users with write permission on BlogEntry can't see unpublished entries
```
adamzap, line 193: Could we capitalize the first word in these comment to stay consistent with the rest of the file?

blog/tests.py
```diff
@@ -187,6 +187,75 @@ def test_past_future_ordering(self):
...
+        in the list, but can view the detail page
+        """
+        e1 = Entry.objects.create(
+            pub_date=self.yesterday, is_active=False, headline="inactive", slug="a"
+        )
+        user = User.objects.create(username="staff", is_staff=True)
+        # Add blog entry change permission
+        from django.contrib.auth.models import Permission
+        from django.contrib.contenttypes.models import ContentType
```
adamzap, line 202: Could we import these at the top of the file instead, or is there a side affect necessitating this?

blog/tests.py
```diff
@@ -187,6 +187,75 @@ def test_past_future_ordering(self):
...
+        """
+        e1 = Entry.objects.create(
+            pub_date=self.yesterday, is_active=False, headline="inactive", slug="a"
+        )
+        user = User.objects.create(username="staff-no-perm", is_staff=True)
+        # No permissions added
+        self.client.force_login(user)
+        self.assertEqual(Entry.objects.all().count(), 1)
+
+        # Test detail view for unpublished entry - should return 404
```
adamzap, line 242: I think comments like this are implied by the test name and the assert(s), but we can keep them if you or others find them helpful.

adamzap (changes requested):

blog/tests.py
adamzap: Ok cool!

blog/views.py
```diff
@@ -52,3 +49,12 @@ class BlogDayArchiveView(BlogViewMixin, DayArchiveView):
...
     banner_is_title = False
+
+    def get_queryset(self):
+        """Allows staff users with blog write permission to view unpublished entries"""
+        if self.request.user.is_staff and self.request.user.has_perm(
+            "blog.change_entry"
+        ):
+            return Entry.objects.all()
+        else:
+            return Entry.objects.published()
```
pauloxnet, line 60:
```suggestion
            return self.get_queryset()
```
I would reuse `get_queryset` from `BlogViewMixin`.

pauloxnet (commented):
I added 2 minor suggestions.

blog/views.py
marksweb: Ok, that's fair :+1:

adamzap:
@marksweb Could you check the CI failure here?

marksweb:
> @marksweb Could you check the CI failure here?

Done - there was an actual issue I'd not spotted 🤦🏼

adamzap:
@bmispelon Do you have any concerns here, or can we merge this?

bmispelon (approved):
Nice work Mark, thanks for this! 👍🏻

PR #1760, opened 2024-11-21
Added billboard for 2024 developer survey

bmispelon:
This was requested by @thibaudcolas .

I took inspiration from a4799dc44be834c6376c07739a222b1a55d6ffd1 for the wording (found using `git log -S survey`).

Here's how it looks (light mode):

Screenshot 2024-11-21 at 21-07-37 Django overview Django

### DIFF
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -676,6 +676,7 @@ blockquote {
 /* banner for billboard */
 #billboard {
     background-color: $green-medium;
+    border-top: 1px solid $black;
 
     * {
         color: $black;
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -700,6 +700,7 @@ blockquote {
 
     a, h1 a {
         color: $black;
+        font-weight: bold;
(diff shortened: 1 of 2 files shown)


djangoproject/scss/_style.scss
```diff
@@ -699,6 +699,7 @@ blockquote {
```
sabderemane, line 699: I can't make the suggestion above but could you add a border-top in #billboard style like this?

```suggestion
#billboard {
    ...
    border-top: 1px solid $black;
    ...
```

IMO, it better differentiates with the banner and title above.

How it's look like with it:
Capture d’écran 2024-11-25 à 19 06 21
Capture d’écran 2024-11-25 à 19 06 12

sabderemane (commented):
Small suggestion but otherwise seems to look good to me! Thank you Baptiste ⭐

djangoproject/scss/_style.scss
bmispelon: Good idea, I've added the change. Thanks!

bmispelon:
Thanks for the review @sabderemane ! 💚

PR #1761, opened 2024-11-21
Fix #1722 -- Add more tests for fundraising webhook

### LINKED ISSUE #1722: Add tests for fundraising web hooks
It seems that the code in `fundraising.views.WebhookHandler` doesn't have any tests associated with it.

It would be nice to have a basic suite of tests, at least for the "happy path". This could prevent bugs like the one fixed recently in e9952130db15e603477a5bc2a3a8f202c99cbd65.

ontowhee:
Fix for #1722 

Added test for `checkout_session_completed`. The test fails when reverting the change in [e995213](https://github.com/django/djangoproject.com/commit/e9952130db15e603477a5bc2a3a8f202c99cbd65).

New files are added in `test_data`. They were generated by calling calling `stripe.Customer.list().data[0]`, `stripe.PaymentIntent.list().data[0]`, etc.
- customer.json
- payment_intent.json
- session_completed.json

The `payment_succeeded` webhook still needs a test, but I wanted a quick review to see if I've approached this correctly.

### DIFF
--- a/fundraising/tests/test_views.py
+++ b/fundraising/tests/test_views.py
@@ -14,6 +14,7 @@ from django_hosts.resolvers import reverse as django_hosts_reverse
 from django_recaptcha.client import RecaptchaResponse
 
 from ..models import DjangoHero, Donation
+from ..views import WebhookHandler
 
 
 class TestIndex(TestCase):
@@ -253,3 +254,13 @@ class TestWebhooks(TestCase):
         response = self.post_event()
         self.assertEqual(response.status_code, 201)
         self.assertEqual(self.donation.payment_set.count(), 0)
+
+    @patch("stripe.Customer.retrieve")
+    @patch("stripe.PaymentIntent.retrieve")
+    def test_checkout_session_completed(self, payment_intent, customer):
+        customer.return_value = self.stripe_data("customer")
+        payment_intent.return_value = self.stripe_data("payment_intent")
+        event = self.stripe_data("session_completed")
+        WebhookHandler(event).handle()
+        customer.assert_called_once()
+        payment_intent.assert_called_once()
--- a/fundraising/test_data/customer.json
+++ b/fundraising/test_data/customer.json
@@ -1,0 +1,62 @@
+{
+    "account_balance": 0,
+    "address": null,
+    "balance": 0,
+    "cards": {
+      "data": [],
+      "has_more": false,
+      "object": "list",
+      "total_count": 0,
+      "url": "/v1/customers/cus_3MXPY5pvYMWTBf/cards"
+    },
+    "created": 1732222262,
+    "currency": null,
+    "default_card": null,
+    "default_currency": null,
+    "default_source": null,
+    "delinquent": false,
+    "description": null,
+    "discount": null,
+    "email": "customer@email.com",
+    "id": "cus_3MXPY5pvYMWTBf",
+    "invoice_prefix": "17A431B3",
+    "invoice_settings": {
+      "custom_fields": null,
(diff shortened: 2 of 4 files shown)


fundraising/tests/test_views.py
```diff
@@ -253,3 +254,13 @@ def test_zero_invoice_amount(self, event):
...
         self.assertEqual(response.status_code, 201)
         self.assertEqual(self.donation.payment_set.count(), 0)
+
+    @patch("stripe.Customer.retrieve")
+    @patch("stripe.PaymentIntent.retrieve")
+    def test_checkout_session_completed(self, payment_intent, customer):
+        customer.return_value = self.stripe_data("customer")
+        payment_intent.return_value = self.stripe_data("payment_intent")
+        event = self.stripe_data("session_completed")
+        WebhookHandler(event).handle()
```
alexgmin, line 264: What's the reason behind not using `self.post_event()` like the other tests and calling the class directly?

fundraising/tests/test_views.py
```diff
@@ -253,3 +254,13 @@ def test_zero_invoice_amount(self, event):
...
+        customer.assert_called_once()
+        payment_intent.assert_called_once()
```
alexgmin, line 266: Thinks that should also be tested:
- That the email is sent. There's an example above, `self.assertEqual(len(mail.outbox), 1)`
- That the response of the handler is the correct one, It should be a 204.
- What happens if the `DjangoHero` instance doesn't exist (this should probably be a different test (or could be a new PR to add this to every event test since right now they don't check this case))

fundraising/tests/test_views.py
ontowhee: I've updated this to call `self.post_event()`!

fundraising/tests/test_views.py
ontowhee: It now checks that the email is sent, and the status code is 204.

ontowhee: For the case where `DjangoHero` does not exist, I added a test that modifies the existing DjangoHero's stripe_customer_id, and then check that count for DjangoHero is increased to 2. Is that the right approach?

fundraising/tests/test_views.py
```diff
@@ -281,3 +284,23 @@ def test_unknown_event_type(self):
...
+        )
+        self.assertTrue(expected_url in mail.outbox[0].body)
+
+        self.assertEqual(DjangoHero.objects.count(), 1)
+
+    def test_checkout_session_completed_new_hero(self):
+        self.hero.stripe_customer_id = ""
+        self.hero.save()
+        response = self.post_event(self.stripe_data("session_completed"))
+
```
alexgmin, line 304: Can we check here too that the mail is sent?

fundraising/tests/test_views.py
```diff
@@ -281,3 +284,23 @@ def test_unknown_event_type(self):
...
+        self.assertEqual(len(mail.outbox), 1)
+        expected_url = django_hosts_reverse(
+            "fundraising:manage-donations", kwargs={"hero": self.hero.id}
+        )
+        self.assertTrue(expected_url in mail.outbox[0].body)
+
+        self.assertEqual(DjangoHero.objects.count(), 1)
+
+    def test_checkout_session_completed_new_hero(self):
+        self.hero.stripe_customer_id = ""
```
alexgmin, line 301: I'm wondering if there's a better way to do this. Either by deleting the Hero in this test before the event or by not hardcoding the stripe customer id in the `setUp`. 
This approach works, tho.
@bmispelon Any suggestions?

PR #1762, opened 2024-11-22
Fixed sponsor notes appearing on download page

bmispelon:
Thanks to user TheEpic-dev on the forum for reporting
the issue.

This was introduced in c024b32f65a48b6366baf2930e43c5f2014b1ace


bmispelon:
I verified manually with `git grep -w notes` that the field was not used in other templates: it has 42 results but none of them are relevant here.

PR #1763, opened 2024-11-23
Remove unneeded vendor prefixes for `box-shadow`

adamzap:
Browser versions that made these vendor prefixes obsolete were released by the middle of 2011: https://caniuse.com/css-boxshadow.

This is the first in what could be a long series of CSS/JS simplification PRs if this kind of thing is desired by the maintainers. I think we can reduce some complexity by taking small steps like this.

For instance, next I would want to factor the two affected mixins out. They are each used in one place.

### DIFF
--- a/djangoproject/scss/_utils.scss
+++ b/djangoproject/scss/_utils.scss
@@ -172,14 +172,10 @@ html {
 
 // Secondary content box-shadow:
 @mixin secondary-shadow-top {
-    -moz-box-shadow: 0 4px 8px rgba(12, 60, 38, 0.07);
-    -webkit-box-shadow: 0 4px 8px rgba(12, 60, 38, 0.07);
     box-shadow: 0 4px 8px rgba(12, 60, 38, 0.07);
 }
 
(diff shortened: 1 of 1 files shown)


bmispelon (approved):
This seems like a good change, thanks for submitting it 👍🏻 

Thanks also for linking to caniuse, it makes reviewing easier.

In terms of further PRs, I'm looking forward to them and I think we could increase the scope a little bit. I checked with `git grep -F -- '-moz-' '*.scss'` and that yields about 20 results, which I think should be reviewable in a single PR (assuming all of these vendor prefixes can be removed, which might be incorrect).

I also wonder if there's a tool we could use to automatically flag these kinds of outdated CSS (having a linter we could run in CI would be amazing). I don't mind reviewing PRs like these one by one, but I'd prefer we could do it all in an automated single pass if that's possible.

adamzap:
@bmispelon Ok thanks! I'll try to increase the scope while keeping different efforts separate, like removing vendor prefixes versus factoring Sass utilities out.

Issue #1764, opened 2024-11-23
Stripe event handling of the event checkout_session_completed should be in an atomic transaction block


alexgmin:
https://github.com/django/djangoproject.com/blob/5537f965ece29b766efe17c1d2137c3d2037811c/fundraising/views.py#L293
Multiple write queries are done here, if one fails it can end with inconsistent data.

arpan8925:
Working on this Issue

arpan8925:
Kindly Review #1768  Pull Request.

PR #1768, opened 2024-11-23
Fixed the atomic transaction issue

arpan8925:
1. Added **@transaction.atomic decorator** to the **checkout_session_completed** method

2. All database operations are now wrapped in a transaction:

        - Creating/getting the DjangoHero
        - Creating the Donation
        - Creating the Payment (for one-time donations)


arpan8925:
Solved issue of #1764

fundraising/views.py
```diff
@@ -290,10 +291,14 @@ def get_donation_interval(self, session):
...
             return "monthly"
 
+    @transaction.atomic
     def checkout_session_completed(self):
         """
         > Occurs when a Checkout Session has been successfully completed.
         https://stripe.com/docs/api/events/types#event_types-checkout.session.completed
+
+        All database operations are now wrapped in a transaction - if any operation fails,
+        all changes will be rolled back to maintain data consistency.
```
bmispelon, line 301: This addition is not needed

bmispelon (changes requested):
Hi and thanks for working on this! 🎸 

Decorating `checkout_sesion_completed()` like this shoudl work I think, but I'm not sure it's the best place to add the `atomic` block. If instead we had `with transaction.atomic()` before calling `handler()` inside `WebhookHandler.handle()`, then all web hooks would be fixed, not just the one.

There also need to be tests added, to make sure the behavior continues to work even as we refactor the code in the future.

Let me know what you think!

arpan8925:
@bmispelon Check if it fixes the issue, I made a separate file for the webhook test, review please

arpan8925:
It Seems , test working now !

The test is failing because the webhook handler isn't properly handling the InvalidRequestError from Stripe.

However, Everything fixed.. Please Check & Review

arpan8925:
image

arpan8925:
@bmispelon  Can you review ?

arpan8925:
@bmispelon

marksweb:
@arpan8925 I'm sure you'll get another review on here when someones free to do so. We're all volunteers here doing this in our free time.

bmispelon:
Hi,

There are several issues with your PR:
* This version is very different and much more complicated than the one you submitted initially, and it contains unexplained code changes
* Instead of adding a new test method to the `TestWebhooks` you created a whole new file with some of the old tests copy/pasted.
* There are now conflicts with other changes that have been merged since this pull request was opened (this is related to the first point)

Could you revert the changes and go back to the version you had at the beginning with only a `with atomic()` block? Then I will make some proposals for how to add tests for that.

Thanks! 🎸

PR #1769, opened 2024-11-24
Fixed #1758 -- Removed usage of internal Python API

### LINKED ISSUE #1758: Blog test using internal python package
Having recently switched to using `uv`, my version of python 3.12 doesn't come with the `test` module.

As stated at the top of the [docs](https://docs.python.org/3/library/test.html)

> Note
> The [test](https://docs.python.org/3/library/test.html#module-test) package is meant for internal use by Python only. It is documented for the benefit of the core developers of Python. Any use of this package outside of Python’s standard library is discouraged as code mentioned here can change or be removed without notice between releases of Python.

The [blog tests](https://github.com/django/djangoproject.com/blob/main/blog/tests.py#L2) import from `test`;
```python
from test.support import captured_stderr
```

So running `tox` on my machine results in;

```bash
ERROR: blog.tests (unittest.loader._FailedTest.blog.tests)
ImportError: Failed to import test module: blog.tests
  File "/Users/mark/.local/share/uv/python/cpython-3.12.7-macos-aarch64-none/lib/python3.12/unittest/loader.py", line 396, in _find_test_path
    module = self._get_module_from_name(name)
  File "/Users/mark/.local/share/uv/python/cpython-3.12.7-macos-aarch64-none/lib/python3.12/unittest/loader.py", line 339, in _get_module_from_name
    __import__(name)
  File "/Users/mark/dev/djangoproject.com/blog/tests.py", line 2, in <module>
    from test.support import captured_stderr
ModuleNotFoundError: No module named 'test'
```

In pytest there's a fixture for stderr, but not sure on the solution here.

bmispelon:


bmispelon:
I tried to set up a [flake8 plugin](https://github.com/K-dash/flake8-import-guard) to prevent this type of error in the future, but couldn't get it to work 🤷🏻

PR #1771, opened 2024-11-24
Remove unneeded CSS vendor prefixes

adamzap:
- `box-sizing`: https://caniuse.com/css3-boxsizing
- `::placeholder`: https://caniuse.com/css-placeholder
- `::selection`: https://caniuse.com/css-selection
- `linear-gradient`: https://caniuse.com/css-gradients
- `appearance`: https://caniuse.com/css-appearance
- `user-select`: (partial): https://caniuse.com/user-select-none

I didn't remove vendor prefixes from `normalize.scss` or `font-awesome` assets because those libraries should probably be updated as a whole.

### DIFF
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -1205,6 +1205,7 @@ a.cta {
     // Call To Action buttons
     @include sans-serif;
     font-weight: 700;
+    appearance: none;
     background: var(--primary);
     border: none;
     border-radius: 5px;
@@ -2877,6 +2878,7 @@ form {
     textarea {
         @include sans-serif;
         @include font-size(16);
+        appearance: none;
         background: var(--white-color);
         border: 1px solid var(--hairline-color);
         color: var(--body-fg);
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -61,10 +61,6 @@ button:focus {
     background-color: var(--selection);
 }
 
-::-moz-selection {
-    background-color: var(--selection);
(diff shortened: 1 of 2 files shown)


djangoproject/scss/_style.scss
```diff
@@ -1209,8 +1205,6 @@ a.cta {
     // Call To Action buttons
     @include sans-serif;
     font-weight: 700;
-    -webkit-appearance: none;
-    -moz-appearance: none;
```
bmispelon, line 1213: Shouldn't this get a non-prefixed replacement (`appearance: none`)?

bmispelon (commented):
This looks great, and again the caniuse links were super helpful for doing the review 🙇🏻 

I noticed two cases where the original css only had prefixed versions and no standard one (`appearance`). I think we should add the non-prefixed version there, what do you think?

djangoproject/scss/_style.scss
adamzap: You're right! Thanks, fixed!

djangoproject/scss/_style.scss
```diff
@@ -2887,8 +2877,6 @@ form {
     textarea {
         @include sans-serif;
         @include font-size(16);
-        -webkit-appearance: none;
```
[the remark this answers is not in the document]
adamzap: Fixed!

bmispelon (approved):
Looking great 👍🏻

bmispelon:
Merged and deployed, thanks again! 🚀

PR #1772, opened 2024-11-24
Remove unneeded `box-shadow` mixins

adamzap:
As of 3851b2f2, these are only used once and contain a single rule each.


bmispelon (approved):
Nice work 👍🏻

PR #1776, opened 2024-11-25
Remove unused `retina-display` Sass mixin

adamzap:


bmispelon (approved):
Nice catch, a `git grep -F retina-display` confirms that the mixin is indeed unused.

Are you using a tool to find these, or is this all manual?

Thanks! 🚀

adamzap:
Thanks! I'm doing it manually at this point as I'm learning more about the front-end aspects of the codebase.

PR #1777, opened 2024-11-25
Fixed #1629 -- Fixed search highlighting not working on stemmed words

### LINKED ISSUE #1629: Search highligt is not working online in some circumstances
Searching highlights is not working online for some words:
- https://docs.djangoproject.com/en/5.1/search/?q=Triaging

Screen Shot 2024-09-26 at 14 48 45

It works for other search terms:
- https://docs.djangoproject.com/en/5.1/search/?q=text

Screen Shot 2024-09-26 at 14 49 17

It's working in both case locally as tested by @bmispelon with PostgreSQL 16.

Remotely the website is using PostgreSQL 14.

bmispelon:


docs/tests.py
```diff
@@ -616,6 +617,31 @@ def test_search_update(self):
         self.assertEqual(Document.objects.search_update(), 6)
         self.assertEqual(Document.objects.exclude(search=None).count(), 6)
 
+    def test_search_highlight_stemmed(self):
+        # The issue only manifests itself when the defaut search config is not english
+        with connection.cursor() as cursor:
+            cursor.execute("SET default_text_search_config TO 'simple'", [])
```
bmispelon, line 623: Ideally this would be set globally on the connection, but I couldn't figure out how to do that 😿

pauloxnet (approved):
Thanks for the fix. 👍🏻

Issue #1698, opened 2024-10-25
Showcase Django users on the homepage


thibaudcolas:
Follow-up to #110. On the homepage, we have [commented-out](https://github.com/django/djangoproject.com/blob/main/djangoproject/templates/homepage.html#L68) code showcasing Django users. What do people think of resurrecting this, to make Django more compelling by showcasing "big names" that are universally recognized?

Here’s what the current implementation looks like as-is:

![Django users showcase screenshot, titled Who is Django For?](https://github.com/user-attachments/assets/2741efa3-00f5-426d-bd25-60614c5e23fe)

A lot of other projects do this. For example here’s what [Ruby on Rails](https://rubyonrails.org/) does:

![Rails users showcase screenshot, titled You’re in Good Company](https://github.com/user-attachments/assets/46666326-12b9-4230-af50-7650ba72c695)

marksweb:
I like this rails example. Looks less like a backend dev built it. 

I know we all love how easy bootstrap cards are, but a grid of icons works well for this purpose.

arpan8925:
May I work on this Issue and Make the UI, Similar like Ruby Rails Brand Section ?

ulgens:
@arpan8925 I don't think you need to ask for permission, you can just state you're going to handle it and submit a PR.

arpan8925:
@ulgens  Thanks for mentioning, I am checking

PR #1778, opened 2024-11-25
Remove assets for removed django-companies element

adamzap:
The SCSS and images removed here predate even 73d29b78, so it's hard to tell when they were added. Searching shows that the images, class, and mixin are no longer used. Plus, the Rdio service was discontinued at the end of 2015, so that's a hint.


bmispelon (approved):
Nice find! This looks like some leftovers from the "case study" feature that never launched (and predates the last big redesign 10 years ago I think). There's a bit more information if you dig a bit starting from #1698 .

Thanks! 🎸

PR #1779, opened 2024-11-26
Remove Modernizr integration

adamzap:
[Modernizr](https://modernizr.com/) is a feature-detection library that has been a part of this project for around ten years. Using its CSS classes, this project was detecting the following browser features:

- Border Radius: https://caniuse.com/border-radius
- Box Shadow: https://caniuse.com/css-boxshadow
- CSS Animations: https://caniuse.com/css-animation
- SVG: https://caniuse.com/svg

Each of these features has been available in modern browsers for many years, so we no longer need Modernizr.

This patch does the following:

- Removes `modernizr.js`
- Removes PNG logos that were fallbacks for SVG logos
- Refactors SCSS that was pivoting on Modernizr CSS classes (`mdzr-*`)


djangoproject/scss/_style.scss
```diff
@@ -1761,89 +1708,70 @@ h2+.list-link-soup {
...
-                &.icon-cogs {
-                    background-position: -334px -12px;
-                }
+            &.inview {
+                transform: rotate(0turn);
             }
 
-            // Border Radius, SVG, and Animation Enabled
-            .mdzr-svg.mdzr-borderradius.mdzr-cssanimations & {
-                -webkit-transition: all 0.3s ease-out;
```
adamzap, line 1798: I also removed unneeded `-webkit-` vendor prefixes from `transform` and `transition` where the SCSS was changing anyway. I hope this isn't too far out of scope!

- https://caniuse.com/transforms2d
- https://caniuse.com/mdn-css_properties_transition

djangoproject/scss/_style.scss
```diff
@@ -3131,23 +3052,6 @@ div[role=main] {
                 color: var(--white-color);
             }
         }
-
-        .mdzr-no-borderradius & {
```
adamzap, line 3135: This class was looking for the absence of `border-radius` support, so it can be removed altogether.

djangoproject/scss/_style.scss
```diff
@@ -343,23 +333,19 @@ blockquote {
...
 .full-width {
+    border: none;
+    box-shadow: 0 4px 8px rgba(12, 60, 38, 0.07);
 
     // Single column, full width layout. See /styleguide for explanation.
     &.container {
         width: 100%;
         padding: 0;
         max-width: none;
-        border-bottom: 1px solid #ddd;
```
adamzap, line 352: This is the only oddity in the SCSS refactor. Because of the order of the rules and nesting, the `border-bottom` rule was persisting over the `border: none` rule above. The intention of the SCSS seems to be to favor a box shadow over a border, so the `border-bottom` rule can be safely removed. It is always overridden in production:

image

djangoproject/scss/_style.scss
thibaudcolas: Indeed!

djangoproject/scss/_style.scss
thibaudcolas: Thank you!

djangoproject/scss/_style.scss
thibaudcolas: Thanks, spot on.

thibaudcolas (approved):
Thank you @adamzap! As far as I can tell from the code this is all spot on.

bmispelon:
Thank you both, great work 🙇🏻

PR #1780, opened 2024-11-26
Fixed trac revision author query.

sarahboyce:


tracdb/stats.py
```diff
@@ -34,44 +35,46 @@ def _inner(f):
...
 @stat("Commits")
-def commit_count(username):
-    count = Revision.objects.filter(author=username).count()
-    # This assumes that the username is their GitHub username, this is very
-    # often the case. If this is incorrect, the GitHub will show no commits.
-    link = f"https://github.com/django/django/commits/main/?author={username}"
+def commit_count(user):
+    count = Revision.objects.filter(
+        Q(author__istartswith=f"{user.username} <")
+        | Q(author__istartswith=f"{user.get_full_name()} <")
```
sarahboyce, line 49: This logic is not perfect but was what I could think of based off a few examples that cover most cases and hopefully won't double count :+1:

bmispelon:
Before reviewing this I want to test locally what Trac ends up storing in `Revision.author`, and I would also like to see if using `trac_user_rlookup = enabled` [1] in the config could improve things.

The lookup by username/full name seems a bit brittle, but I don't think I have any better idea unfortunately (Trac's lack of actual user table is again making things difficult).

[1] https://github.com/edgewall/trac/blob/trunk/tracopt/versioncontrol/git/git_fs.py#L321-L328

tracdb/stats.py
codingjoe: Hm... is there's no index on the column. You might want to add a case-insensitive index to avoid slow table scans on each query.

tracdb/stats.py
```diff
@@ -34,44 +35,46 @@ def _inner(f):
...
+    ).count()
+    # This assumes that the username is their GitHub username.
+    link = f"https://github.com/django/django/commits/main/?author={user.username}"
```
codingjoe, line 52: Hm… is the username URL-safe? You might want to add some escaping here.

tracdb/stats.py
```diff
@@ -34,44 +35,46 @@ def _inner(f):
...
     return StatData(count=count, link=link)
 
 
 @stat("Tickets fixed")
-def tickets_fixed(username):
-    query = f"owner={username}&resolution=fixed"
+def tickets_fixed(user):
+    query = f"owner={user.username}&resolution=fixed"
```
codingjoe, line 58: Same, is the username guaranteed to be safe?

tracdb/stats.py
codingjoe: As someone who's chosen their wife's family name, maybe checking the mail is a better identifier than the name? That's how GitHub attributes commits if I am not mistaken.

PR #1784, opened 2024-11-30
Remove `unveil` JavaScript library

adamzap:
The [`unveil`](https://github.com/luis-almeida/unveil) library is used to lazy-load images, and its last release was in 2013. This is useful functionality, but we don't need JavaScript to get it. All modern browsers now support it natively with the `loading="lazy"` attribute on the `img` HTML tag:

- https://caniuse.com/loading-lazy-attr
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#loading

This patch removes the `unveil` library and all references to it and uses the `loading` attribute in the one place where `unveil` was being used. This includes `fundraising-heroes.js`, the sole purpose of which was to initialize `unveil`.

It seems that `unveil` was also trying to swap out higher-quality images on retina display. My sense is that this is not needed, and the associated attributes have been removed.


djangoproject/templates/fundraising/includes/_hero_with_logo.html
```diff
@@ -2,10 +2,8 @@
 <div class="hero">
   <div class="hero-logo">
     {% if obj.url %}<a href="{{ obj.url }}" rel="nofollow">{% endif %}
-    <img src="{% static 'img/fundraising-heart.svg' %}"
-         data-src="{% if obj.logo %}{{ obj.thumbnail.url }}{% else %}{% static 'img/fundraising-heart.svg' %}{% endif %}"
-         data-src-retina="{% if obj.logo %}{{ obj.thumbnail.url|resolution:'2x' }}{% else %}{% static 'img/fundraising-heart.svg' %}{% endif %}"
-         alt="{{ obj.display_name }}" />
+    <img src="{% if obj.logo %}{{ obj.thumbnail.url }}{% else %}{% static 'img/fundraising-heart.svg' %}{% endif %}"
+         loading="lazy" alt="{{ obj.display_name }}" />
```
bmispelon, line 6: I realize that the code was already there before you made your change, but since we're changing it I figured we might as well improve the alt text a little bit here, maybe something like:
```suggestion
    {% if obj.logo %}
        <img
            src="{{ obj.thumbnail.url }}"
            loading="lazy"
            alt="{% blocktranslate trimmed with company_name=obj.display_name %}
                Logo of company {{ company_name }}{% endblocktranslate %}"
        >
    {% else %}
        <img src="{% static 'img/fundraising-heart.svg' %}" alt="{% translate 'Pixelated heart logo' %}">
    {% endif %}
```

bmispelon (commented):
Fantastic work again 👏🏻 

I've left a non-blocking accessibility improvement suggestion, and I'd also be curious if you could elaborate on my higher quality images on retina displays might not be needed (is it because the images are not big enough to make a difference?)

djangoproject/templates/fundraising/includes/_hero_with_logo.html
adamzap: Done!

adamzap:
@bmispelon Thanks for the kind words!

I was thinking that it isn't worth the complexity to swap retina images out. The non-retina versions total ~550K, and the retina versions total 1.2M. My typical approach is to just use the retina version, so I may not have a good argument here.

I realized that the `unveil` library is causing the heart fallback to load unconditionally, then it loads the appropriately-sized images. Do we need to maintain that functionality? I'm assuming we don't, but it's up to you.

If you want to keep having two versions of the images, I realized that we could use the `srcset` attribute: https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/srcset#html

Let me know how you want to proceed. I'm happy to implement any of the above, though maintaining the eager-loading of the heart fallback before the actual image is loaded would increase complexity and require JavaScript.

djangoproject/templates/fundraising/includes/_hero_with_logo.html
```diff
@@ -2,10 +2,16 @@
 <div class="hero">
   <div class="hero-logo">
     {% if obj.url %}<a href="{{ obj.url }}" rel="nofollow">{% endif %}
-    <img src="{% static 'img/fundraising-heart.svg' %}"
-         data-src="{% if obj.logo %}{{ obj.thumbnail.url }}{% else %}{% static 'img/fundraising-heart.svg' %}{% endif %}"
-         data-src-retina="{% if obj.logo %}{{ obj.thumbnail.url|resolution:'2x' }}{% else %}{% static 'img/fundraising-heart.svg' %}{% endif %}"
-         alt="{{ obj.display_name }}" />
+    {% if obj.logo %}
+      <img
+        src="{{ obj.thumbnail.url }}"
```
bmispelon, line 7: If I understood the `srcset` page on MDN, it should be as easy as:
```suggestion
        src="{{ obj.thumbnail.url }}"
        srcset="{{ obj.thumbnail.url|resolution:'2x' }} 2x"
```

djangoproject/templates/fundraising/includes/_hero_with_logo.html
```diff
@@ -2,10 +2,16 @@
...
+        loading="lazy"
+        alt="{% blocktranslate trimmed with company_name=obj.display_name %}
```
bmispelon, line 9: `blocktranslate` and `translate` need to load the `i18n` template library on line 1 (I should have mentionned it in my previous comment, sorry).

Come to think of it, it's suspicious that no test caught this, that's probably a sign that we're missing a testcase that actually renders this template 🤔

bmispelon (changes requested):
> I realized that the unveil library is causing the heart fallback to load unconditionally, then it loads the appropriately-sized images. Do we need to maintain that functionality? I'm assuming we don't, but it's up to you.

I agree with your assumption, this does not seem that useful. It's quite jarring to see the hearts switching out to the real image once it loads (a less prominent spinner image would work better but I'm not sure how hard it'd be to implement). One thing I wonder (and have found hard to test locally) is whether the lack of a default image make the element change size once the image loads. The little I've read about this says we should be setting `width`/`height` as HTML attributes on the `img` and we're not doing that, which leads me to believe that we're at risk of the page jumping around while elements are loaded.

> If you want to keep having two versions of the images, I realized that we could use the srcset attribute [...]

This seems like a good usecase for `srcset`, especialy that it seems easy to implement on our end (and it seems [well supported](https://caniuse.com/srcset)). I've added a suggestion for it.

djangoproject/templates/fundraising/includes/_hero_with_logo.html
adamzap: Done in [8aa3112](https://github.com/django/djangoproject.com/pull/1784/commits/8aa3112a9aa11fe3673a816c356c0359018631c6)!

djangoproject/templates/fundraising/includes/_hero_with_logo.html
adamzap: I fixed the import in [4bcb6b4](https://github.com/django/djangoproject.com/pull/1784/commits/4bcb6b476e6259315182d637340bdfe365a400ed). Do you want me to add a view test on this branch, or is that out of scope?

bmispelon: I mean if you're offering to write tests I'm not going to say no 😁 

What about starting with one that exercises the `{% if obj.logo %}` that was added in this PR: one case with a logo, one without and testing for the presence of the right image?

adamzap: I pushed a commit (then had it fixed by `pre-commit` 😎). Is that what you had in mind? I'm not familiar with the testing patterns in the project yet.

I wanted to add tests at the `display_django_heroes` template tag level, but I didn't see a straightforward way to make assertions about the rendered HTML rather than the context returned by the template tag.

It's also unfortunate that I can't assert the presence of `fundraising-heart.svg` since it's always going to be on the page higher-up. I also can't assert the URL of the logo since it seems like `ManifestStaticFilesStorage` is enabled when the test suite runs.

Anyway, I may need some more detailed guidance if you want any other changes here. Thanks!

bmispelon: I see what you mean about the tests being annoying to write because of the `STORAGES` configuration we inherit from `settings.prod`.

I've pushed a commit that adds a workaround for `settings.STORAGES` using a combination of a temporary directory (using `tempfile.mkdtmp`) and settings override (using `TestCase.settings()`). I also changed the assertions to use `assertContains(..., html=True)` which I think makes the test a bit nicer (and maybe more roboust).

Let me know what you think!

adamzap: Yes, this looks good to me. Thanks! It might be nice to make the storage back-end override more of a generic utility later if more tests need it.

fundraising/tests/test_views.py
```diff
@@ -23,8 +29,71 @@ def test_redirect(self):
...
+            )
+            response = self.client.get(self.index_url)
+
+        self.assertContains(
+            response,
+            """<img
+                src="/m/cache/9b/e7/9be7b86ebc112b001cad84f900bf0bf7.png"
+                srcset="/m/cache/9b/e7/9be7b86ebc112b001cad84f900bf0bf7@2x.png 2x"
+                width="170"
+                height="170"
```
adamzap, line 91: @bmispelon There may be a bit of value in making the width and height of the test image different to ensure that the correct properties are being used in the template. Up to you!

adamzap:
@bmispelon I think everything has been addressed on my end. Let me know if there's anything else I can do to move this forward.

bmispelon:
I've dug a bit deeper into the existing tests and noticed something weird. It turns out we did have tests that were exercising the `djangoproject/templates/fundraising/includes/_hero_with_logo.html` template.
I've created a draft PR that intentionally breaks the template: #1793 and I can see an error in the test logs, but for some reason this error is not making the test fail:
```
   test_anonymous_donor_with_logo (fundraising.tests.test_views.TestCampaign.test_anonymous_donor_with_logo) ... [Errno 2] No such file or directory: '/home/runner/work/djangoproject.com/djangoproject.com/djangoproject/static/img/logo-django.png'
    File "/home/runner/work/djangoproject.com/djangoproject.com/.tox/py312-tests/lib/python3.12/site-packages/sorl/thumbnail/base.py", line 104, in get_thumbnail
      source_image = default.engine.get_image(source)
    File "/home/runner/work/djangoproject.com/djangoproject.com/.tox/py312-tests/lib/python3.12/site-packages/sorl/thumbnail/engines/pil_engine.py", line 83, in get_image
      buffer = BytesIO(source.read())
                       ^^^^^^^^^^^^^
    File "/home/runner/work/djangoproject.com/djangoproject.com/.tox/py312-tests/lib/python3.12/site-packages/sorl/thumbnail/images.py", line 164, in read
      f = self.storage.open(self.name)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/home/runner/work/djangoproject.com/djangoproject.com/.tox/py312-tests/lib/python3.12/site-packages/django/core/files/storage/base.py", line 22, in open
      return self._open(name, mode)
             ^^^^^^^^^^^^^^^^^^^^^^
    File "/home/runner/work/djangoproject.com/djangoproject.com/.tox/py312-tests/lib/python3.12/site-packages/django/core/files/storage/filesystem.py", line 68, in _open
      return File(open(self.path(name), mode))
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/work/djangoproject.com/djangoproject.com/djangoproject/static/img/logo-django.png'
  Remote file [static/img/logo-django.png] at [170x170] does not exist
  ok
```

Locally however, the test is failing for me.
So I'd like to figure out what's going on here before moving forward. The `unveil` part of this PR is ready to go though. If I can't figure out this weirdness within a few days, maybe we'll just merge that and leave the tests for later. I'll keep you posted 🏣

adamzap:
I was getting that error locally too, but I figured it was expected. It's being generated by [this line on the `main` branch](https://github.com/django/djangoproject.com/blob/4f9acf7c8735db92d279fa7b3763737712597d06/fundraising/tests/test_views.py#L44) ([this line on this PR's branch](https://github.com/adamzap/djangoproject.com/blob/1c1b8c082854b9acc43894309ee5e3f9efc5aadd/fundraising/tests/test_views.py#L113)). The [`sorl-thumbnail`](https://github.com/jazzband/sorl-thumbnail/) package is logging then swallowing the exception [here](https://github.com/jazzband/sorl-thumbnail/blob/7d4ca2b00f541ade011a51ae777eb6eefd1010d0/sorl/thumbnail/base.py#L106). Then it logs the other more thumbnail-specific message [here](https://github.com/jazzband/sorl-thumbnail/blob/7d4ca2b00f541ade011a51ae777eb6eefd1010d0/sorl/thumbnail/base.py#L113-L116).

One way to fix it would be to use `self.imagefile` on the line I linked above instead of using the string image path for the `logo` parameter to `DjangoHero`. I tried it locally, and it works. Do you like that approach?

bmispelon:
I agree with the fix you're proposing (I'd also tried it locally), but I still don't understand why things are behaving differently between my local version and CI and I'd like to get to the bottom of that.

I noticed that the tracebacks are quite different, here's what I'm getting locally (seems the exception is happening inside a signal, I wonder if that's a clue 🤔 ):
```
  File "/path/to/venv/lib/python3.12/site-packages/django/test/utils.py", line 443, in inner
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "./djangoproject.com/fundraising/tests/test_views.py", line 40, in test_anonymous_donor_with_logo
    hero = DjangoHero.objects.create(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/venv/lib/python3.12/site-packages/django/db/models/manager.py", line 87, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/venv/lib/python3.12/site-packages/django/db/models/query.py", line 679, in create
    obj.save(force_insert=True, using=self.db)
  File "./djangoproject.com/fundraising/models.py", line 56, in save
    return super().save(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/venv/lib/python3.12/site-packages/django/db/models/base.py", line 822, in save
    self.save_base(
  File "/path/to/venv/lib/python3.12/site-packages/django/db/models/base.py", line 924, in save_base
    post_save.send(
  File "/path/to/venv/lib/python3.12/site-packages/django/dispatch/dispatcher.py", line 189, in send
    response = receiver(signal=self, sender=sender, **named)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "./djangoproject.com/fundraising/models.py", line 176, in create_thumbnail_on_save
    return kwargs["instance"].thumbnail
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "./djangoproject.com/fundraising/models.py", line 100, in thumbnail
    return get_thumbnail(self.logo, "170x170", quality=100) if self.logo else None
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/venv/lib/python3.12/site-packages/sorl/thumbnail/shortcuts.py", line 8, in get_thumbnail
    return default.backend.get_thumbnail(file_, geometry_string, **options)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/venv/lib/python3.12/site-packages/sorl/thumbnail/base.py", line 136, in get_thumbnail
    default.kvstore.get_or_set(source)
  File "/path/to/venv/lib/python3.12/site-packages/sorl/thumbnail/kvstores/base.py", line 63, in get_or_set
    self.set(image_file)
  File "/path/to/venv/lib/python3.12/site-packages/sorl/thumbnail/kvstores/base.py", line 44, in set
    image_file.set_size()  # make sure its got a size
    ^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/venv/lib/python3.12/site-packages/sorl/thumbnail/images.py", line 141, in set_size
    image = default.engine.get_image(self)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/venv/lib/python3.12/site-packages/sorl/thumbnail/engines/pil_engine.py", line 83, in get_image
    buffer = BytesIO(source.read())
                     ^^^^^^^^^^^^^
  File "/path/to/venv/lib/python3.12/site-packages/sorl/thumbnail/images.py", line 164, in read
    f = self.storage.open(self.name)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/venv/lib/python3.12/site-packages/django/core/files/storage/base.py", line 22, in open
    return self._open(name, mode)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/venv/lib/python3.12/site-packages/django/core/files/storage/filesystem.py", line 68, in _open
    return File(open(self.path(name), mode))
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: './djangoproject.com/djangoproject/static/img/logo-django.png'
```

bmispelon:
I've been looking at this for several hours now and slowly losing my mind. But maybe also making progress, question mark?

First off, I was incorrect in thinking that the existing `test_anonymous_donor_with_logo()` test would make use of the `_hero_with_logo.html` template: because the donor is anonymous that template is not being used.

Second, the `test_anonymous_donor_with_logo()` was "broken" by 132af0b38c6c8dbd8e9da720bc1618eaf710f24b because it reference the file `logo-django.png` which was removed in that commit. Using a filename that still exists (like `fundraising-heart.png`) makes the test pass with no warning.

Third - and this is the part that utterly puzzles me - I can turn the test failure into a warning (like the one that appears in the CI logs) simply by changing the logo to another filename that also doesn't exist. That is to say, changing [line 44](https://github.com/django/djangoproject.com/blob/ede018f69a7ae5ecacb99978d067f53613d05353/fundraising/tests/test_views.py#L44) from `logo="static/img/logo-django.png"` to `logo="static/img/logo-django0.png"` makes the test pass with a warning about a `FileNotFound`. I cannot explain this.

adamzap:
Wow, good sleuthing there! Do you want to do that work in this PR?

bmispelon:
> Wow, good sleuthing there! Do you want to do that work in this PR?

I think the unveil removal should probably be done on its own, and I can do the test improvements in a different ticket/pr. But it's getting late here, so that's a problem for tomorrow 😴

adamzap:
Ok, sounds good. Whatever you think is best!

bmispelon:
I've pushed two commits to the branch:

* eac573c0356f28fa96423221a8b90b81ce6f188e which moves all the `thumbnail` properties into a mixin, and adds a `THUMBNAIL_SIZE` attribute on the model class.
* aff1dd658e8117984b69b4e8b64f2665eb1ae2df which moves the "temporary MEDIA_ROOT" logic into a mixin and simplifies the code a bit. I also change the template to use the new `THUMBNAIL_SIZE` instead of retrieving it from the thumbnail's `width` and `height`. I couldn't find how to write a test for that, but I believe that accessing the `width` and `height` properties of the thumbnail means that the file needs to be read from the filesystem, which could be quite expensive especially on a page with 100s of thumbnails.

If those two seem ok to you, my plan is to keep the `LogoThumbnailMixin` commit on its own, then squash everything else into a single "remove unveil.js" commit. Does that sound ok?

fundraising/tests/test_views.py
```diff
@@ -1,30 +1,96 @@
...
+class TemporaryMediaRootMixin:
+    """
+    A TestCase mixin that overrides settings.MEDIA_ROOT for every test on the
+    class to point to a temporary directory that is destroyed when the tests
+    finished.
+    The content of the directory persists between different tests on the class.
+    """
+
+    @classmethod
+    def setUpClass(cls):
```
adamzap, line 33: Do you want to call `super().setUpClass()` in this method? Same question for `tearDownClass` below.

adamzap:
I defer to you on the implementation details, but this is definitely cleaner. Nice work!

Another approach would be to implement the `TemporaryMediaRootMixin` behavior in a custom test runner class so that the whole project's tests would benefit from it automatically. I can respect the idea of opting into the behavior though. Whatever you want to do is fine with me!

fundraising/tests/test_views.py
bmispelon: Excellent catch, I've added them both 👍🏻

bmispelon:
> Another approach would be to implement the TemporaryMediaRootMixin behavior in a custom test runner class so that the whole project's tests would benefit from it automatically. I can respect the idea of opting into the behavior though. Whatever you want to do is fine with me!

Yes, that's a good approach and has the benefit of applying to all tests, even future ones. It is a bit heavier duty though and I worry about the performance impact of creating and destroying all these directories (I haven't benchmarked though, so maybe my worries are not warranted), especially that the vast majority of tests don't handle files at all.

I would say let's keep this a mixin for now, and if we notice in the future that it keeps being a problem then we can devise a more robust solution.

adamzap:
Sounds good!

Issue #1782, opened 2024-11-28
No way to delete your Django account


kevinrenskers:
I have an old-style non-Github account that I used on https://code.djangoproject.com/ in the past, and I found no way to delete this account myself (which seems in violation of GDPR rules?). It should probably be possible to delete your own account, or at least the procedure to ask for this should be made clear.

See also https://discord.com/channels/856567261900832808/1311792124286730512.

p.s. I was kind of surprised that the URLs https://www.djangoproject.com/accounts/, https://www.djangoproject.com/accounts/account/ and https://www.djangoproject.com/account/ all didn't work, I would've expected that I could manage my Django account there (since there are https://www.djangoproject.com/accounts/register/ and https://www.djangoproject.com/accounts/password_reset/ after all).

(p.p.s. Why doesn't the forum work with Django accounts? It's a strange split system with Django accounts that can't be managed anywhere on the Django site or the Issues site, and a forum that doesn't work with these accounts.)

bmispelon:
@kevinrenskers I've been working on this in the past few days and put together a "delete account" view on the site: https://github.com/django/djangoproject.com/pull/1786. I'd appreciate your feedback on it.

Regarding your questions:

* The "manage account" page does exist though I do agree that it's basically impossible to discover. It's at https://www.djangoproject.com/accounts/edit/
* The split system for login is a historical artefact and has been a big pain in my backside for a while. Long story short: the forum is hosted completely externally and has no access to the user database that djangoproject.com has. The issues site is hosted locally, but it's not a Django project and as such is not integrated to the user databases beyong being able to use it for logging in.

kevinrenskers:
It seems like good improvements in that PR, I'm just not sure how a user would find this delete account option? This is on https://www.djangoproject.com/accounts/edit/ I assume? Which I never knew about.

PR #1786, opened 2024-12-01
Add view to delete current logged in user

bmispelon:
Refs #1782 

I also added two other small commits on the same branch:
- 60eb512122b5713ca5256d820dcf769e77cf7de1 changes the `on_delete` value of `BoardMember.account` to `PROTECT` to avoid users being able to delete their active board memberships
- f4234de2de5530b78ad43c3b3cfbb76c1207fe16 doesn't send a default `next` value of `/` when logging in. Doing so bypasses the standard `settings.LOGIN_REDIRECT_URL` and makes it harder for users to discover their profile edit page (which is already hard to discover)

Here are some screenshots of the new pages that were added (light mode only):
**confirmation page:**
Screenshot 2024-12-01 at 17-31-59 Confirmation delete your profile Django

**error page:**
Screenshot 2024-12-01 at 17-52-38 Confirmation delete your profile Django

**success page:**
Screenshot 2024-12-01 at 18-00-23 The web framework for perfectionists with deadlines Django


accounts/forms.py
```diff
@@ -36,3 +38,52 @@ def save(self, commit=True):
             if commit:
                 instance.user.save()
         return instance
+
+
+class DeleteProfileForm(forms.Form):
+    """
+    A form for delete the request's user and their associated data.
```
adamzap: A form for **deleting**...

accounts/forms.py
```diff
@@ -36,3 +38,52 @@ def save(self, commit=True):
...
+
+    This form has no fields, it's used as a container for validation and deltion
```
adamzap: deltion => deletion

djangoproject/templates/accounts/delete_profile_success.html
```diff
@@ -0,0 +1,17 @@
+{% extends "base.html" %}
+
+{% load i18n %}
+
+{% block content %}
+    <h2>{% translate "Account deleted" %}</h2>
+    <p>
+        {% translate "Your account and its data was successfully deleted and you've been logged out." %}
```
adamzap: Two small issues here:

1. was => were
2. deleted, and

djangoproject/templates/accounts/delete_profile_success.html
```diff
@@ -0,0 +1,17 @@
...
+    </p>
+    <p>
+        {% url "community-index" as community_index_url %}
```
adamzap, line 10: You're setting this here because you can't set it inside of `blocktranslate`, correct?

djangoproject/templates/accounts/delete_profile_success.html
```diff
@@ -0,0 +1,17 @@
...
+        {% blocktranslate trimmed %}
+            Thanks for spending your time with us, we hope we'll still see you
+            around on our <a href="{{ community_index_url }}">various community spaces, online and off.
```
adamzap: Could you close this `a` tag?

djangoproject/templates/accounts/delete_profile_success.html
```diff
@@ -0,0 +1,17 @@
+{% extends "base.html" %}
+
```
adamzap: The other template doesn't have a blank line between the `extends` and `load` tags. Could we be consistent?

djangoproject/templates/accounts/delete_profile.html
```diff
@@ -0,0 +1,40 @@
...
+{% block content %}
+    {% if form.errors %}
+        <h2>{% translate "Could not delete account" %}</h2>
+
+        <p>{% blocktranslate trimmed %}
+            Sorry, something went wrong when trying to delete your account.
+            That means there's probably some protected data still associated
+            with your account.
+            Please contact
+            <a href="mailto:ops@djangoproject.com?{{ OPS_EMAIL_PRESETS }}">the operations team</a>
```
adamzap: Could you add a comma after the closing `a` tag?

djangoproject/templates/accounts/delete_profile.html
```diff
@@ -0,0 +1,40 @@
...
+            and we'll sort it out for you.
+        {% endblocktranslate %}</p>
+    {% else %}
+        <h2>{% translate "Are you sure?" %}</h2>
+
+        <p>{% blocktranslate trimmed with username=request.user.username %}
+            ⚠️ You are about to delete all data associated with the username
+            <strong>{{ username}}</strong>.
```
adamzap: `{{ username}}` => `{{ username }}`

accounts/views.py
```diff
@@ -34,6 +36,42 @@ def edit_profile(request):
     return render(request, "accounts/edit_profile.html", {"form": form})
 
 
+@login_required
+def delete_profile(request):
+    if request.method == "POST":
+        form = DeleteProfileForm(data=request.POST, user=request.user)
+        if form.is_valid():
```
adamzap: Could this be `if form.is_valid() and form.delete():`?

accounts/views.py
```diff
@@ -34,6 +36,42 @@ def edit_profile(request):
...
+            if form.delete():
+                logout(request)
```
adamzap: Could you help me understand why we need to call `logout` if the user has already been deleted? What's the danger if we don't call it? Does it have something to do with session data?

accounts/views.py
```diff
@@ -34,6 +36,42 @@ def edit_profile(request):
...
+                return redirect("delete_profile_success")
+    else:
+        form = DeleteProfileForm(user=request.user)
+
+    context = {
+        "form": form,
+        # Strings are left translated on purpose (ops prefer english :D)
```
adamzap: Did you mean "untranslated" in this comment? Or maybe you meant translated into English. 😉

accounts/forms.py
```diff
@@ -36,3 +38,52 @@ def save(self, commit=True):
...
+    logic.
+    """
+
+    class InvalidFormError(Exception):
+        pass
+
+    def __init__(self, *args, user=None, **kwargs):
+        if user.is_anonymous:
```
adamzap, line 55: Is this check needed since the form is only used in a view with `login_required` applied?

accounts/tests.py
```diff
@@ -169,3 +171,50 @@ def test_profile_view_reversal(self):
         """
         for username in ["asdf", "@asdf", "asd-f", "as.df", "as+df"]:
             reverse("user_profile", host="www", args=[username])
+
+
+class UserDeletionTestCase(TestCase):
```
adamzap: The other test classes in this file end in `Tests`, not `TestCase`. Should we be consistent here?

accounts/tests.py
```diff
@@ -169,3 +171,50 @@ def test_profile_view_reversal(self):
...
+        return DeleteProfileForm(**formkwargs)
+
+    def test_deletion(self):
+        form = self.create_user_and_form()
+        self.assertFormError(form, None, [])
+        form.delete()
+        self.assertQuerySetEqual(User.objects.all(), [])
+
+    def test_anonymous_user_error(self):
+        self.assertRaises(TypeError, DeleteProfileForm, user=AnonymousUser)
```
adamzap, line 215: I've only seen [assertRaises](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises) used as a context manager. Do you prefer to using it this way? I defer to you on this but wanted to check.

accounts/forms.py
bmispelon: Fixed, thanks 👍🏻

djangoproject/templates/accounts/delete_profile_success.html
bmispelon: Yes, you get an error otherwise:
> 'blocktranslate' doesn't allow other block tags (seen 'url "community-index" as community_index_url') inside it

accounts/views.py
bmispelon: Yes, good point 👍🏻

accounts/views.py
bmispelon: That's a great question. To be honest if there was a reason why I added that, I've long forgotten it 😅 

I've tried and other than the `test_view_deletion_also_logs_out` breaking (expectedly), everything seems to keep working.

I guess the session middleware is smart enough to detect the user object is gone, and to invalidate the session. I'm not sure that's guaranteed though, maybe it depends on the kind of session backend?

I appreciate that's not a great argument for keeping the code around, so if you'd prefer to remove that I won't stand in your way 😅

accounts/views.py
bmispelon: I meant untranslated, thanks for catching that!

accounts/forms.py
bmispelon: Good catch, that's indeed redundant in the current usage of the form. I quite like the safety it brings in case that form ever get used somewhere else (in which case we'd only get an error if the form is saved, having the error at init time seems more debug-friendly).

accounts/tests.py
bmispelon: I use both where it makes sense. Here it saves a line and one indentation level, that's the only advantage 😅

bmispelon:
Wow, thanks for the thorough review @adamzap ! 🙏🏻 

I've rebased the branch on the latest `main` (there was a conflict) and pushed an extra `!fixup` commit for easier review (which should get squashed automatically on rebase, though I'm not sure GitHub will do that).
I've left open some threads where I thought it relevant, let me know what you think!

Cheers!

accounts/tests.py
adamzap: Ok great!

accounts/forms.py
adamzap: Ok, fair enough!

accounts/views.py
adamzap: I looked into it, and it seems to me that the only benefit is removing the unreachable session data from the `django_session` table. I'm not opposed to that kind of cleanup if you want to keep it.

djangoproject/templates/accounts/delete_profile_success.html
adamzap: Ok thanks!

marksweb (approved):
I think this looks good 👍🏼

accounts/forms.py
SaptakS: I agree with having the check at the form level as well, even though redundant.

PR #1789, opened 2024-12-03
Syntax fix 'URLs and views' collapsing-content in start.html template

RobbieTClark:
## start.html template syntax fix

Before:
Screenshot 2024-12-03 at 20 44 41
After:
Screenshot 2024-12-03 at 20 46 59

### DIFF
--- a/djangoproject/templates/conduct/faq.html
+++ b/djangoproject/templates/conduct/faq.html
@@ -71,7 +71,7 @@
       <a class="plink" href="{{ coc_url }}">require a code of conduct</a>.
       Isn't this redundant?{% endblocktranslate %}</p>
 
-  <p>{% translate "No: there's a difference between the two, and they're complementary.</p>
+  <p>{% translate "No: there's a difference between the two, and they're complementary." %}</p>
 
   <p>
     {% blocktranslate trimmed %}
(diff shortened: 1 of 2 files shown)


bmispelon:
Nice catch, thanks! 🎸 

I used `git grep -E '\{% translate "[^"]+$' -- '*.html'` to check if there was nother examples of the same typo in the repository and found another issue in https://github.com/django/djangoproject.com/blob/main/djangoproject/templates/conduct/faq.html#L74.
Could you fix that one too as part of you PR?

marksweb:
@RobbieTClark Thanks for catching this - no doubt another bad merge conflict resolution from me here. Too much time looking at `{% translate` tags doing all these.

RobbieTClark:
@bmispelon added a commit to correct conduct/faq.html 

Before:
Screenshot 2024-12-04 at 06 38 31
After:
Screenshot 2024-12-04 at 06 40 37

@marksweb I can imagine they all start becoming a blur after sifting through and adding them over and over!

bmispelon (approved):
Nice work, thanks 🙇🏻 

(I especially appreciate you taking the time to make screenshots, it makes the reviewing easier 👍🏻 )

PR #1791, opened 2024-12-04
Use 'docker compose' instead of 'docker-compose'

ulgens:
Compose v2 was released on June 2021 and hit GA on April 2022, changing how should we call compose.

> Unlike Compose V1, Compose V2 integrates into the Docker CLI platform and the recommended command-line syntax is `docker compose`.

https://docs.docker.com/compose/releases/migrate/#what-are-the-differences-between-compose-v1-and-compose-v2

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -370,18 +370,18 @@ Running Locally with Docker
 
 1. Build the images::
 
-    docker-compose build
+    docker compose build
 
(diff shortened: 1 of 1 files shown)


bmispelon (approved):
Nice work, thank you! 🎸

PR #1792, opened 2024-12-04
1633: Update Django dependency to version 5.1.5

### LINKED ISSUE #1633: Update Django to 5.1
At the moment we are using Django 4.2 but Django 5.1 to be able to use all the latest features and bugfixes.

In particular can be useful to start using `GeneratedField` to improve how page are shown in the documentation form the `metadata` JSONfield.

We have to wait for the resolution of #1574

ulgens:
https://docs.djangoproject.com/en/5.1/releases/5.1/
https://docs.djangoproject.com/en/5.1/releases/5.1/#backwards-incompatible-5-1

Resolves #1633


accounts/migrations/0002_migrate_sha1_passwords.py
```diff
@@ -1,17 +1,5 @@
...
-
-def forwards_func(apps, schema_editor):
-    User = apps.get_model("auth", "User")
-    users = User.objects.filter(password__startswith="sha1$")
-    hasher = PBKDF2WrappedSHA1PasswordHasher()
-    for user in users:
-        algorithm, salt, sha1_hash = user.password.split("$", 2)
-        user.password = hasher.encode_sha1_hash(sha1_hash, salt)
-        user.save(update_fields=["password"])
-
```
ulgens, line 14: @bmispelon Is this okay? I didn't want to squash migrations when there is only 2 of them.

ulgens: I may not have selected the all correct lines here, please check all changes in the related file.

ulgens:
@bmispelon The failure in the test looks like a flake but I don't have permission to re-run the pipeline.

accounts/migrations/0002_migrate_sha1_passwords.py
bmispelon: Yes, I think that should be OK. I'd add a comment to explain why we have an empty migration (that could be confusing otherwise). Maybe something like:

> See git history of this file for why this migration is empty

bmispelon:
The test failure looks like it might have something to do with `tracdb.test_utils.TracDBCreateDatabaseMixin`. It might be a fluke, but it might also be genuine (it could be that the mixin relied on an internal Django API that was modified).

You might be able to reproduce the failure locally by running all tests (not just a subset).

ulgens:
I'm trying to reproduce it locally but turned out creating the local env was harder than I remember, and docker setup is not ready for local testing. When I find a quick solution on this, I'll be creating several other issues 🙂

ulgens:
It looks like the errors are somehow related to composite PKs, but I thought it was merged after 5.1, and I couldn't find anything related on the 5.1 changelog. I'll try to tackle it another day, but if it's really about composite PKs, I guess it will require bigger changes than I thought.

pauloxnet:
> I'm trying to reproduce it locally but turned out creating the local env was harder than I remember, ...

Please open an issue for the Readme if you found that something is missing or wrong.

Write here if you need help in the local setup.

bmispelon:
Using `git bisect` on the Django repo, I found that our tests started breaking after this change: https://github.com/django/django/commit/e65deb7d14a57ba788b978d50bd8198e659faa91

This points further to something being broken with the `TracDBCreateDatabaseMixin` hack. I'll look further into it. 🔬

bmispelon:
Don't ask me how it works, but I just pushed a commit that fixes the tests 😅

ulgens:
I feel like we pushed the bug into the 5.2 update 😬 but I'm okay with passing tests for now ✨

bmispelon:
> I feel like we pushed the bug into the 5.2 update 😬 but I'm okay with passing tests for now ✨

I tried and the `tracdb` tests still pass with 5.2 so I think we should be good on that front. And once we have 5.2 we should be able to get rid of all the hacks and use real composite primary keys I think 🎉 

I think all is good with this PR now, I'm going to squash all the commits then merge, let's see how the deploy goes 🤞🏻

accounts/migrations/0002_migrate_sha1_passwords.py
ulgens: Thanks for adding the comment.

PR #1793, opened 2024-12-04
Make test fail

bmispelon:
This is a throwaway PR to test a strange behavior observed in the `unveil` PR (see link under)


bmispelon:
Figured it out, see https://github.com/django/djangoproject.com/pull/1784#issuecomment-2521515017

PR #1794, opened 2024-12-05
Remove unneeded `getElementsByClassName` polyfill

adamzap:
The `document.getElementsByClassName` method has been a standard browser feature for many years, so we no longer need a polyfill for it. See https://caniuse.com/getelementsbyclassname.

It may be worth refactoring this file a bit in a future pull request. I'll add that to my list to evaluate.


bmispelon (approved):
10/10, no notes.

I'm loving all these cleanup PRs, keep'em coming! 💯

PR #1796, opened 2024-12-06
Remove unused `jquery.payment` JavaScript library

adamzap:
It looks like the removal of this library was missed in e51eba97. Its only use was in `fundraising-donation.js`, which was removed in that commit.

`jquery.payment` was archived by Stripe in early 2018: https://github.com/stripe-archive/jquery.payment.


bmispelon (approved):
Nice catch! It's always nice when 2k lines of dead unmaintained code disapear from your project 😁

PR #1798, opened 2024-12-06
Fixed #1765 -- Actually validate signature of stripe webook requests

### LINKED ISSUE #1765: Stripe event handling doesn't verify the source
https://github.com/django/djangoproject.com/blob/5537f965ece29b766efe17c1d2137c3d2037811c/fundraising/views.py#L196-L201
Instead of calling Stripe with the event id it should verify the source of the event and prevent replay attacks as indicated here
https://docs.stripe.com/webhooks#verify-events

The Stripe library already has a method to do this https://docs.stripe.com/webhooks?lang=python#verify-webhook-signatures-with-official-libraries

bmispelon:


bmispelon:
I believe I've figured out what was causing the failures we saw in production (I misconfigured the loading the new `stripe_endpoint_secret` key in `secrets.json`).

I'm going to give this another go.

PR #1799, opened 2024-12-06
Use CSS instead of JavaScript for floating warning

adamzap:
Some of the functionality of the floating warning on unsupported and in-development documentation versions was implemented using JavaScript. This is longer necessary with `position: sticky` in CSS. This feature is widely-supported: https://caniuse.com/css-sticky.

Other than setting `position: sticky` and removing the related JavaScript, this patch does the following:

- Simplify the CSS so that the class on all floating warning elements implements the common rules and the CSS for the IDs on the elements implements the colors
  - A `z-index` value was added to the floating warning element because [the banner element in the header has one specified](https://github.com/django/djangoproject.com/blob/429a48d6450972d1369d27309f02b39696c3800e/djangoproject/scss/_style.scss#L458)
- Move the floating warning element to the top of the page in the HTML
  - Remove the `body_extra` block that was only being used by the floating warning element
  - Add a new `before_header` block for the floating warning element


bmispelon:
Heya 👋🏻 

I agree it would be nice to be able to use standard CSS features for this, but there's a few bugs/features that are missing:

1) The new sticky banner only seem to appear when the page is scrolled and is not visible when the page loads, or when you scroll to the top of the page.
2) There's some z-index issues with some elements appearing over the sticky banner (like the navigation header and the search bar):
djangoproject_pr_screenshot
3) The previous `floating-warning`  implementation had some extra logic to avoid the sticky banner covering headings when navigating to them: https://github.com/django/djangoproject.com/blob/429a48d6450972d1369d27309f02b39696c3800e/djangoproject/static/js/mod/floating-warning.js#L28-L40 (though there were still some issues with that implementation, see #1624)

Is there a way to address there while keeping the CSS-only banner (3 might be tricky, but I have the feeling that the other two should be doable).

Thanks! 🎸

adamzap:
Which browser are you using? I'm unable to reproduce the first two issues you mentioned above.

bmispelon:
> Which browser are you using? I'm having a hard time reproducing the first two issues you mentioned above.

Firefox

adamzap:
Hmmm ok. It looks good to me on Firefox on macOS. Do you think this PR is worth pursuing further? Your third point is a good one, and we'd have to add a good bit of the JavaScript back to fix it. I mean lean toward closing it?

image
.

bmispelon:
If we're going to have to re-add or keep a lot of the js, then maybe it's worth focusing on other (easier) cleanups. We can always come back to this in the future.

I don't have the time to investigate this right now, so you would be the one doing the work. If you'd rather work on something else then I think it's totally fine to close this PR.

I really appreciate all the work you've been doing so far. Cutting down on needless jaavascript and css is making the site better for both our users and for us maintainers. Everybody wins 🏆

adamzap:
I agree. Thanks for the kind words!

PR #1800, opened 2024-12-07
Fix initial position of two icons on overview page

adamzap:
The initial `background-position`s of the lock and dashboard icons on [the overview page](https://www.djangoproject.com/start/overview/) were not hiding them like the other icons. This patch adjusts the initial `background-position` values so that the icons are out of view when the animation starts.

The bug is subtle in production. You can see it in detail if you remove the following lines:

https://github.com/django/djangoproject.com/blob/429a48d6450972d1369d27309f02b39696c3800e/djangoproject/static/js/main.js#L12-L15

Below is the error state for the two icons whose CSS is being adjusted. The other icons are not visible here.

Screenshot 2024-12-06 at 9 36 44 PM

### DIFF
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -1752,7 +1752,7 @@ h2+.list-link-soup {
             }
 
             &.icon-lock {
-                background-position: 144px -206px;
+                background-position: 148px -206px;
 
                 &.inview {
                     background-position: -36px -96px;
@@ -1760,7 +1760,7 @@ h2+.list-link-soup {
             }
 
             &.icon-dashboard {
-                background-position: -360px 201px;
+                background-position: -367px 201px;
 
                 &.inview {
                     background-position: -270px -9px;


bmispelon:
Thanks for the report, that's very subtle indeed 😅 

I can see the issue you're referring to on the `icon-lock` element, but not `icon-dashboard`, here's how it looks like for me when I disable the `mod/list-feature` like you suggest:
Screenshot 2024-12-07 at 19-55-07 Django overview Django

Also, how did you figure out the correct value for `background-position`? Was it just a matter of trial an error?

adamzap:
Oh, weird! I'm not sure why it's different. That makes me a bit less confident in this PR.

To get the new values, I just nudged them further in the direction they were going until the icons were no longer visible.

bmispelon:
To be honest I'm struggling to fully understand how the CSS works, or how those magic values were chosen in the first place.

I also find it strange that the `background-position` seems to be defined twice, consider:
https://github.com/django/djangoproject.com/blob/429a48d6450972d1369d27309f02b39696c3800e/djangoproject/scss/_style.scss#L1711-L1713
and
https://github.com/django/djangoproject.com/blob/429a48d6450972d1369d27309f02b39696c3800e/djangoproject/scss/_style.scss#L1738-L1744

adamzap:
I'm confused by that as well. I guess it's just moving [the background SVG](https://static.djangoproject.com/img/bg-features.a10a0100c8c0.svg) around in different directions from different starting points.

I saw the redefinition as well but was afraid to remove it because if you try each of them in the inspector, they each lead to a different result. I'm not sure what the original intent was, and the SCSS further obscures things for me.

I would personally prefer a solid color circle like on the homepage, but that's not my call to make! 😉

bmispelon (approved):
Even though we seem to be observing two different bahaviors, the new result is a net improvement for both of us:

* For you, the two changed icons were broken before and fixed after
* For me, one icon was broken before and fixed after, and the other was not broken before and is still fine after.

So I think I will merge this as-is so we can move on with our lives and forget about this confusing bit of CSS whose intent is lost to time 😁

adamzap:
Great, thanks!

PR #1802, opened 2024-12-07
Removed obsolete version element from compose file.

pbratkowski:
As per <https://docs.docker.com/reference/compose-file/version-and-name/#version-top-level-element-obsolete>:

> The top-level `version` property is defined by the Compose Specification for backward compatibility. It is only informative and you'll receive a warning message that it is obsolete if used.

Now that the README refers to Compose V2, I believe it makes sense to remove the reference to the obsolete value, and suppress the warning.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,3 @@
-version: "3.7"
-
 services:
     web:
         build:


bmispelon (approved):
Good call, thanks for the PR! 🎸 

Would you mind opening another one for https://github.com/django/code.djangoproject.com/blob/main/docker-compose.yml too?

Thank!

pbratkowski:
Sure thing, done :)

PR #1803, opened 2024-12-07
Remove `jquery.inview` JavaScript library

adamzap:
The `jquery.inview` library was being used to trigger an animation when an icon in a feature list element is fully visible on screen. This patch replaces `jquery.inview` with an implementation based on `IntersectionObserver`, which is widely supported: https://caniuse.com/intersectionobserver.

This patch also simplifies the structure of `list-feature.js`.


djangoproject/static/js/mod/list-feature.js
adamzap, line 1: Let me know if anything should change here style-wise.

adamzap:
To verify that this works, you'll want to check that the icons on the homepage and the overview page animate as before:

image

image

djangoproject/static/js/mod/list-feature.js
```diff
@@ -1,23 +1,18 @@
...
-                }
-            });
-        }
-    };
+            observer.unobserve(entry.target);
+        });
 
-    // Export a single instance of our module:
-    return new FeatureList('.list-features');
+    }, { threshold: 1.0 });
```
bmispelon, line 13: A comment similar to the previous `// element completely visible` would be welcome here (had to look up the specs to understand what `threshold` meant)

djangoproject/static/js/mod/list-feature.js
```diff
@@ -1,23 +1,18 @@
...
+
+    $('.list-features').find('i').each(function () {
```
bmispelon, line 15: Isn't this equivalent to:
```suggestion
    $('.list-features i').each(function () {
```
?

djangoproject/static/js/mod/list-feature.js
bmispelon: No comment on the style, other than I think we should use an autoformatter so we don't have to think about it (I thought the project had `prettier` configured in pre-commit, but maybe it's not run on js files? 🤔 )

That's out of scope for this PR though.

bmispelon (commented):
Another really nice cleanup 🚀 

It's pretty amazing to see how far the web has come in the 10 years since this design was made. All these jquery plugins are now either not needed or can be replaced by native APIs.

I just had two very minor comments. The functionality is indeed working. 👍🏻

djangoproject/static/js/mod/list-feature.js
adamzap: Oh yes, thanks!

adamzap: Fixed in b24f10ba.

djangoproject/static/js/mod/list-feature.js
adamzap: Fixed in 7b85d55d.

bmispelon (approved):
Excellent 👍🏻

PR #1807, opened 2024-12-08
fix #1727: Replace reference to mailing list with Django Forum

### LINKED ISSUE #1727: Retire django-developers and django-users mailing list
Following the forum thread ([Proposal: retiring django-users and django-developers mailing lists](https://forum.djangoproject.com/t/proposal-retiring-django-users-and-django-developers-mailing-lists/35747)), we can now move ahead with updating the website to remove or contextualize references to the django-developers and django-users mailing list. Corresponding Django ticket: [Retire django-developers and django-users mailing list django#35908](https://code.djangoproject.com/ticket/35908)

## Tasks

Here are relevant changes I think we should go ahead with:

- [x] In the footer, remove the `Django Users Mailing List` item
- [x] On the [Community](https://www.djangoproject.com/community/) page, remove the "Mailing Lists" section
- [x] On [Download / How to get Django](https://www.djangoproject.com/download/), replace "be sure to sign up for the django-users mailing list" with "be sure to sign up to the Django Forum"
- [x] On the [homepage](https://www.djangoproject.com/), remove the "Stay in the loop" section
- [ ] CoC-related pages

ontowhee:
fix #1727

Text changes for the page https://www.djangoproject.com/foundation/individual-members/

Before
Screenshot 2024-12-08 at 9 37 29 AM

After
Screenshot 2024-12-08 at 9 49 24 AM

### DIFF
--- a/djangoproject/templates/members/individualmember_list.html
+++ b/djangoproject/templates/members/individualmember_list.html
@@ -65,7 +65,7 @@
 
   <p>
     {% blocktranslate trimmed %}
-      As a member of the DSF, you will be recognized for your contributions to the community. Your name will appear below and you'll be added to the DSF Members Forum and Discord channel. You will also be eligible to vote for the DSF Board and Steering Council.
+      As a member of the DSF, you will be recognized for your contributions to the community. Your name will appear below and you'll be added to the various DSF Members communication channels (mailing list, forum, Discord). You will also be eligible to vote for the DSF Board and Steering Council.
     {% endblocktranslate %}
   </p>
 
--- a/djangoproject/templates/members/individualmember_list.html
+++ b/djangoproject/templates/members/individualmember_list.html
@@ -65,7 +65,7 @@
 
   <p>
     {% blocktranslate trimmed %}
-      As a member of the DSF, you will be recognized for your contributions to the community. Your name will appear below and you'll be added to the DSF Members Mailing list. You will also be eligible to vote for the DSF Board and Steering Council.
+      As a member of the DSF, you will be recognized for your contributions to the community. Your name will appear below and you'll be added to the DSF Members Forum and Discord channel. You will also be eligible to vote for the DSF Board and Steering Council.
     {% endblocktranslate %}
   </p>
 


bmispelon (approved):
Changes look good to me, but I'm going to ask for another review from someone from the DSF board just to be safe.

Thanks! 🎸

thibaudcolas:
Thank you! I’ll switch the review to @sabderemane specifically as in her new role as Vice President (:tada:) she is taking charge of management of Individual Members. Then Sarah can decide if/when to raise this with other board members.

djangoproject/templates/members/individualmember_list.html
```diff
@@ -65,7 +65,7 @@ <h1>{% translate "Individual members" %}</h1>
 
   <p>
     {% blocktranslate trimmed %}
-      As a member of the DSF, you will be recognized for your contributions to the community. Your name will appear below and you'll be added to the DSF Members Mailing list. You will also be eligible to vote for the DSF Board and Steering Council.
+      As a member of the DSF, you will be recognized for your contributions to the community. Your name will appear below and you'll be added to the DSF Members Forum and Discord channel. You will also be eligible to vote for the DSF Board and Steering Council.
```
thibaudcolas, line 68: I think this needs a different wording. As far as I know we still do add people to the DSF Members mailing list, which we’ve kept even if it’s no longer used for conversations.

- For the members-only Forum section – this should definitely be added here as it’s intended as the primary communication channel for members, but I believe it’s "opt-in" for users to decide if they want to be added there, rather than something people are automatically added to.
- For the Discord channel, it’s definitely "opt-in" currently.

bmispelon: What about:
```suggestion
      As a member of the DSF, you will be recognized for your contributions to the community. Your name will appear below and you'll be added to the various DSF Members communication channels (mailing list, forum Discord). You will also be eligible to vote for the DSF Board and Steering Council.
```
?

thibaudcolas: I’ll defer to @sabderemane :) don’t have a strong opinion nor lots of awareness of how this works personally.

thibaudcolas (approved):
Coming back to this – I think @sabderemane might further change the wording, but this is definitely better than what we had before! Thank you both.

PR #1809, opened 2024-12-08
Remove `clipboard` JavaScript library

adamzap:
The `clipboard` JavaScript library was being used to copy code snippets to the user's clipboard when a button is clicked. This patch replaces it with the built-in [`navigator.clipboard.writeText`](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText) API, which is widely available: https://caniuse.com/mdn-api_clipboard_writetext.

I tested this in Brave (Chromium), Firefox, and Safari on macOS, and it works as expected.


djangoproject/static/js/mod/clippify.js
```diff
@@ -18,20 +18,20 @@ define(['jquery', 'clipboard'], function($, Clipboard) {
...
-        success.prependTo(e.trigger).delay(1000).fadeOut();
-    });
-    clip.on('error', function(e) {
-        // Safari doesn't support the execCommand (yet) but because clipboardjs
-        // also uses Selection API, we can instruct users to just press the keyboard shortcut
-        // See https://clipboardjs.com/#browser-support
-        var success = $('<span class="clipboard-success">').text('Press ⌘-C to copy');
-        success.prependTo(e.trigger).delay(5000).fadeOut();
+
+        function on_error(el) {
```
adamzap, line 30: I'm not sure if this error handler is strictly needed any longer, but it feels a bit unwise to remove it. It's definitely not needed for the same reason as the removed comment details, but I suppose a user could disable clipboard access in their browser.

djangoproject/static/js/mod/clippify.js
adamzap, line 1: I tried to change as little code in this file as possible. As always, let me know if any style tweaks are needed.

djangoproject/static/js/mod/clippify.js
bmispelon: I don't think that this error handler is accurate anymore since the instruction to the user "Press ⌘-C to copy" will not work.

It was accurate before because clipboardjs had a fallback behavior of selecting the text to be copied (I think), and so instructing the user to press the "copy" keyboard shortcut worked.
But the "select text on error" fallback doesn't exist anymore, so instructing the user to press ⌘-C will not work.

I suggest turning this into a generic error instead, maybe something like "Error: clipboard disabled"? (couldn't come up with something better)

adamzap: Oh sorry, I missed that. I could try to add code to select the text for the user, but that might be overkill. What do you think?

bmispelon: That doesn't seem worth it, seeing how well the modern clipboard API is supported.
I think a generic error message is enough.

adamzap: Ok cool. My suggestion would be "Could not copy!", but it's up to you. Do you like "Error: clipboard disabled" better? I'm not sure if that's the only reason the feature can fail.

adamzap: I pushed 2423433f to update the message. You can change it if you want before you merge, or I can change if you'd rather. Thanks!

bmispelon (approved):
Excellent, thanks! 🎸

PR #1813, opened 2024-12-09
fix #1806: N+1 query issue on dashboard index page

### LINKED ISSUE #1806: [dashbaord] Investigate and fix N+1 query issue
I noticed today that Sentry was reporting a possible N+1 query issue on the [dashboard homepage](https://dashboard.djangoproject.com).
Here's the query that Sentry is claiming is being repeated:
```
SELECT dashboard_datum.id, dashboard_datum.content_type_id, dashboard_datum.object_id,
  dashboard_datum.timestamp, dashboard_datum.measurement
FROM dashboard_datum
WHERE (
  dashboard_datum.content_type_id = %s AND dashboard_datum.object_id = %s
)
ORDER BY dashboard_datum.timestamp DESC
LIMIT 1
``` 

I'd recomment trying to see ifthe Django Debug Toolbar (installed by default I think) is reporting the duplicated query also. And if so,  we should try to fix it. Usually it's a matter of adding a `select_related()` or `prefetch_related()` somewhere in the view, but it might also be more complicated than that.

ontowhee:
fix #1806 

The existing query was looping on metrics to grab the latest Datum for each metric, which led to 18 similar queries reported by django-debug-toolbar. There were 45 sql queries overall for the page.

Screenshot 2024-12-09 at 7 33 50 AM

This is WIP. The change here reduces the number of queries down to 1 large statement of unions of each metric's latest datum. There are 9 sql queries overall for the page.

Changes include:
- Query on the content type for each metric, which is used to build the union.
- Query on Datum by building a union of all the metric querysets.
- Adding a `select_related('category')` for the metric query that is used for sorting based on `display_position`.

Screenshot 2024-12-10 at 10 28 08 AM

### DIFF
--- a/dashboard/tests.py
+++ b/dashboard/tests.py
@@ -17,6 +17,7 @@ from tracdb.tractime import datetime_to_timestamp
 from .models import (
     METRIC_PERIOD_DAILY,
     METRIC_PERIOD_WEEKLY,
+    Category,
     GithubItemCountMetric,
     GitHubSearchCountMetric,
     Metric,
@@ -69,6 +70,53 @@ class ViewTests(ReleaseMixin, TestCase):
         response = metric_json(request, "new-tickets-week")
         self.assertEqual(json.loads(response.content.decode())["data"][0][1], 42)
         self.assertEqual(response.status_code, 200)
+
+
+class AbstractMetricTestCase(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        category = Category.objects.create(name="test category")
+        cls.metrics = [
+            TracTicketMetric.objects.create(
+                slug=f"test{i}", name=f"Test metric {i}", category=category
+            )
+            for i in range(3)
+        ]
+        for metric, measurement, year in [
+            (0, 1, 2020),
+            (0, 2, 2021),
+            (0, 3, 2022),
+            (1, 4, 2023),
+        ]:
+            cls.metrics[metric].data.create(
+                measurement=measurement,
+                timestamp=datetime.datetime(year, 1, 1),
+            )
+
+    def test_with_latest(self):
+        self.assertQuerySetEqual(
+            TracTicketMetric.objects.with_latest().order_by("name"),
+            [
+                (
+                    "Test metric 0",
+                    {"measurement": 3, "timestamp": "2022-01-01T00:00:00-06:00"},
+                ),
+                (
+                    "Test metric 1",
+                    {"measurement": 4, "timestamp": "2023-01-01T00:00:00-06:00"},
+                ),
+                ("Test metric 2", None),
+            ],
+            transform=attrgetter("name", "latest"),
+        )
+
+    def test_for_dashboard(self):
+        with self.assertNumQueries(1):
+            for row in TracTicketMetric.objects.for_dashboard():
+                # weird asserts to make sure the related objects are evaluated
+                self.assertTrue(row.category.name)
+                self.assertTrue(row.latest is None or row.latest["timestamp"])
+                self.assertTrue(row.latest is None or row.latest["measurement"])
 
 
 class MetricMixin:
--- a/dashboard/views.py
+++ b/dashboard/views.py
@@ -17,17 +17,23 @@ def index(request):
 
     data = cache.get(key, version=generation)
     if data is None:
-        metrics = []
-        for MC in Metric.__subclasses__():
-            metrics.extend(MC.objects.filter(show_on_dashboard=True))
-        metrics = sorted(metrics, key=operator.attrgetter("display_position"))
-
-        data = []
-        for metric in metrics:
-            data.append({"metric": metric, "latest": metric.data.latest()})
+        data = [m for MC in Metric.__subclasses__() for m in MC.objects.for_dashboard()]
+        data.sort(key=operator.attrgetter("display_position"))
         cache.set(key, data, 60 * 60, version=generation)
 
-    return render(request, "dashboard/index.html", {"data": data})
+    # Due to the way `with_latest()` is implemented, the timestamps we get back
+    # are actually strings (because JSON) so they need converting to proper
+    # datetime objects first.
+    timestamps = [
+        datetime.datetime.fromisoformat(m.latest["timestamp"])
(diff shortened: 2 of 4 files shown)


dashboard/views.py
```diff
@@ -19,12 +19,28 @@ def index(request):
...
+        latest_datums = {
+            (datum.object_id, datum.content_type_id): datum for datum in datum_queryset
+        }
 
         data = []
-        for metric in metrics:
-            data.append({"metric": metric, "latest": metric.data.latest()})
+        for metric, content_type in content_types.items():
+            if latest := latest_datums.get((metric.id, content_type.id)):
+                data.append({"metric": metric, "latest": latest})
```
ontowhee, line 42: Is there a better way to write this?

dashboard/tests.py
```diff
@@ -33,10 +33,12 @@ def setUp(self):
     def test_index(self):
         for MC in Metric.__subclasses__():
             for metric in MC.objects.filter(show_on_dashboard=True):
+                metric.data.create(measurement=44)
```
ontowhee, line 36: Adding extra data to ensure the query will retrieve the latest datum for each metric.

dashboard/views.py
bmispelon: I've been trying to find an alternate approach (got pretty close with a `Datum.objects.values('content_type', 'object_id').annotate(Max('timestamp'))` but unfortunately that can only get the latest timestamp for each metric, not the associated measurement).

Your original approach using `UNION` seems quite good (and it's quite clever).

I've made a suggestion for a slight rewrite that is a bit shorter and also closer to the original view implementation (it does so by pushing some logic onto the `Datum` manager). If you like it I can push a commit onto this branch.

ontowhee: Sounds good! It's very interesting how you've moved most of the logic to `DatumQuerySet`!

The only concern is that `latest = latest_by_metric.get((metric.content_type.pk, metric.pk))` could return None. Will the frontend be able to handle rendering of None?

bmispelon: Well, the current behavior for a metric with no data is for the view to crash, so it can't be worse than that 😁 
(good call though, I'll test it out)

bmispelon:
@ontowhee It took a few tries, but I think I finally landed on something that I think I like. I moved a lot of the logic onto manager methods so I could simplify the view (and it also simplified the N+1 tests sinceI could test directly against the manager method rather thant involve the view).

I managed to switch to a subquery instead of a union which I think simplifies the code, but I don't know if the performance is better/worse/equivalent. Do you know anything about that?

I also added an explicit `last_updated` into the template to deal with metrics that have no data. That's not really supposed to happen in production, but I did run into a crash a few times when I was testing things locally.

Let me know what you think, if you're OK with it I'll merge all commits on this branch into one, with both of us as co-authors.

ontowhee:
For commit [8fe1e01](https://github.com/django/djangoproject.com/pull/1813/commits/8fe1e0171ca3c1406ab9d452e5fdc6b1fe841ac5), the overall performance was 110ms.
 
Screenshot 2024-12-12 at 10 41 36 PM

For commit [2288df6](https://github.com/django/djangoproject.com/pull/1813/commits/2288df6adf8fa22b47fc48f0ae9adf0760e5aae0), using the subquery seems to be worse for TracTicketMetric. It took 186ms to query that table, and overall it took 224ms. Will this performance be noticeable by the end user or the server?

Screenshot 2024-12-12 at 10 42 17 PM

Django debug toolbar also identifies "4 similar" queries on ContentType due to calling .for_dashboard() in the loop. Will Sentry flag these 4 similar queries as an N+1 issue?

bmispelon:
Thanks for doing the measurements! The difference between 110 and 186ms doesn't seem so bad to me, but I would like to know your opinion too.

> Django debug toolbar also identifies "4 similar" queries on ContentType due to calling .for_dashboard() in the loop. Will Sentry flag these 4 similar queries as an N+1 issue?
Does that warning persist after a page refresh? Django should cache the content type globally (across requests), so I would expect those queries in the loop to happen only on the first load, but not once the cache has been populated.

ontowhee:
> The difference between 110 and 186ms doesn't seem so bad to me, but I would like to know your opinion too.

It doesn't seem too bad to me either. I wasn't sure how similar the local database was to production. I noticed that the results are returned in much less time in production too.

> Django should cache the content type globally (across requests), so I would expect those queries in the loop to happen only on the first load, but not once the cache has been populated.

You're right, on first load it shows the warning, but on refresh, the content type queries are cached along with all the other queries. (0 queries!)

I think this is good to merge! Thanks so much for refactoring this. I learned so much from your code!!

ontowhee:
I was playing around to understand the queries better. Here's another [approach](https://github.com/ontowhee/djangoproject.com/compare/fix-1806-dashboard-n-plus-1-query...ontowhee:djangoproject.com:fix-1806-alternate-query-datum?expand=1) that queries on the Datum model. It uses `.distinct()` and `.order_by()` to get the latest measurement for each metric, and it uses `prefetch_related()` to fetch the metric. (I didn't know prefetch_related could be called on generic foreign keys.) This avoids making subqueries, and it also avoids building the unions or Q expressions that were done in the previous approaches.

Not asking for any changes here. I think your approach of querying on the Metric is easier to understand, and it’s already an improvement from the current dashboard!

sabderemane:
Hey @ontowhee ! Would it be possible to fix the merge conflict and eventually merge all commits into one? (And add Baptiste as co-author 😄 )
It seems everyone is in agreement to merge this as an improvement, that would be great to have this merged. Happy to review the history if you don't feel confident on this.

ontowhee:
I've resolved the merge conflict, squashed the commits, and added Baptiste as the co-author!

The failing test is making an external api call, and looks like several other PRs are encountering it as well.

adamzap (approved):
👍 This looks reasonable to me, and I'm seeing an nice performance increase locally.

My only concern is that the tests are not checking the `show_on_dashboard` filtering that happens in the `MetricQuerySet.for_dashboard` method. However, given the history and age of this PR, I will merge it and potentially open another PR to enhance the tests.

Issue #1003, opened 2020-06-05
Remove Google Fonts


stefins:
Would you be willing to remove this Google Fonts  from this website? There are better alternatives available 
Some reputable alternatives can be seen here: https://switching.software/replace/google-fonts/

TheEvilSkeleton:
i second this

resynth1943:
I third this.

taminaru:
I fourth this

bobycloud:
I fifth this.

arthtyagi:
Google fonts API is actually nice.

resynth1943:
Sure, it's nice, but it's always better to self-host fonts, instead of injecting code [from a third party](https://github.com/google/fonts/issues/2345).

resynth1943:
#1019 😉

knyghty:
Well, okay, I had a quick look at system fonts but for some areas it looks bad. See screens. I am not sure about the nav bar, I think it's maybe an improvement but much longer using my system font (presumably SF Pro), but the big banner text "Django makes it easier ..." looks really bad. Probably we should choose a different font for headers, ideally something that is installed on (most) people's machines already but maybe something that we just use as a web font (could use a downloaded Roboto, but seems annoying just for headers).

Or maybe people think it looks fine, I dunno.

Before:
<img width="1509" alt="Screenshot 2022-10-09 at 11 51 29" src="https://user-images.githubusercontent.com/3871354/194750130-d69659ee-5d4f-4376-9d10-e10d2395d450.png">

After:
<img width="1512" alt="Screenshot 2022-10-09 at 11 51 40" src="https://user-images.githubusercontent.com/3871354/194750143-a7c7b3fc-0f44-4e6f-b136-fed3d3c48e09.png">

pauloxnet:
The screenshot without Google Fonts looks fine for me.

For my blog I used PicoCss which is ultralight and does not use external fonts.

Here is an extract from its CSS in case it may be of interest.
```css
  --font-family: system-ui, -apple-system, "Segoe UI", "Roboto", "Ubuntu",
    "Cantarell", "Noto Sans", sans-serif, "Apple Color Emoji", "Segoe UI Emoji",
    "Segoe UI Symbol", "Noto Color Emoji";
```
https://github.com/picocss/pico/blob/master/css/pico.classless.css#L10

knyghty:
Okay, if you're happy with it I'll press on 👍🏻

knyghty:
[...]

PR #1815, opened 2024-12-10
Remove `webfontloader` and simplify web fonts

adamzap:
This patch removes the `webfontloader` JavaScript library in favor of using `link` tags to load web font assets. `webfontloader` had its last release in mid-2017, and we were using it in a very simple way.

This patch also makes the following related changes:

- Remove all font assets other than `woff`, which is widely supported: https://caniuse.com/woff
- Flatten font assets into `djangoproject/static/fonts/`
- Remove the `FiraMono-Medium` font variant assets since this variant was not being used
- Update outdated font-related references in the `styleguide.html` template

@bmispelon This is a particularly exciting PR for me because it drops ~7.7 megabytes from the repository. The diff is huge, but the actual changes are small.

If this gets merged, some nice follow-up changes could be:

1. Update fonts to their latest versions
2. Use `.woff2` font assets instead of `.woff` ([it's safe](https://caniuse.com/woff2))
3. Self-host the Roboto font since [Google open-sourced it](https://opensource.googleblog.com/2015/05/roboto-googles-signature-font-is-now.html)? Would this be ok license-wise? #1003 is kind of related.

If there's anything I can do to help test this, please let me know. It looks good to me locally.


djangoproject/templates/base.html
```diff
@@ -39,22 +39,17 @@
 
     <title>{% block title %}The web framework for perfectionists with deadlines{% endblock %} | Django</title>
 
+    <link href="{% static "fonts/FiraMono-Regular.woff" %}" as="font" type="font/woff" crossorigin="anonymous" />
+    <link href="{% static "fonts/FiraMono-Bold.woff" %}" as="font" type="font/woff" crossorigin="anonymous" />
+
+    <link href="{% static "fonts/fontawesome-webfont.woff" %}" as="font" type="font/woff" crossorigin="anonymous" />
```
adamzap, line 45: Here's my reference from https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link:

```html
<link
  rel="preload"
  href="myFont.woff2"
  as="font"
  type="font/woff2"
  crossorigin="anonymous" />
```

I didn't use the `rel=preload` attribute because `FiraMono` isn't needed on most non-documentation pages, and `fontawesome` isn't needed on most documentation pages. We could optimize this, but I personally don't think it's worth it. I strongly doubt that `webfontloader` was doing that kind of optimization.

djangoproject/templates/base.html
```diff
@@ -39,22 +39,17 @@
...
+
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,300;0,400;0,700;1,400;1,700&display=swap" rel="stylesheet">
```
adamzap, line 49: These three links tags are from Google's embedding instructions:

1. Go to https://fonts.google.com/specimen/Roboto
2. Click "Get Font"
3. Click "Get Embed Code"

The variants in the URL above should match what `webfontloader` was asking for.

djangoproject/templates/styleguide.html
adamzap, line 1: [This file](https://www.djangoproject.com/styleguide/) may be a bit outdated in general, but grepping led to these small changes.

djangoproject/templates/base.html
bmispelon: Doubled check and got the same result 👍🏻

djangoproject/templates/styleguide.html
bmispelon: Yes, this file needs some love. I'm OK with those two minor changes 👍🏻

bmispelon (approved):
What can I say, I wish all PRs were like this 💚 

20k lines deleted, clear instructions on how to verify the changes, links to reference documentation: this is a pleasure and a breeze to review.

A+, would merge again 💯

bmispelon:
> If this gets merged, some nice follow-up changes could be:
> 
>     1. Update fonts to their latest versions
> 
>     2. Use `.woff2` font assets instead of `.woff` ([it's safe](https://caniuse.com/woff2))
> 
>     3. Self-host the Roboto font since [Google open-sourced it](https://opensource.googleblog.com/2015/05/roboto-googles-signature-font-is-now.html)? Would this be ok license-wise?

Yes, yes, and yes (please) 😁

PR #1820, opened 2024-12-11
Fixed #1756 -- hide unpublished events from side menus for staff

### LINKED ISSUE #1756: Unpublished Events show up for staff in side menus
This is related to #1755, but I would classify this issue as a **bug** because there is no visual way to tell the difference between an unpublished event and a published event. 

Please see this screenshot and notice that DjangoCon Africa has not been published from the admin yet. (**Please note:** They have already announced their dates and location, but they haven't blogged about it yet in case anyone is wondering.)

Screenshot 2024-11-20 at 1 51 34 PM

See the admin:
Screenshot 2024-11-20 at 3 15 10 PM

ontowhee:
Fixed #1756. Only return published events, even for staff users.

Screenshot 2024-12-10 at 10 43 10 PM
Screenshot 2024-12-10 at 11 14 13 PM


blog/tests.py
```diff
@@ -191,6 +191,28 @@ def test_no_past_upcoming_events(self):
...
+        )
+        Event.objects.create(
+            date=self.tomorrow,
+            pub_date=self.tomorrow,
+            is_active=True,
+            headline="future publish date",
+        )
+        response = self.client.get(reverse("weblog:index"))
+        self.assertEqual(response.status_code, 200)
+        self.assertQuerySetEqual(response.context["events"], [])
```
bmispelon, line 214: This new test already passes with the old version of the code, because it's not useing a logged-in staff user.
I suggest to test things with three types of users: anonymous (not logged in), logged in (not staff), and logged in (staff). Something like this should work (not tested):

```suggestion
        for user in [
            None,
            User.objects.create(is_staff=False),
            User.objects.create(is_staff=True),
        ]:
            if user:
                self.client.force_login(user)
        response = self.client.get(reverse("weblog:index"))
        with self.subTest(user=user):
            self.assertEqual(response.status_code, 200)
            self.assertQuerySetEqual(response.context["events"], [])
```

bmispelon (changes requested):
Thanks for the PR, nice work! 🎸 

The code change looks good, I just had a note about the test (I like the use of `assertQuerySetEqual` on `response.context["events"]` by the way, that seems like a robust and effective way to test what we want here 👍🏻 ).

blog/tests.py
ontowhee: Thank you for these tests! I didn't think to check the anonymous, staff, and non-staff users. I've added these cases along with a superuser for good measure!

jefftriplett:
@ontowhee awesome work. Thank you!

PR #1822, opened 2024-12-11
Added default silenced system checks to dev

pbratkowski:
After the setting was deleted and added anew, it overrides the original rather than extending it. This adds the default silenced checks back.

### DIFF
--- a/djangoproject/settings/dev.py
+++ b/djangoproject/settings/dev.py
@@ -63,4 +63,6 @@ if DEBUG:
             "djangoproject.middleware.CORSMiddleware",
         )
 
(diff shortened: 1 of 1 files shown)


bmispelon (approved):
It's hard to interpret the original intent of the overriding (rather than extending) the silenced checks. I think your proposed change make sense, since the `common.py` settings file should - as the name imply - contain settings that are common to all envs.
If some checks should only be silenced in prod or docker but not in dev, then I think those should be moved to the `prod.py` or `docker.py` settings files.

Thanks for flagging this and quickly submitting a PR! 👍🏻

PR #1808, opened 2024-12-08
fix #1727: Remove reference to IRC and django-users and django-developers mailing lists in CoC

ontowhee:
Fix #1727

Text changes for the page (https://www.djangoproject.com/conduct/faq/)

Before
Screenshot 2024-12-08 at 10 02 03 AM

After
Screenshot 2024-12-08 at 10 18 54 AM

PR #1823, opened 2024-12-12
Removed IRC from example Django spaces for CoC violations.

sarahboyce:
Following https://forum.djangoproject.com/t/proposal-retire-irc/37129 and https://code.djangoproject.com/ticket/35999, I suggest that IRC is not used as an example for CoC violations.
Although it is still valid, we should use a space which is more active (and more likely to have a CoC violation)

We may want to combine these changes with https://github.com/django/djangoproject.com/pull/1808


sarahboyce:
These changes will be included in https://github.com/django/djangoproject.com/pull/1808

PR #1824, opened 2024-12-12
Fixed #1816 -- Fixed wrong aspect ratio of supporter logos

### LINKED ISSUE #1816: 🖼️ Unreadable sponsorship logos (images are skewed)
Many of the Sponsorship logos are not readable. The logos are stretched and skewed. 

Semi-related to #1431 

Screenshot 2024-12-10 at 8 24 34 AM

bmispelon:
Here's how it looks like:

Screenshot 2024-12-12 at 14-31-13 Support Django Django

### DIFF
--- a/fundraising/tests/test_views.py
+++ b/fundraising/tests/test_views.py
@@ -1,24 +1,32 @@
 import json
 import shutil
 import tempfile
-from base64 import b64decode
 from datetime import date, datetime
+from io import BytesIO
 from operator import attrgetter
 from unittest.mock import patch
 
 import stripe
 from django.conf import settings
 from django.core import mail
-from django.core.files.base import ContentFile
+from django.core.files.images import ImageFile
 from django.template.defaultfilters import date as date_filter
 from django.test import TestCase
 from django.urls import reverse
 from django_hosts.resolvers import reverse as django_hosts_reverse
 from django_recaptcha.client import RecaptchaResponse
+from PIL import Image
 
 from members.models import CorporateMember, Invoice
 
 from ..models import DjangoHero, Donation
+
+
+def _make_image(name, width=1, height=1, color=(12, 75, 51)):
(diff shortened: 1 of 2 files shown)


bmispelon:
@adamzap Does that seem reasonable to you?

fundraising/tests/test_views.py
```diff
@@ -94,6 +95,27 @@ def test_corporate_member_with_logo(self):
...
+        Invoice.objects.create(amount=100, expiration_date=date.today(), member=member)
+        response = self.client.get(self.index_url)
+
+        self.assertContains(
+            response,
+            """<img
+                src="/m/cache/3a/ea/3aeaccc1f60ee53bf3317aae87c1c6a0.png"
+                srcset="/m/cache/3a/ea/3aeaccc1f60ee53bf3317aae87c1c6a0@2x.png 2x"
+                width="170"
+                height="17"
```
adamzap, line 112: Would there be value in deriving these values from the object? I could say the same about the paths above, but that would probably be a lot harder. I'm fine with this either way!

adamzap (approved):
👍 Looks great!

fundraising/tests/test_views.py
bmispelon: For the width/height I'd prefer them to stay hardcoded since that's what the test really is about: making sure the aspect ratio is preserved (the original image is 10x1, the thumbnail size is 170x170, so the expected size should be 170x17 to preserve the ratio).

For the path I guess we could be using `member.logo.url` but at that point the test doesn't feel so valuable because we're basically testing that `member.logo.url == member.logo.url`. It could get annoying if `sorl-thumbnail` changes its logic and generates new names, but I'd say we can fix the tests if and when that happens.

adamzap: Makes sense, thanks!

jefftriplett:
@bmispelon great work! Thank you.

PR #1828, opened 2024-12-16
Feat/docker dx

pbratkowski:
GitHub closed my previous PR (#1819) after I renamed a branch, so here is a new one.

Thanks to @tobiasmcnulty for pointing out a flaw in my previous implementation. This uses a build argument to determine when to remove packages, which allows removing them in the production build, but keeping them during local development which allows `tox` to build `psycopg` successfully in docker.

This also adds a container for the Trac database, which eliminates some errors upon startup, and makes tests pass in docker in most situations, and eliminates a lot of redundancy from the Docker settings file.


pbratkowski:
Tests currently still fail if `PARENT_HOST` is set to a non-default value, e.g. if running on a LAN rather than on `localhost`.

The issue with those is that they use hardcoded domains, e.g. https://github.com/django/djangoproject.com/blob/ff2fd01dba5de6b0d641e86df9bba64f4770e8f9/docs/tests.py#L377-L385

Replacing the hardcoded values with something like the following makes them pass successfully:

```python
    def test_sitemap_index(self):
        response = self.client.get(
            "/sitemap.xml", headers={"host": f"docs.{settings.PARENT_HOST}"}
        )
        self.assertContains(response, "<sitemap>", count=2)
        self.assertContains(
            response,
            f"<loc>http://docs.{settings.PARENT_HOST}/sitemap-en.xml</loc>",
        )
```
If anyone has alternative solutions, let me know.

Dockerfile
```diff
@@ -34,11 +37,13 @@ RUN apt-get update \
...
     && python3 -m pip install --no-cache-dir -r ${REQ_FILE} \
-    && apt-get purge --assume-yes --auto-remove \
+    && if [ "${KEEP_DEPENDENCIES}" != "True" ]; then \
+        apt-get purge --assume-yes --auto-remove \
         gcc \
         libc6-dev \
         libpq-dev \
-        zlib1g-dev \
+        zlib1g-dev; \
+        fi \
```
pauloxnet, line 46:
```suggestion
    fi \
```
Maybe unintended, this not seems the name of a package?

pauloxnet (commented):
What's the value in keeping the dependencies in Docker?
I see mixed changes in this PR (e.g. dockers, settings.PARENT_HOST, ...)
Can you provide seprataed PRs ?

Dockerfile
pbratkowski: This is indeed part of the `if; then; fi` syntax, and not a package. Dedenting it makes sense.

pbratkowski:
@pauloxnet how granular would you like the changes to be? I have removed the commit that fixes tests when `settings.PARENT_HOST` is set.

pbratkowski:
I converted this back to a Draft for now, as we are working out smaller scope PRs.

adamzap:
@pbratkowski We're going to close this for now since many improvements have been made to the Docker setup since it was opened, and a rebase would probably be difficult. However, if you see other possible improvements, please let us know via an issue or PR. Thanks!

PR #1829, opened 2024-12-16
Removed erroneous template code from corporate member list

tobiasmcnulty:
The corporate member list has some stray trailing template tags left from a previous change.

Screenshot 2024-12-16 at 10 52 39 AM


bmispelon (approved):
That's probably my bad, I must have missed that during review.

Thanks for catching that and submitting a PR! 🎸

Issue #1817, opened 2024-12-10
Instructions for running tox with docker don't work


bmispelon:
As mentionned by Pat on Discord (not sure what their github username is), our [current instructions](https://github.com/django/djangoproject.com/blob/75f3fdee0bcc49b9502a2dd5d1d0221978c79091/README.rst?plain=1#L383) for running tox with docker don't seem to work:

<details>
<summary>Full log</summary>
```
$ docker compose exec web tox
py312-tests: install_deps> python -I -m pip install -r /usr/src/app/requirements/tests.txt
Collecting django-push@ git+https://github.com/brutasse/django-push.git@22fda99641cfbd2f3075a723d92652a8e38220a5 (from -r /usr/src/app/requirements/common.txt (line 6))
  Cloning https://github.com/brutasse/django-push.git (to revision 22fda99641cfbd2f3075a723d92652a8e38220a5) to /tmp/pip-install-xbk4s6l8/django-push_869cd076d70d473da4c86b60e8cecf7d
  Resolved https://github.com/brutasse/django-push.git to commit 22fda99641cfbd2f3075a723d92652a8e38220a5
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Collecting Babel==2.16.0 (from -r /usr/src/app/requirements/common.txt (line 1))
  Downloading babel-2.16.0-py3-none-any.whl.metadata (1.5 kB)
Collecting django-contact-form==5.1.1 (from -r /usr/src/app/requirements/common.txt (line 2))
  Downloading django_contact_form-5.1.1-py3-none-any.whl.metadata (2.7 kB)
Collecting django-countries==7.6.1 (from -r /usr/src/app/requirements/common.txt (line 3))
  Downloading django_countries-7.6.1-py3-none-any.whl.metadata (31 kB)
Collecting django-hosts==5.1 (from -r /usr/src/app/requirements/common.txt (line 4))
  Downloading django_hosts-5.1-py3-none-any.whl.metadata (4.7 kB)
Collecting django-money==3.5.3 (from -r /usr/src/app/requirements/common.txt (line 5))
  Downloading django_money-3.5.3-py3-none-any.whl.metadata (18 kB)
Collecting django-read-only==1.18.0 (from -r /usr/src/app/requirements/common.txt (line 7))
  Downloading django_read_only-1.18.0-py3-none-any.whl.metadata (7.1 kB)
[...]

PR #1833, opened 2024-12-18
Skipped test building of docker image on push

bmispelon:
The image will actually get built (and published) anyway.

### DIFF
--- a/.github/workflows/docker-test-build.yml
+++ b/.github/workflows/docker-test-build.yml
@@ -2,9 +2,6 @@ name: Docker test build
 
 on:
     pull_request:
-    push:
-        branches:
-            - main
 
 concurrency:
     group: ${{ github.workflow }}-${{ github.ref }}


bmispelon:
@tobiasmcnulty does that make sense? Or am I missing a usecase for testing the build before actually publishing it?

tobiasmcnulty:
@bmispelon Good catch. I agree it would be nice to remove the duplicate build of the Docker image with the production/deployed requirements file.

This change would no longer test the Docker build with test requirements after a PR is merged (relevant only if the offending PR was not updated to latest `main` before merging). That may be an edge case not worth worrying about.

In the future, maybe the two files could be consolidated to eliminate the duplicate build, but still always test both builds on pushes to `main`?

bmispelon:
> This change would no longer test the Docker build with test requirements after a PR is merged (relevant only if the offending PR was not updated to latest `main` before merging). That may be an edge case not worth worrying about.

Ah right, good point. That does seem to be quite the edge case, plus as it turns out our instructions for testing in docker haven't been working for a while (#1817 )

> In the future, maybe the two files could be consolidated to eliminate the duplicate build, but still always test both builds on pushes to `main`?

That's an interesting idea: test the build with the `tests.txt` requirements then build+publish the one with `prods.txt`? I could file a new ticket for that.

tobiasmcnulty:
> > This change would no longer test the Docker build with test requirements after a PR is merged (relevant only if the offending PR was not updated to latest `main` before merging). That may be an edge case not worth worrying about.
> 
> Ah right, good point. That does seem to be quite the edge case, plus as it turns out our instructions for testing in docker haven't been working for a while (#1817 )

Interesting, maybe we could test this flow in CI, too. I vaguely recall thinking about that at DjangoCon and then running out of time and/or steam. 😅
 
> > In the future, maybe the two files could be consolidated to eliminate the duplicate build, but still always test both builds on pushes to `main`?
> 
> That's an interesting idea: test the build with the `tests.txt` requirements then build+publish the one with `prods.txt`? I could file a new ticket for that.

That should work. I think it could still be a matrix build so they both happen at the same time.

Another option would be to keep the workflow files separate (maybe better if we want CI to run tests in Docker too?), but add a variable to push only if the branch is `main`.

Anyways, I think this is good, thanks! 👍🏻

marksweb:
@bmispelon This looks good to merge?

Issue #1827, opened 2024-12-15
Proposal for further front-end cleanups


adamzap:
I'm curious if the maintainers would accept pull requests for the following front-end changes. Some of these ideas are more opinionated or need more guidance than my previous PRs, so I wanted to get feedback first.

## Remove `bower`

As its website says, [`bower`](https://bower.io/) is maintained but not recommended for use by its developers. [Our `bower.json`](https://github.com/django/djangoproject.com/blob/ff2fd01/bower.json) file now only includes two dependencies: `jquery` and `jquery-flot`.

Instead of using `bower`, could we instead put `jquery.min.js` and `jquery-flot.min.js` directly in [`djangoproject/static/js/lib/`](https://github.com/django/djangoproject.com/tree/ff2fd01/djangoproject/static/js/lib)? This would include removing the `jquery` and `jquery-flot` directories.

## Remove `require.js`

[`require.js`](https://requirejs.org/) had a small security release in July 2024, but the release before that was in 2018, and [the project is not under active development](https://github.com/requirejs/requirejs/issues/1854#issuecomment-2049998599). I'm still trying to understand the full influence that `require.js` is having on this project, but removing it would require at least the following changes in addition to removing the library itself:

1. Remove [the setup for `require.js` in the `base.html` template](https://github.com/django/djangoproject.com/blob/ff2fd01/djangoproject/templates/base.html#L137-L167)
2. Remove the `define` calls in our JavaScript modules ([example](https://github.com/django/djangoproject.com/blob/ff2fd01dba5de6b0d641e86df9bba64f4770e8f9/djangoproject/static/js/mod/clippify.js#L1))
3. Remove or repurpose [`main.js`](https://github.com/django/djangoproject.com/blob/ff2fd01/djangoproject/static/js/main.js)
    - **Option 1**: Remove the file and add `<script>` tags to the templates where needed. Adding a new template block for JavaScript includes would also be a good idea. *This approach is simple but is riskier and requires more work and testing, but it's the one I would choose.*
    - **Option 2:** Keep the file for its class-based logic and dynamically add `<script>` tags to the DOM based on it. *This approach requires less changes and may be a good way just to remove `require.js` without broader changes.*

Which approach would be preferred here, or is there a better way than what I presented above?

## Remove Boilerplate in Many JavaScript Modules

[...]

PR #1835, opened 2024-12-18
Refactored `doc-switcher.js`

adamzap:
- Simplified code
- Stopped using jQuery
- Moved refactored code to `app.js`

This is the first in a series of patches that will move JavaScript code out of `require.js` modules and into a single file while also refactoring.

To test, ensure that the functionality of this component on documentation pages still works as expected:

image

If the "Language" or "Documentation version" elements are clicked, it should toggle them remaining open when not hovered over. They will get a green border when activated. 

Part of #1827


djangoproject/static/js/app.js
adamzap, line 1: Should this file be renamed? I wasn't sure what to call it.

djangoproject/static/js/app.js
```diff
@@ -0,0 +1,5 @@
+document.querySelectorAll('.doc-switcher li.current').forEach(function (el) {
```
adamzap, line 1: This will run on all pages, but in my testing, it seems to be about twice as fast as the old `hasClass` method from `main.js`. So I think we're safe here.

djangoproject/static/js/mod/doc-switcher.js
```diff
@@ -1,25 +0,0 @@
...
-    };
-
-    DocSwitcher.prototype = {
-        init: function(){
-            var self = this;
-            $(document).ready(function () {
-                // Make version switcher clickable for touch devices
-
-                self.switchers.find('li.current').on('click', function () {
-                    $(this).closest("ul").toggleClass('open');
```
adamzap, line 17: I used the parent element in the new code instead of trying to find the "closest" `ul` element. The `li` elements we are looking for should always be the direct children of the `ul` element that needs the class.

adamzap:
I suppose we could try to replace this JavaScript with a `details` element, but I might rather circle back to the effort.

djangoproject/static/js/app.js
```diff
@@ -0,0 +1,7 @@
+// Toggle persistent display of documentation version and language options if
+// the element is clicked
```
adamzap, line 2: Should we keep a comment like this? It may be valuable to have a comment for each bit of behavior that goes into this file. I don't feel strongly about it.

bmispelon:
I've only skimmed the PR, but it seems pretty good so far. Thanks for working on this! 🎸 

Could we start by setting up a formatter/linter/ci for js before we start the refactors?

djangoproject/static/js/app.js
pauloxnet: Comments can be useful for developers but increases the size of the website and page lightness is a priority if we want to be accessible also for people with old devices or slow connection.

What do you think about having a minified version of the JavaScript file with all comments stripped?

adamzap: I wonder if we should start getting refactors in to see how much JavaScript we're going to end up with before making the decision to minify it. I'm confident that it's going to be way less than what's currently on production, and that's not minified.

Just to be clear, are you saying that we should set up `django-compressor` for `app.js`?

Also, I'm fine just not having the comment too.

adamzap:
@bmispelon Sure. How about I open a PR that does the following:

1. Enable JavaScript for `prettier` in `.pre-commit-config.yaml`
2. Add a configuration file for `prettier` for JavaScript (I'll start with my preferences, and we can refine it)
3. Run `prettier` to reformat code in `main.js` and `static/js/mod/*.js`
4. Enable `prettier` in linting mode in CI

That way, once it's merged, linting will be enforced and enabled in `pre-commit`.

adamzap:
This has been rebased to get 6dc8f41 under it. Let me know about renaming `app.js` and whether or not we want a comment before the new code.

djangoproject/static/js/app.js
bmispelon: I don't have strong feelings, but `app.js` is a bit meaningless to me, yeah. How about `djangoproject.js`?

adamzap: Great! I just pushed the change in f3ab40d9.

djangoproject/static/js/app.js
bmispelon: It's an interesting consideration about size of js and slow connections.

My opinion on this is that Adam has removed so many lines of js that a few comments are allowed 😁 

`django-compressor`would be the solution to this for me, but that's out of scope for this PR I would say.

bmispelon (approved):
This is looking good to me, I'm going to do a final test on my local version before I merge. Thanks again 👍🏻

adamzap:
@bmispelon Thanks! I pushed one tiny edit to the comment!

djangoproject/static/js/app.js
pauloxnet: Yes, an amazing job has been done here. We can think about minified JS in another PR

pauloxnet (approved):
LGTM

PR #1836, opened 2024-12-19
test github action for collectstatic

### LINKED ISSUE #1831: Add a test (or a github action) to catch static files issues early
I recently had to revert a commit because deploying it on the production server failed when running `collectstatic`, even though all the tests and github actions passed.

The issue is that the production server uses `ManifestStaticFilesStorage` for its static files, which adds a post-processing step the the `collectstatic` command and that step will crash with a `ValueError` if a missing static file is referenced (in my case, a minified javascript file had a comment referencing a `.map` file that had been deleted).

This type of issue should be caught early so we can fix it during the PR and avoid a revert.

ontowhee:
Fix #1831 

WIP. This is probably not the right approach. It adds the collectstatic command to tox, although tox is running under development environment and we are trying to test production.

### To Do
- [ ] Remove jquery.min.map file, then run collectstatic to observe the ValueError.
- [ ] Try adding a new step in github actions to `.github/workflows/tests.yml`, something along the lines of:
```
            - name: Run collectstatic for production
              run: |
                pip install -r djangoproject/requirements/prod.txt
                python manage.py collectstatic --settings=djangoproject.settings.prod
```
What environment variables need to be set?

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -36,42 +36,42 @@ jobs:
               uses: actions/setup-python@v5
               with:
                   python-version: "${{ matrix.python-version }}"
-            - name: Install dependencies
-              run: |
-                  python -m pip install --upgrade pip setuptools coveralls "tox<5" "tox-gh-actions<4"
-            - name: Set up databases
-              run: |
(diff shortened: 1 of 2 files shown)


adamzap:
The issue was addressed in #2742 by taking a different approach. Thanks for your effort here!

PR #1837, opened 2024-12-19
Test collectstatic with error

ontowhee:
Fix #1836 

This builds on [432f859](https://github.com/django/djangoproject.com/commit/432f85945e8138b9be7dce7db1ca71421fe231cb) to demonstrate the ValueError described in #1831. It removes `jquery.min.map`, and adds a new step in github actions to run collectstatic with production settings.


adamzap:
The issue was addressed in #2742 by taking a different approach. Thanks for your effort here!

PR #1838, opened 2024-12-20
Run `prettier` on JavaScript via `pre-commit`

adamzap:
This patch removes the global `djangoproject/static/*` exclude rule in `.pre-commit-config.yaml` to enable `prettier` to run on this project's JavaScript files. The `djangoproject/static/js/lib/` directory is now excluded in the `prettier` hook because there is no value in reformatting a minified JavaScript file.

I also added a `.prettierrc` file to the root of the project to control JavaScript formatting.

@bmispelon Is this what you had in mind as a requirement to get #1835 in?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,4 +1,4 @@
-exclude: '(^djangoproject\/static\/.*$)'
+exclude: '^djangoproject\/static\/(css\/|fonts\/|img\/|robots).*$'
 default_language_version:
     python: python3
 repos:
@@ -19,6 +19,7 @@ repos:
           - id: debug-statements
           - id: detect-private-key
           - id: end-of-file-fixer
+            exclude: '(^djangoproject\/static\/js\/lib\/.*$)'
             exclude_types: [json, sql]
           - id: file-contents-sorter
             files: ^(requirements/\w*.txt)$
@@ -46,6 +47,7 @@ repos:
       hooks:
           - id: prettier
             exclude_types: [html, json, scss]
+            exclude: '(^djangoproject\/static\/js\/lib\/.*$)'
     - repo: https://github.com/pycqa/isort
       rev: "5.13.2"
       hooks:
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,11 @@
+{
+    "overrides": [
+        {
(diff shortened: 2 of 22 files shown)


adamzap:
I added a commit to show that `pre-commit` will reformat JavaScript.

.pre-commit-config.yaml
```diff
@@ -1,4 +1,4 @@
-exclude: '(^djangoproject\/static\/.*$)'
+exclude: '^djangoproject\/static\/(css\/|fonts\/|img\/|robots).*$'
```
adamzap, line 1: This makes `pre-commit` only run on the `js` directory in `static`. In a series of future PRs I'd like to remove this `exclude` line and fix all of the warnings.

.pre-commit-config.yaml
```diff
@@ -19,6 +19,7 @@ repos:
           - id: debug-statements
           - id: detect-private-key
           - id: end-of-file-fixer
+            exclude: '(^djangoproject\/static\/js\/lib\/.*$)'
```
adamzap, line 22: We don't need to modify minified JavaScript files to add a newline at the end.

.pre-commit-config.yaml
```diff
@@ -46,6 +47,7 @@ repos:
       hooks:
           - id: prettier
             exclude_types: [html, json, scss]
+            exclude: '(^djangoproject\/static\/js\/lib\/.*$)'
```
adamzap, line 50: We don't want to run the formatter on minified JavaScript files.

.prettierrc
```diff
@@ -0,0 +1,11 @@
+{
+    "overrides": [
+        {
+            "files": "*.js",
+            "options": {
+                "tabWidth": 2,
+                "singleQuote": true
+            }
```
adamzap, line 8: These options are my personal preference. Let me know if this should be modified.

https://prettier.io/docs/en/options

adamzap:
You can see in 0091a87 that `pre-commit`/`prettier` has reformatted all relevant JavaScript according to the new rules I added.

pauloxnet (approved):
LGTM

bmispelon (approved):
This is great, I'm very happy about having a an automatically enforced standard now 🎸

PR #1839, opened 2024-12-20
Removed executable bit on image assets

adamzap:
These files need to be read, not executed.


bmispelon (approved):
Nicely spotted. I'm assuming you used a command to change the file permissions (and maybe find the files too), can you share it?

adamzap:
I found this through experimenting with running `pre-commit` on `djangoproject/static/`. I believe [this rule](https://github.com/django/djangoproject.com/blob/6dc8f418e6b652863ccb16a3e2e01795e8a89d9f/.pre-commit-config.yaml#L12) failed on these.

pauloxnet (approved):
LGTM

PR #1853, opened 2024-12-28
Used local `fontawesome-webfont.woff` file

adamzap:
For the last ten years (since 7abaceb1), the site has fetched its Font Awesome font asset from a remote CDN. It seems to me that since 7d21f20a, we would rather host our static assets ourselves. This patch makes that change for the current version of Font Awesome.

`base.html` was already referring to the local version, so there was an inconsistency here. Merging #1854 will remove the possibility of inconsistency here.

### DIFF
--- a/djangoproject/scss/font-awesome/_variables.scss
+++ b/djangoproject/scss/font-awesome/_variables.scss
@@ -1,8 +1,7 @@
 // Variables
 // --------------------------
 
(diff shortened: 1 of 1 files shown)


pauloxnet (approved):

bmispelon:
Nice work! 🚀

PR #1854, opened 2024-12-28
Remove `link` tags for font assets

adamzap:
Since we're using `src: url('...')` in CSS, these `link` tags shouldn't be needed. A look in DevTools proves that the font asset requests are coming from the CSS file.

image

image

image

Related to #1853


pauloxnet (approved):

bmispelon:
Nice catch, thanks! 🎸

PR #1861, opened 2025-01-04
Remove Debian packages from Dockerfile

pauloxnet:
This contains the following changes:
- removing `zlib1g` from Dockerfile because it's already part of the base image
- used `pg_isready` to check the database readiness, removing `netcat-openbsd` from Dockerfile
- uninstalling `g++` after installing requirements

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,10 +15,8 @@ RUN apt-get update \
         git \
         libpq5 \
         make \
-        netcat-openbsd \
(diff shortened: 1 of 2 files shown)


marksweb (approved):
Nice bit of tidying up 👍🏼 

Seems to work for me so LGTM.

bmispelon:
Nice work, thanks 🎸

PR #1879, opened 2025-01-18
Updated "Fira Mono" font assets

adamzap:
- Upgraded to the latest version
- Used `woff2` format instead of `woff`

The files were fetched from GitHub: https://github.com/mozilla/Fira/tree/master/woff2.

There's not much space savings here, but I think it's still worth upgrading.

This change can be verified by checking that the new files are being loaded in the DevTools network panel:

image

Another way verify this change is observing that certain glyphs are updated to match the hosted version [here](https://fonts.google.com/specimen/Fira+Mono). Note the "r":

Hosted (new):

image

Old:

image

New:

image

Refs #1815


pauloxnet (approved):
LGTM

PR #1889, opened 2025-01-21
Improve `depends_on: db` in docker-compose.yml

trottomv:
Add the `condition: service_healthy` to improve the `db` service healthcheck support

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -22,7 +22,8 @@ services:
             - SQL_HOST=db
             - SQL_PORT=5432
         depends_on:
-            - db
+            db:
+                condition: service_healthy
     db:
         image: postgres:14-alpine
         environment:


pauloxnet (approved):
Thanks

ulgens:
Does this work without actually testing the connection? As far as I remember, docker saying a service healthy and Django being able to connect that service is not the same thing.

trottomv:
> Does the container/docker image define what it means to be ready to accept connections?

The docker compose directive `healthcheck` and its test command define when the service is health, in the specific case of PostgreSQL, the Docker image natively includes the command `pg_ready` to check if the service is ready. This has the advantage of not requiring additional libraries or custom scripts to be added to the web service image. With the healthcheck directive, at that point, Docker Compose tells the web service, "Okay, db is ready, you can start."

ulgens:
@trottomv Got it, thank you so much.

PR #1891, opened 2025-01-21
Add tblib to test dependencies

ulgens:
So we can see test failures using --parallel with tests

### DIFF
--- a/requirements/tests.txt
+++ b/requirements/tests.txt
@@ -1,4 +1,5 @@
 -r dev.txt
 coverage==7.6.10
 requests-mock==1.12.1
+tblib>=3.0.0
 tox<5.0.0


pauloxnet (approved):
I +1 to add it but I would set a version as for the other dependencies

ulgens:
@pauloxnet That makes sense, but I'm not sure what to set as a version. Anything that support Python 3.12 will work just fine, which is the default behaviour as not having a version. Do you have any recommendations?

pauloxnet:
> @pauloxnet That makes sense, but I'm not sure what to set as a version. Anything that support Python 3.12 will work just fine, which is the default behaviour as not having a version. Do you have any recommendations?

You can set the minimum version you know will work correctly with Python 3.12 and Django 5.0

pauloxnet (approved):

PR #1894, opened 2025-01-22
Use "run --rm" instead of exec for tests in README

ulgens:
This is something pretty small but I find `run --rm` more robust than `exec`, because:

* exec requires a running container, "run --rm" starts a new one and deletes when the job is done
* exec requires the related container to stay healthy, "run --rm" run things on a separate container

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -365,8 +365,8 @@ Running Locally with Docker
 
 4. Run the tests::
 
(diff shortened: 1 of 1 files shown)


pauloxnet (approved):

bmispelon:
Good idea, thanks! 🎸

PR #1900, opened 2025-01-24
Upgrade Django version to 5.2

ulgens:
https://docs.djangoproject.com/en/dev/releases/5.2/

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -36,6 +36,7 @@ repos:
       rev: "1.22.2"
       hooks:
           - id: django-upgrade
+            # TODO: Update to 5.2 when django-upgrade supports it
             args: [--target-version, "5.1"]
     - repo: https://github.com/psf/black
       rev: 24.10.0
--- a/djangoproject/settings/common.py
+++ b/djangoproject/settings/common.py
@@ -172,7 +172,6 @@ SERVER_EMAIL = "root@djangoproject.com"
 SESSION_COOKIE_HTTPONLY = True
 
 SILENCED_SYSTEM_CHECKS = [
-    "fields.W342",  # tracdb has ForeignKey(unique=True) in lieu of multi-col PKs
     "security.W008",  # SSL redirect is handled by nginx
     "security.W009",  # SECRET_KEY is setup through Ansible secrets
 ]
--- a/requirements/common.txt
+++ b/requirements/common.txt
@@ -7,7 +7,7 @@ django-push @ git+https://github.com/brutasse/django-push.git@22fda99641cfbd2f30
 django-read-only==1.18.0
 django-recaptcha==4.0.0
 django-registration-redux==2.13
-Django==5.1.5
+Django==5.2a1
 docutils==0.21.2
 feedparser==6.0.11
 Jinja2==3.1.5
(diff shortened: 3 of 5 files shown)


pauloxnet:
> Considering that this upgrade doesn't seem to require any code change, I would like to deploy it in production this week.

Do you think it's safe to deploy in production the Django 5.2 alpha version?

I thought we open this PR to check in advance for issue with 5.2 version and then upgrade after the stable release.

I'm not strongly against the idea of merging this PR now, but I'm interested on other point of view.

pauloxnet:
> I've pushed a commit that makes use of the new support for composite primary keys and allows us to delete a bunch of hacks, but it's not strictly required for the upgrade.

BTW your commit is great ... starting using this new feature in the Django website is very good, also as an example.

adamzap:
> As with all alpha and beta packages, this is not for production use.

From https://www.djangoproject.com/weblog/2025/jan/16/django-52-alpha-1-released/

😉 

Do you mean putting it on production to test, or upgrading production to Django 5.2 and keeping it there? I'm also excited about the Trac hacks being removed, but security or other issues could still be lurking as the community tries the release out.

Maybe I'm too conservative on upgrades!

ulgens:
I don't remember when was the last time I had issues with a pre-release version of Django. If all goes bad, I don't expect a worse scenario than a rollback.

Also, I believe it's important to test new Django versions with projects under the Django organization. I understand why it sounds a bit risky, but isn't it weird not to take that risk ourselves when we expect it from other users?

adamzap:
@ulgens I think the idea of testing a pre-release version is to run it on your dev and staging servers as well as running your test suite against it.

pauloxnet:
> ... I believe it's important to test new Django versions with projects under the Django organization.

It's ok to test as much as possible Django alphas version in project, and we ask Django developer to do the same locally or in pipelines.

> ... isn't it weird not to take that risk ourselves when we expect it from other users?

Actually, we don't ask users to take the risk to use alpha version in production, in fact we warn them against this risk as @adamzap already pointed out.

That said, we can decide to do that if we have strong motivation.

ulgens:
@adamzap @pauloxnet I see. I do test Django versions in prod with active commercial products regularly, but I see that that is not common or expected 🙂 Do we have anything resembles that a staging environment, a remote setup that is not production?

bmispelon (approved):
Thanks for all the feedback everyone.

As an experiment, I deployed this PR in production for the last two hours and I'm happy to report that no errors were logged during that time 🎉 

I've now un-deployed the change and we're back to running 5.1 again, but this makes me even more confident that we can deploy 5.2 once it's out.

pauloxnet:
> As an experiment, I deployed this PR in production for the last two hours and I'm happy to report that no errors were logged during that time 🎉
> 
> I've now un-deployed the change and we're back to running 5.1 again, but this makes me even more confident that we can deploy 5.2 once it's out.

Thanks for the test and for the feedback. :+1: 

I propose to create a stage environment for the future https://github.com/django/djangoproject.com/issues/1922

bmispelon:
This was done in 4257af94f8a40a07a1d3038fcc1e99a0cca559f5 (I should have used this PR then, that's my bad)

PR #1902, opened 2025-01-24
Clean pending migrations check

ulgens:
The pending migrations check was running twice. I removed it from the makefile and added a reference to the origin of the code block (thinking it would be nice to give credit 🎈)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ migrations-check:
 run:
 	python -m manage runserver 0.0.0.0:8000
 
-test: migrations-check
+test:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -25,13 +25,10 @@ isort:
 isort-check:
 	python -m isort --check $(APP_LIST)
```
pauloxnet, line 27:
```suggestion

migrations-check:
	python -m manage makemigrations --check --dry-run
```
I would leave this method in the Makefile, so that you can run it also without running the test.

pauloxnet (approved):

bmispelon:
Nice catch, thanks! 🎸

PR #1904, opened 2025-01-27
Clean cache settings

ulgens:
Small cleaning on cache settings

I was going to replace `DummyCache` with `LocMemCache` but apparently, that change breaks the tests with `assertNumQueries`, passing it for now.


ulgens:
Superseded by https://github.com/django/djangoproject.com/pull/1911

PR #1906, opened 2025-01-27
Tidy up Docker settings

ulgens:
Several cleaning tasks on the docker settings file. 

Depends on
- https://github.com/django/djangoproject.com/pull/1904

Relates to
- https://github.com/django/djangoproject.com/pull/1828

### DIFF
--- a/djangoproject/settings/dev.py
+++ b/djangoproject/settings/dev.py
@@ -58,6 +58,7 @@ if DEBUG:
         "djangoproject.middleware.CORSMiddleware",
     )
 
-SILENCED_SYSTEM_CHECKS = SILENCED_SYSTEM_CHECKS + [
+SILENCED_SYSTEM_CHECKS += [
+    # Default test keys for development.
     "django_recaptcha.recaptcha_test_key_error"
 ]
--- a/djangoproject/settings/dev.py
+++ b/djangoproject/settings/dev.py
@@ -47,21 +47,16 @@ PUSH_SSL_CALLBACK = False
 # Enable optional components
 
 if DEBUG:
-    try:
-        import debug_toolbar  # NOQA
-    except ImportError:
-        pass
-    else:
-        INSTALLED_APPS.append("debug_toolbar")
-        INTERNAL_IPS = ["127.0.0.1"]
-        MIDDLEWARE.insert(
-            MIDDLEWARE.index("django.middleware.common.CommonMiddleware") + 1,
-            "debug_toolbar.middleware.DebugToolbarMiddleware",
-        )
-        MIDDLEWARE.insert(
-            MIDDLEWARE.index("debug_toolbar.middleware.DebugToolbarMiddleware") + 1,
-            "djangoproject.middleware.CORSMiddleware",
-        )
+    INSTALLED_APPS.append("debug_toolbar")
+    INTERNAL_IPS = ["127.0.0.1"]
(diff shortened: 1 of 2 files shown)


djangoproject/settings/docker.py
```diff
@@ -1,4 +1,4 @@
-from .common import *  # noqa
+from .dev import *  # noqa
```
pauloxnet, line 1:
```suggestion
from .dev import *  # noqa: F403
```
Since you're editing this line, we can add the exact flake8 rule to exclude.

djangoproject/settings/dev.py
```diff
@@ -47,22 +47,20 @@
...
-            MIDDLEWARE.index("django.middleware.common.CommonMiddleware") + 1,
-            "debug_toolbar.middleware.DebugToolbarMiddleware",
-        )
-        MIDDLEWARE.insert(
-            MIDDLEWARE.index("debug_toolbar.middleware.DebugToolbarMiddleware") + 1,
-            "djangoproject.middleware.CORSMiddleware",
-        )
-
-SILENCED_SYSTEM_CHECKS = SILENCED_SYSTEM_CHECKS + [
+    import debug_toolbar  # NOQA
```
pauloxnet, line 50:
```suggestion
    import debug_toolbar  # NOQA: E402
```
Since you're editing this line, we can add the exact flake8 rule to exclude.

djangoproject/settings/docker.py
ulgens: Done.

djangoproject/settings/dev.py
ulgens: I removed the import, it wasn't required without the try-except block

ulgens:
@pauloxnet If you have the time, I'd use your review on https://github.com/django/djangoproject.com/pull/1904 too 🌷

djangoproject/settings/dev.py
pauloxnet: I would restore the try except with debug toolbar import because it lets you use dev settings also if you have installed other requirements files (eg: common, prod, ...) and it's very convenient when you're working locally

ulgens: @pauloxnet I was going for a single settings file at the end. The pattern used in this repo is a bit outdated IMO, and overly complicates things. The same goes for the requirements files; these things should be dependency groups under a single pyproject.toml file. 

Considering the actual goal, I'd like to clean and simplify as much as possible, and soon the need for these conditionals will be gone. If it's going to create issues in the short term, please let me know.

ulgens: That makes sense. Let me roll back that change.

I was aiming to create minimal sized PRs, but I see why it doesn't look same from different angles. What I had in mind was to
* Clean settings files
* Help https://github.com/django/djangoproject.com/pull/1828 and get it merged
* Fix local testing under docker setup
* Clean duplicated checks that are run by both pre-commit and tox
* Remove tox
* Test with uv and ruff
* Optimize CI

Considering each step requires multiple PRs and discussions, I didn't really think to handle all of them at one go; but I'll try to minimize further.

pauloxnet: I agree with most all your goals, but we need to try to reach all that goals as clean as possible and together.

I would suggest to you to search for already opened issue that cover your goals, and if you don't find them, open a new issue for every point, put all your idea on it so that we can discuss each point together.

PR #1911, opened 2025-01-28
Clean duplicate settings from settings/docker.py

ulgens:
`settings/docker.py` file consists of many duplicate settings already covered by `settings/dev.py`. By changing the base, we no longer need to define those and file size can be minimized.

The goal is to clean redundant code and help to minimize https://github.com/django/djangoproject.com/pull/1828

The PR doesn't include any logic or functional change.

Supersedes
- https://github.com/django/djangoproject.com/pull/1904
- https://github.com/django/djangoproject.com/pull/1906

### DIFF
--- a/djangoproject/settings/dev.py
+++ b/djangoproject/settings/dev.py
@@ -64,5 +64,6 @@ if DEBUG:
         )
 
 SILENCED_SYSTEM_CHECKS = SILENCED_SYSTEM_CHECKS + [
+    # Default test keys for development.
     "django_recaptcha.recaptcha_test_key_error"
 ]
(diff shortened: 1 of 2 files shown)


ulgens:
@bmispelon Sorry for the individual ping but who would be the right person to review this?

alexgmin (approved):

alexgmin:
We could probably remove 
`LOCALE_MIDDLEWARE_EXCLUDED_HOSTS = ["docs.djangoproject.localhost"]` too

Not sure about `ALLOWED_HOSTS,` and `PARENT_HOST`. 

But this is a small improvement at least.

Issue #1912, opened 2025-01-28
"Request req_...... Received unknown parameter: source" error when updating monthly donation


tobiasmcnulty:
Our credit card expired and I went to update the recurring monthly donation after receiving the following email:

> Hi ____,

> This is to let you know that payment for your recurring donation to the Django Software Foundation has failed.
> This might be because your payment card has expired or you don't have enough funds.

Clicking the link and updating the credit card number appears successful according to the Stripe dialog, but this page error appears and I don't believe the card was actually updated because I've received the failure email a couple times now.

<img width="936" alt="Image" src="https://github.com/user-attachments/assets/0f5892b8-7a7b-4b9a-83ca-8ed4e55993fc" />

bmispelon:
This must have broken after I updated `stripe` in 10ca17abf72add4ecc71e4fd836e7c137283d939. I'll take a look, thanks for reporting the issue!

bmispelon:
I'm being told that a new error now appears:
```
Request [redacted]: No such source: 'tok_[redacted]'
```

adamzap:
I haven't looked too deeply at this, but I think you need to create the payment method first (attached to the customer) then pass its ID instead of the token [here](https://github.com/django/djangoproject.com/commit/d2f080ef7de2225686a00b0e76d20693faf28532#diff-8d4f767d61bc750e1b90e8cd319ebfe9231520e44f2f135ea3a62c529790f532R163).

PR #1915, opened 2025-01-29
Fixed handling of token in card update view

bmispelon:
Refs #1912


bmispelon:
A reviewer wishing to stay anonymous said:

> I think it may work looking at the token api

Considering how hard it is to test this locally, I'm going to try and deploy it and see if that fixes things 🤞🏻

PR #1917, opened 2025-01-31
Removed loading of deleted list-feature.js from base.html

bmispelon:
list-feature.js was deleted in af5cd63745f0df0159d06b528956c895b9312902

### DIFF
--- a/djangoproject/templates/base.html
+++ b/djangoproject/templates/base.html
@@ -137,7 +137,6 @@
           "jquery.flot": extless("{% static 'js/lib/jquery.flot.min.js' %}"),
           "mod/floating-warning": extless("{% static 'js/mod/floating-warning.js' %}"),
           "mod/list-collapsing": extless("{% static 'js/mod/list-collapsing.js' %}"),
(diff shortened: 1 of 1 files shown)


bmispelon:
@adamzap I believe this was the cause of the crash when deploying this (had nothing to do with out deploy script). I should have spotted that during review, sorry.

pauloxnet (approved):
LGTM

bmispelon:
> I thought we had a test that makes sure `collectstatic` works in a production-like environment?

We have an issue to implement such a test, but it hasn't been done yet: #1831 (there's a PR for that ticket, but it's not in a mergeable state right now and I haven't had time to give feedback on it).

But that wouldn't have caught this issue anyway, since `collectstatic` would have worked. It might work if we run the test suite with a `ManifestStorage` for static files (assuming there's at least one test that ends up rendering the `base.html`).

In any case, that's a failure on the part of our test suite, not on yours. Plus all it did was cost us a few minutes of down time, and then we ended up making our deploy script more robust. Not a bad outcome at all 😁

PR #1918, opened 2025-01-31
Removed unused arguments in messages JavaScript

adamzap:
These event arguments must have been left from a previous approach. They are not used, so they can be removed.

This was missed in 20cec455.


bmispelon (approved):
Nicely spotted 👍🏻

pauloxnet (approved):

PR #1919, opened 2025-01-31
1347: improving 404 page

srijanid:
Made a change in 404 page.
Uses difflib.get_close_matches to suggest correct URLs.
Uses proper i18n translations and best practices.
Makes it easier to search for documentation.

### DIFF
--- a/djangoproject/templates/404.html
+++ b/djangoproject/templates/404.html
@@ -1,23 +1,51 @@
 {% extends 'base_error.html' %}
 {% load i18n %}
+{% load custom_tags %}
 
 {% block title %}{% translate "Page not found" %}{% endblock %}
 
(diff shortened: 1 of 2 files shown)


djangoproject/templates/404.html
```diff
@@ -1,23 +1,51 @@
 {% extends 'base_error.html' %}
 {% load i18n %}
+{% load custom_tags %}
```
marksweb, line 3: There's no need to have a separate line to load this, but equally it seems that it doesn't exist - at least I can't find a `custom_tags.py` file.

adamzap:
I'm going to close this due to inactivity. @srijanid please reopen it if you want to address the comment.

PR #1921, opened 2025-02-01
Refactored `mobile-menu.js`

adamzap:
- Stopped using jQuery

You can test this by making your window narrow and clicking this button in the top-right corner:

image


djangoproject/static/js/djangoproject.js
```diff
@@ -56,3 +56,11 @@ document.querySelectorAll('.console-block label').forEach(function (el) {
     observer.observe(el);
   });
 })();
+
+// Toggle mobile menu on button click
+document.querySelector('.menu-button').addEventListener('click', function () {
+  const menu_el = document.querySelector('#top [role="navigation"]');
```
adamzap, line 62: I simplified the selector here since [an `id` is present on the element](https://github.com/django/djangoproject.com/blob/5632574606783575c5de886d7092635b4f3971dc/djangoproject/templates/includes/header.html#L1). Before it was `[role="banner"] [role="navigation"]`.

djangoproject/templates/base.html
```diff
@@ -137,7 +137,6 @@
           "jquery.flot": extless("{% static 'js/lib/jquery.flot.min.js' %}"),
           "mod/floating-warning": extless("{% static 'js/mod/floating-warning.js' %}"),
           "mod/list-collapsing": extless("{% static 'js/mod/list-collapsing.js' %}"),
-          "mod/mobile-menu": extless("{% static 'js/mod/mobile-menu.js' %}"),
```
adamzap, line 140: 😎

pauloxnet (approved):
LGTM

bmispelon:
Good call on removing the file from `base.html` too 😁

PR #1928, opened 2025-02-04
Add custom sidebar styles and update base template structure

payaljhamta:


pauloxnet:
> This pull request proposes the following enhancements:
> 
> Fixed Sidebar: Implement a fixed sidebar on the left side of documentation pages that remains visible as users scroll, providing constant access to the list of topics.

Thanks for this PR.

I have 2 first questions:
- it would be better to use the `aside` element instead of another `div` (See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/aside)
- can you test also how it behaves in mobile screen?

Another UX changes that would be better to be reviewed from @django/accessibility :)

payaljhamta:
> > This pull request proposes the following enhancements:
> > Fixed Sidebar: Implement a fixed sidebar on the left side of documentation pages that remains visible as users scroll, providing constant access to the list of topics.
> 
> Thanks for this PR.
> 
> I have 2 first questions:
> 
> * it would be better to use the `aside` element instead of another `div` (See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/aside)
> * can you test also how it behaves in mobile screen?
> 
> Another UX changes that would be better to be reviewed from @django/accessibility :)



Thanks for the update! @pauloxnet 

1.  I’ve replaced the div with the aside element as suggested (per MDN). 
2.  Also, I’ve tested the layout on mobile screens, Refer this Video 

https://github.com/user-attachments/assets/512a6ab0-caed-45eb-8ecc-a5d09bbd6f7d

payal-crossml:
@pauloxnet  Looking forward to the response on the PR.

pauloxnet:
> @pauloxnet  Looking forward to the response on the PR. 

Please don't tag me in every message. It's rude and it waste the few free time I have to review all PR. I already reviewed this PR and asked a review from the accessibility team. We would wait for them.

Please be patient, everyone here are doing this for free.

MHLut:
Hello! Accessibility Team member here; thank you for your contribution!

I do not have time for a full review right now, but I would like to request a revert on using `aside` for the sidebar.

The aside element is only for side content loosely related to the page's main content. With content, I mean the actual text of a specific page, not fixed elements like a page header, footer, and navigation.

Looking at the sidebar, I see it requires more HTML changes beyond this PR's scope, which must be discussed first. Therefore, I recommend keeping this PR as specific as possible and not refactoring beyond what is needed to get this feature to work.

payaljhamta:
> MHLut

@MHLut Thank you for responding ! Reverted the <aside> changes as requested. Let me know if anything else is needed!

MHLut:
Per Paulo's comment above, please do not tag people with every comment or random tickets.

We are volunteers and do not have infinite time, or energy, or priority to spend on issues.

PR #1929, opened 2025-02-04
Refactor SCSS and HTML templates for improved code snippet presentation

payaljhamta:
Added **'Click to Copy**' functionality for command-line code blocks using existing **clippify** integration

Closes: https://github.com/django/djangoproject.com/issues/1276

Refer : 
Screenshot from 2025-02-04 15-23-34

### DIFF
--- a/djangoproject/templates/releases/download.html
+++ b/djangoproject/templates/releases/download.html
@@ -33,13 +33,13 @@
   <p>Linux / macOS:</p>
 
 
-  <div class="snippet"> 
+  <div class="snippet">
     <pre class="literal-block snippet-filename highlight"><code>python -m pip install Django=={{ current.version }}</code></pre>
   </div>
 
-  
+
   <p>Windows:</p>
-  <div class="snippet"> 
+  <div class="snippet">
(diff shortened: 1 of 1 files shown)


payaljhamta:
@MHLut 

Hello! If possible, could you share feedback on this ticket?

MHLut:
@payaljhamta Pleaso do not randomly tag people into issues. It is incredibly rude.

I appeciate you want to draw attention to your ticket, but there are hundreds of people who want the same thing for their tickets. Patience is a virtue.

With all due respect, random tags make people less inclined to help, which is the opposite of what you are trying to achieve here.

bmispelon (changes requested):
Hi and thanks for giving a go at this issue.

I've left a comment on the changes in the SCSS file, I think a different approach is needed.
Overall I'm also not super convinced about this approach in general: I don't like that we need to add an extra `<div>` to support this, and that we're using the old `snippet-filename` class which was made for really old versions of the Django documentation.

It might be worth waiting until #1968 is finished too, since that's likely to interfere with the changes in this PR.

adamzap:
@payaljhamta Could you address the comments on this PR if you're still interested in working on it? Otherwise it will be closed in a week. Thanks!

Issue #1657, opened 2024-10-08
Read-only API for retrieving Trac data


laymonage:
Per discussion in https://github.com/django/code.djangoproject.com/issues/198#issuecomment-2398110549.

The same XmlRpcPlugin for Trac that we removed in https://github.com/django/code.djangoproject.com/pull/198 was also responsible for the JSON RPC (`/jsonrpc` endpoint). The endpoint seems useful for extracting data to gain insights, e.g.
- the "Who contributed to Django X.Y?" project by Katie from DjangoCon Europe 2022: https://github.com/glasnt/dceu2022-sprints-work
- demo by Tobias from this gist: https://gist.github.com/rixx/422392d2aa580b5d286e585418bf6915

Instead of restoring the old API, which relies on an old Trac plugin and does not give us full control over the output, we probably want to create our own API.

There is a draft PR at #1656.

For now, I think it's OK to limit the scope of the API to be read-only. If we find use cases for a write API, we can file a separate issue.

thibaudcolas:
I’d love to see this happen, as this kind of API endpoint seems to be the only reliable way to check people’s commenting and triaging activity. For example, this would help in assessing people’s [eligibility as candidates](https://github.com/django/deps/blob/main/final/0012-steering-council.rst#specification) for Steering Council elections, as triaging contributions are one of the areas where we acknowledge candidates’ work:

> Reviewing pull requests and/or triaging Django project tickets

So I need a way to assess how often someone helps by:

- Accepting or closing tickets
- Sharing git bisect results to find a regression commit
- Answering questions
- Linking tickets

---

In the meantime I’m using `site:code.djangoproject.com inurl:/ticket/ "by Person Name"` in Google to find comments, but it’s proving unreliable. I might scrape the tickets instead.

sarahboyce:
An idea was discussed in the DSF office hours to have a GitHub action for `django/django` which would add labels to PRs based off their status in Trac to acheive easy filtering within GitHub

So for example, adding a "needs review" label (which would check the ticket from the title and needs_docs=0, needs_tests=0, patch_needs_improvement=no), a "needs tests" label etc.

Having an api which includes these flags for tickets would be really great 💚 

I don't mind working on this but would appreciate an agreement on what would be the minimum work required to get this accepted

PR #1931, opened 2025-02-06
Added Trac api to receive information on specific ticket.

sarahboyce:
Relates to https://github.com/django/djangoproject.com/issues/1657
This has a very similar approach to https://github.com/django/djangoproject.com/pull/1656

This only adds an endpoint to receive information on a specific ticket.

This api is simpler than a list endpoint (which would be more useful for statistics) as it doesn't require filters and pagination but is still useful for things such as GitHub actions on PRs to query the associated ticket.

I have avoided information such as owner/reporter as this may contain emails.
I feel we can always add more fields/information in later

I think we will need the list endpoint before the associated issue can be closed (due to the request for statistics)
But I hope we can get this in as something small to get us started :crossed_fingers:

### DIFF
--- a/tracdb/views.py
+++ b/tracdb/views.py
@@ -1,6 +1,9 @@
 from django import db
-from django.shortcuts import render
+from django.http import JsonResponse
+from django.shortcuts import get_object_or_404, render
+from django.views.decorators.http import require_http_methods
 
+from .models import Ticket
 from .tractime import timestamp_to_datetime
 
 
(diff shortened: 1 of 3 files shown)


tracdb/views.py
```diff
@@ -29,3 +32,25 @@ def bouncing_tickets(request):
...
+            "id",
+            "type",
+            "component__name",
+            "summary",
+            "description",
+            "severity",
+            "status",
+            "resolution",
+            "custom",
+            "keywords",
```
alexgmin, line 52: Would it make sense to change the `keywords` field from something like `"migrate primary_key"` to `["migrate", "primary_key"]`?

It's the same the Trac UI does when you look at a ticket, each word is a different link. I assume they just split based on spaces. https://code.djangoproject.com/ticket/22997

sarahboyce: I think that's a very good idea :100:

sarahboyce: For my use case, I only need the information within "custom". 

I have therefore removed "keywords" and "component__name", for the same reason as to not include the reporter/author etc. I do not want to have this be held back by the decision/discussion on how we want to serialize our data and instead want to land something pretty basic and uncontroversial in order to allow future work (with GitHub actions etc)

Also discussed this with @alexgmin :+1:

bmispelon (approved):
Tight scope, quality code, and tests? I mean I think I'm legally required to approve this PR 😻 

I think it would be nice to make it something a bit more "proper", maybe with some built-in documentation and more methods, but I'm OK with starting small and gather some requirements based on actual usage.

bmispelon:
Change is live, thanks everyone! 🎸

PR #1933, opened 2025-02-06
Refactored `search-key.js`

adamzap:
- Stopped using jQuery
- Split concerns and add precondition checks to reduce nesting
- Add `preventScroll: true` option to `focus` call
- Stop passing default `block` option to `scrollIntoView` call

image

To test this:

1. Ensure that the search input placeholder string ends in "(⌘ + K)" on a macOS and "(Ctrl + K)" other operating systems. You can spoof your operating system in the Chrome DevTools network settings.
2. Ensure that the advertised key combination focuses and scrolls to the search input.


djangoproject/static/js/djangoproject.js
```diff
@@ -64,3 +64,37 @@ document.querySelector('.menu-button').addEventListener('click', function () {
   this.classList.toggle('active');
   menu_el.classList.toggle('active');
 });
+
+// Update search input placeholder text based on the user's operating system
+(function () {
+  const el = document.querySelector('#id_q');
```
adamzap, line 70: I'm using the ID instead of a class to select the element since one was already present. Also, I used `querySelector` instead of `getElementById` to stay consistent with the rest of the code in this file.

djangoproject/static/js/djangoproject.js
adamzap, line 1: I split the functionality from `search-key.js` into two sections:

1. Setting the search input placeholder string
2. Handling the key combination to focus the search input

I think selecting the input twice is a small price to pay for simpler, better organized code.

djangoproject/static/js/djangoproject.js
```diff
@@ -64,3 +64,37 @@ document.querySelector('.menu-button').addEventListener('click', function () {
...
+
+  const initial_value = el.getAttribute('placeholder');
+  const is_mac = navigator.userAgent.indexOf('Mac') !== -1;
+  const new_value = `${initial_value} (${is_mac ? '⌘' : 'Ctrl'} + K)`;
+
+  el.setAttribute('placeholder', new_value);
+})();
+
+// Focus, select, and scroll to search input when key combination is pressed
+window.addEventListener('keydown', function (e) {
```
adamzap, line 84: I used precondition checks here to reduce nesting. Let me know if anyone has an issue with this style.

djangoproject/static/js/djangoproject.js
```diff
@@ -64,3 +64,37 @@ document.querySelector('.menu-button').addEventListener('click', function () {
...
+  if (!(e.metaKey || e.ctrlKey) || e.key !== 'k') {
+    return;
+  }
+
+  if (['INPUT', 'TEXTAREA'].includes(document.activeElement.tagName)) {
```
adamzap, line 89: The `document.activeElement` check replaces the `:focus` part of the jQuery-based selector on line 26 of the old file.

djangoproject/static/js/djangoproject.js
```diff
@@ -64,3 +64,37 @@ document.querySelector('.menu-button').addEventListener('click', function () {
...
+    return;
+  }
+
+  e.preventDefault();
+
+  const el = document.querySelector('#id_q');
+
+  el.select();
+  el.focus({ preventScroll: true });
+  el.scrollIntoView({ behavior: 'smooth' });
```
adamzap, line 99: I dropped the `block: 'start'` option because it's the default: https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView#block

djangoproject/static/js/djangoproject.js
```diff
@@ -64,3 +64,37 @@ document.querySelector('.menu-button').addEventListener('click', function () {
...
+  if (['INPUT', 'TEXTAREA'].includes(document.activeElement.tagName)) {
+    return;
+  }
+
+  e.preventDefault();
+
+  const el = document.querySelector('#id_q');
+
+  el.select();
+  el.focus({ preventScroll: true });
```
adamzap, line 98: Using [`preventScroll: true`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus#preventscroll) here improves the user experience of this handler. Previously, both `focus` and `scrollIntoView` were scrolling the viewport, so the user would get an immediate scroll to the top of the document followed by a smooth scroll to the search input. With this change, the user only gets a smooth scroll to the search input.

djangoproject/static/js/djangoproject.js
```diff
@@ -64,3 +64,37 @@ document.querySelector('.menu-button').addEventListener('click', function () {
...
+window.addEventListener('keydown', function (e) {
+  if (!(e.metaKey || e.ctrlKey) || e.key !== 'k') {
+    return;
+  }
+
+  if (['INPUT', 'TEXTAREA'].includes(document.activeElement.tagName)) {
+    return;
+  }
+
+  e.preventDefault();
```
adamzap, line 93: Using `e.preventDefault` here seems more understandable than returning `false` from the event handler at the bottom.

djangoproject/static/js/djangoproject.js
pauloxnet: Do you think "getElementById" would be more efficient? If so, I would use it.

djangoproject/static/js/djangoproject.js
```diff
@@ -64,3 +64,37 @@ document.querySelector('.menu-button').addEventListener('click', function () {
...
+
+// Update search input placeholder text based on the user's operating system
+(function () {
+  const el = document.querySelector('#id_q');
+
+  if (!el) {
+    return;
+  }
+
+  const initial_value = el.getAttribute('placeholder');
```
pauloxnet, line 76: I would use more descriptive variable names (e.g. `original_placeholder`)

djangoproject/static/js/djangoproject.js
```diff
@@ -64,3 +64,37 @@ document.querySelector('.menu-button').addEventListener('click', function () {
...
+  const is_mac = navigator.userAgent.indexOf('Mac') !== -1;
```
pauloxnet, line 77: In the original script the agent query logic was reversed, do you think it would be better to leave it as it was?

pauloxnet (commented):
Great job. I tried to review the PR commenting with some suggestions or questions.

djangoproject/static/js/djangoproject.js
adamzap: Yeah, it's a small fraction of a millisecond faster. I'll change it since that seems to be your preference.

adamzap: Fixed in 8e58b4fc.

djangoproject/static/js/djangoproject.js
adamzap: I personally think short-lived variable names like that can be a bit more generic, but I'm happy to change it. Fixed in a9bce985.

djangoproject/static/js/djangoproject.js
adamzap: I was probably trying to identify the special case on one line then construct the string on the next line, so this makes sense to me. I can change it if you want me to though, just let me know.

djangoproject/static/js/djangoproject.js
pauloxnet: The "speaking" variables are for people that will read your code after you.

pauloxnet (approved):

djangoproject/static/js/djangoproject.js
adamzap: Sorry, could you elaborate? This variable only lives for three lines, and I was thinking that `el.getAttribute('placeholder')` makes it pretty obvious that it's the original placeholder. I'm fine with the new variable name though. I'm not sure what you mean by "speaking".

pauloxnet: Sorry, I wrongly used the word. No problem with the variable names, but I still think that "explicit is better than implicit ", also for variable names that are obvious. Nothing wrong with your code. I think I already approved the PR.

adamzap: No worries! Thanks for elaborating.

djangoproject/static/js/djangoproject.js
```diff
@@ -64,3 +64,37 @@ document.querySelector('.menu-button').addEventListener('click', function () {
...
+(function () {
+  const el = document.getElementById('id_q');
+
+  if (!el) {
+    return;
+  }
+
+  const original_placeholder = el.getAttribute('placeholder');
+  const is_mac = navigator.userAgent.indexOf('Mac') !== -1;
+  const new_value = `${original_placeholder} (${is_mac ? '⌘' : 'Ctrl'} + K)`;
```
medmunds, line 78: Minor nit: on Apple platforms, keyboard shortcuts are usually formatted like ⌘ K without the plus sign (see pictures [here](https://developer.apple.com/design/human-interface-guidelines/keyboards#Platform-considerations) and [here](https://developer.apple.com/design/human-interface-guidelines/menus)).

```suggestion
  const new_value = `${original_placeholder} (${is_mac ? '⌘ K' : 'Ctrl + K'})`;
```

(I used narrow no-break space `\u202F` in there, which looks a little better than a full space and avoids any risk of weird wrapping.)

adamzap: Do you want to open a new issue about this since the PR was merged?

medmunds: Done, #1960

Issue #1734, opened 2024-11-13
Friction in finding specific information via a web search


jonbiemond:
When accessing the documentation via a web search (such as Google), the user is required to take an additional step of navigating the page to find the relevant section. This adds friction to the process of finding specific information for a function or feature.

For example, a [Google search for "django window expressions"](https://www.google.com/search?q=django+window+expression), correctly returns the [Django Query Expressions page](https://docs.djangoproject.com/en/5.1/ref/models/expressions/) as my first result.  However, once I've navigated to the page I'm presented with "Supported arithmetic", quite far off from a window expression. A less optimistic user might abandon the page entirely at this point. I will utilize my browser's find tool to search for "window" (skipping past two unrelated hits), scroll foolishly or scan through the Contents side bar. Finally, I get to [the Windows functions section](https://docs.djangoproject.com/en/5.1/ref/models/expressions/#window-functions) half way down.

#### Possible Solutions
* Move the Contents side bar to the left side. This has the advantage of being the first place one looks (for left-to-right language readers) and moving navigation links to up in the find tool's results.
* Have expandable sections in the Contents side bar. This could also just be worse.
* Break out the documentation into more pages. If the pages are more concise there is less for the user to look through. The jQuery docs do this well. Here is the first link in the results of my search for "jquery remove element": https://api.jquery.com/remove/

Related to #1495

thibaudcolas:
Thank you for reporting this @jonbiemond! Do you prefer using the google search or is it just because the built-in site search doesn’t work?

jonbiemond:
@thibaudcolas I use Google search mostly due to force of habit. So more of a preference I suppose, rather than a reaction to the experience of the built-in search.

sarahboyce:
Considering that the results already have text highlighting, I feel like it would be possible for us to implement a [scroll to text fragment](https://trevorfox.com/2020/06/how-to-highlight-text-on-a-page-with-scroll-to-text-fragments/) which would make the navigation a little nicer 🤔

PR #1934, opened 2025-02-07
Added scroll to text fragment to search.

sarahboyce:
Following on from this [comment](https://github.com/django/djangoproject.com/issues/1734#issuecomment-2640518969), I have added scroll to text fragment to our search result urls.

Before, when you search for "StackedInline", you get these results: https://docs.djangoproject.com/en/5.1/search/?q=StackedInline

search results from StackedInline docs search

If you click the first one, the url is: https://docs.djangoproject.com/en/5.1/ref/contrib/admin/
This navigates you to the top of the page.

After these changes the url is: https://docs.djangoproject.com/en/5.1/ref/contrib/admin/#:~:text=StackedInline%5Bsource%5D 
This navigates to StackedInline in view (when the browser supports it, which is most browsers)

Admin page with line highlighted and in middle of page

~This was difficult to test locally.~ (found the management command)

If there is an issue with the logic, the text fragment will not be found and we still land at the top of the page. e.g https://docs.djangoproject.com/en/5.1/ref/contrib/admin/#:~:text=potato

### DIFF
--- a/docs/templates/docs/search_results.html
+++ b/docs/templates/docs/search_results.html
@@ -36,7 +36,7 @@
         {% for result in page.object_list %}
           <dt>
             <h2 class="result-title">
-              <a href="{% url 'document-detail' lang=result.release.lang version=result.release.version url=result.path host 'docs' %}">{{ result.headline|safe }}</a>
+              <a href="{% url 'document-detail' lang=result.release.lang version=result.release.version url=result.path host 'docs' %}{% if not start_sel in result.headline %}{{ result.highlight|fragment }}{% endif %}">{{ result.headline|safe }}</a>
             </h2>
             <span class="meta breadcrumbs">
               {% for breadcrumb in result.breadcrumbs %}
--- a/docs/templatetags/docs.py
+++ b/docs/templatetags/docs.py
@@ -1,4 +1,8 @@
+import re
+from urllib.parse import quote
+
 from django import template
+from django.template.defaultfilters import stringfilter
 from django.utils.safestring import mark_safe
 from django.utils.version import get_version_tuple
 from pygments import highlight
@@ -7,6 +11,7 @@ from pygments.lexers import get_lexer_by_name
 
 from ..forms import DocSearchForm
 from ..models import DocumentRelease
+from ..search import START_SEL, STOP_SEL
 from ..utils import get_doc_path, get_doc_root
 
 register = template.Library()
@@ -81,3 +86,38 @@ def do_pygments(parser, token):
     nodelist = parser.parse(("endpygment",))
     parser.delete_first_token()
     return PygmentsNode(parser.compile_filter(tokens[1]), nodelist)
+
+
+@register.filter(name="fragment")
+@stringfilter
+def generate_scroll_to_text_fragment(highlighted_text):
+    """
+    Given the highlighted text generated from Document.objects.search()
+    constructs a scroll to text fragment.
+
+    This will not work when:
+    *  the highlighted test starts from a partial word, e.g it starts from
+       test_environment rather than DiscoverRunner.setup_test_environment().
+    *  it is trying to highlight a Python code snippet and the spacing logic
+       has fallen down e.g. test = 5 not test=5 but test(a=5) not test(a = 5).
+    """
+    first_non_empty_line = next(
+        (
+            stripped
+            for line in highlighted_text.split("\n")
+            if (stripped := line.strip())
+        ),
+        "",
+    )
+    # Remove highlight tags and unwanted symbols.
+    line_without_highlight = re.sub(
+        rf"{START_SEL}|{STOP_SEL}|¶", "", first_non_empty_line
+    )
+    line_without_highlight = line_without_highlight.replace("&quot;", '"')
+    # Remove excess spacing.
+    single_spaced = re.sub(r"\s+", " ", line_without_highlight).strip()
+    # Handle punctuation spacing.
+    single_spaced = re.sub(r"\s([.,;:!?)(\]\[])", r"\1", single_spaced)
+    # Due to Python code such as timezone.now(), remove the space after a bracket.
+    single_spaced = re.sub(r"([(\[])\s", r"\1", single_spaced)
+    return f"#:~:text={quote(single_spaced)}"
--- a/docs/models.py
+++ b/docs/models.py
@@ -28,6 +28,8 @@ from . import utils
 from .search import (
     DEFAULT_TEXT_SEARCH_CONFIG,
     DOCUMENT_SEARCH_VECTOR,
+    START_SEL,
+    STOP_SEL,
     TSEARCH_CONFIG_LANGUAGES,
 )
 
@@ -265,15 +267,15 @@ class DocumentQuerySet(models.QuerySet):
                     headline=SearchHeadline(
                         "title",
                         search_query,
-                        start_sel="<mark>",
-                        stop_sel="</mark>",
+                        start_sel=START_SEL,
+                        stop_sel=STOP_SEL,
                         config=models.F("config"),
                     ),
                     highlight=SearchHeadline(
                         KeyTextTransform("body", "metadata"),
                         search_query,
-                        start_sel="<mark>",
-                        stop_sel="</mark>",
+                        start_sel=START_SEL,
+                        stop_sel=STOP_SEL,
                         config=models.F("config"),
                     ),
                     breadcrumbs=models.F("metadata__breadcrumbs"),
--- a/docs/search.py
+++ b/docs/search.py
@@ -48,3 +48,6 @@ DOCUMENT_SEARCH_VECTOR = (
         KeyTextTransform("parents", "metadata"), weight="D", config=F("config")
     )
 )
+
+START_SEL = "<mark>"
+STOP_SEL = "</mark>"
--- a/docs/tests.py
+++ b/docs/tests.py
@@ -19,7 +19,7 @@ from releases.models import Release
 
 from .models import DOCUMENT_SEARCH_VECTOR, Document, DocumentRelease
 from .sitemaps import DocsSitemap
-from .templatetags.docs import get_all_doc_versions
+from .templatetags.docs import generate_scroll_to_text_fragment, get_all_doc_versions
 from .utils import get_doc_path, sanitize_for_trigram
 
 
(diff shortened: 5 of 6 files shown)


sarahboyce:
> For example, if I search for the term "databases" I get many excerpts of the target document, but several pieces of text are appended to the link and the mechanism doesn't work
> 
> E.g.: http://docs.djangoproject.localhost:8000/en/5.1/ref/databases/#:~:text=database%20configuration%20inDATABASES%20%3A

I don't 100% understand the logic around spacing in order to strip the extra spaces correctly but in this case, the correct url would be:
http://docs.djangoproject.localhost:8000/en/5.1/ref/databases/#:~:text=database%20configuration%20in%20DATABASES%3A
So, "in DATABASES:" not "inDATABASES :"

docs/templatetags/docs.py
```diff
@@ -81,3 +84,20 @@ def view_article(request, pk):
...
+def generate_scroll_to_text_fragment(highlighted_text):
+    """
+    Given the highlighted text generated from Document.objects.search()
+    constructs a scroll to text fragment.
+    """
+    clean_text = (
+        highlighted_text.replace("<mark>", "")
+        .replace("</mark>", "")
+        .replace("¶", "")
+        .replace("  ", "")
```
sarahboyce, line 100: Note that this space cleaning logic is currently incorrect. We might be better to do the text start and text end notation so for example: https://docs.djangoproject.com/en/5.1/ref/databases/#:~:text=database%20configuration,in%20DATABASES

pauloxnet:
> > For example, if I search for the term "databases" I get many excerpts of the target document, but several pieces of text are appended to the link and the mechanism doesn't work
> > E.g.: http://docs.djangoproject.localhost:8000/en/5.1/ref/databases/#:~:text=database%20configuration%20inDATABASES%20%3A
> 
> I don't 100% understand the logic around spacing in order to strip the extra spaces correctly but in this case, the correct url would be: http://docs.djangoproject.localhost:8000/en/5.1/ref/databases/#:~:text=database%20configuration%20in%20DATABASES%3A So, "in DATABASES:" not "inDATABASES :"

The text that `SearchHeadline` returns is some excerpts from the `HTML` document in the `JSONfield`.

This document is exactly the one generated by `Sphinx` and contains all the tags to show the code or for formatting.

PostgreSQL does a great job of stripping all the HTML tags, but what remains now is something like this.

```HTML
<dt>
  <h2 class="result-title">
    <a href="http://docs.djangoproject.localhost:8000/en/5.1/ref/databases/"><mark>Databases</mark></a>
  </h2>
  <span class="meta breadcrumbs">
      <a href="http://docs.djangoproject.localhost:8000/en/5.1/ref/">API Reference</a>
  </span>
</dt>
<dd>
…&nbsp;<mark>database</mark> configuration in    <mark>DATABASES</mark>   : 
settings.py      
<mark>DATABASES</mark>   =   { 
  "default"  :   { 
    "ENGINE"  :   "django.db.backends.postgresql"  , 
    "OPTIONS&nbsp;…
</dd>
```
To clarify, this text above in the database is stored as you see it on the website, with all the HTML code (PostgreSQL removes it all and applies the `<mark>`)
```HTML
database configuration in <a class="reference internal" href="[../settings/#std-setting-DATABASES](view-source:http://docs.djangoproject.localhost:8000/en/5.1/ref/settings/#std-setting-DATABASES)"><code class="xref std std-setting docutils literal notranslate"><span class="pre">DATABASES</span></code></a>:</p>
<div class="literal-block-wrapper docutils container" id="id15">
<div class="code-block-caption"><span class="caption-text"><code class="docutils literal notranslate"><span class="pre">settings.py</span></code></span><a class="headerlink" href="[#id15](view-source:http://docs.djangoproject.localhost:8000/en/5.1/ref/databases/#id15)" title="Link to this code">¶</a></div>
<div class="highlight-python notranslate"><div class="highlight"><pre><span></span><span class="n">DATABASES</span> <span class="o">=</span> <span class="p">{</span>
    <span class="s2">&quot;default&quot;</span><span class="p">:</span> <span class="p">{</span>
        <span class="s2">&quot;ENGINE&quot;</span><span class="p">:</span> <span class="s2">&quot;django.db.backends.postgresql&quot;</span><span class="p">,</span>
        <span class="s2">&quot;OPTIONS&quot;</span><span class="p">:</span> <span class="p">{</span>
```

Trying to search directly in an HTML document without PostgreSQL would be much more complicated, and the extracted snippets are clean enough to show in search results, but manipulating them or wanting to extract a quote from them would require more work than simply copying them.

The main problem is that there is a lot of code interspersed with text in the documentation. It wouldn't be such a big problem if we were searching Wikipedia texts for example.

I can't think of a good way to do this at the moment, I'll try to think about it over the weekend.

sarahboyce:
Thank you! This is really really helpful
It looks like each html tag is stripped and replaced with a space, these spaces build up when there are nested html tags

sarahboyce:
Given the complications around the mixture of text and code, I think getting a 100% success rate is probably not realistic. :thinking: 

I think let's try to aim for a 80-90% success rate (so 8/9 out of 10 search results highlights correctly)
I pushed up changes and it should be better :crossed_fingers: 

If anyone wants to test this when you share feedback, please share the success percentage (number of successes/total number of results tested) and details on the failures :heart:

tim-schilling:
The text fragment highlight seems  better suited when the text start and end options are specified so that it can be more exact. That would help avoid the problem of highlighting an instance before the one in the search result.

From what I understand, and search that it's in a term that was earlier in the doc won't be scrolled to without a more exact directive.

---

Do we have a list of the top 25 searches easily accessible? That may help with testing.

sarahboyce:
> The text fragment highlight seems better suited when the prefix and suffix options are specified so that it can be more exact. That would help avoid the problem of highlighting an instance before the one in the search result.

This makes the url shorter rather than more precise. It can be less precise if you are not specific enough for the prefix and suffix

> From what I understand, and search that it's in a term that was earlier in the doc won't be scrolled to without a more exact directive.

That's correct 👍
This is also what I'm doing I believe as I am not adding the search term but a section of the highlighted text from the search results

> Do we have a list of the top 25 searches easily accessible? That may help with testing.

Oooooh, this is a great idea 👀
@bmispelon might be able to help with this ❤️

tim-schilling:
I got off my phone and went to a laptop.

> That's correct 👍
This is also what I'm doing I believe as I am not adding the search term but a section of the highlighted text from the search results

Sounds good. I also did a few searches and it looks like it's unlikely for a term to appear in the search results twice for the same page.

pauloxnet:
> I got off my phone and went to a laptop.
> 
> > That's correct 👍
> This is also what I'm doing I believe as I am not adding the search term but a section of the highlighted text from the search results
> 
> Sounds good. I also did a few searches and it looks like it's unlikely for a term to appear in the search results twice for the same page.

If you try generic search with terms like 'django' or 'database' you can see the terms highlighted in the title and more than one in the excerpt.

tim-schilling:
> If you try generic search with terms like 'django' or 'database' you can see the terms highlighted in the title and more than one in the excerpt.

I was able to produce it with `form` too. My opinion is that those generic terms are unlikely to be searched for. I'm happy to be proven wrong with data, but otherwise I'm in favor of seeing this released.

docs/templatetags/docs.py
```diff
@@ -81,3 +84,32 @@ def view_article(request, pk):
     nodelist = parser.parse(("endpygment",))
     parser.delete_first_token()
     return PygmentsNode(parser.compile_filter(tokens[1]), nodelist)
+
+
+@register.filter(name="fragment")
+@stringfilter
+def generate_scroll_to_text_fragment(highlighted_text):
```
pauloxnet, line 91: This method seems to be returning better results now. I would try to optimize it a bit, perhaps using a regex instead of the many occurrences of replace or the for cycle over punctuations. Also I would try defining "<mark>" in a variable in the search queryset to reuse here.

docs/templates/docs/search_results.html
```diff
@@ -36,7 +36,7 @@ <h2>{% trans "No search query given" %}</h2>
         {% for result in page.object_list %}
           <dt>
             <h2 class="result-title">
-              <a href="{% url 'document-detail' lang=result.release.lang version=result.release.version url=result.path host 'docs' %}">{{ result.headline|safe }}</a>
+              <a href="{% url 'document-detail' lang=result.release.lang version=result.release.version url=result.path host 'docs' %}{{ result.highlight|fragment }}">{{ result.headline|safe }}</a>
```
pauloxnet, line 39: As a user, when I click on a title that contains the highlighted search term, I would like to be redirected to the title, and not to the first occurrence of the term on the page.

pauloxnet (commented):
Thanks, @sarahboyce, for the additional work here. I suggested a couple of further improvements before merging the PR. Then I would gather some feedback from users to see how to further improve the feature.

docs/templates/docs/search_results.html
sarahboyce: I agree with that from a user perspective.
In order to achieve that, I need the database query to be tweaked so that it returns the title of the section of the highlighted text. If we can do that, the existing function would work :+1:

docs/templatetags/docs.py
sarahboyce: > using a regex instead of the many occurrences of replace

Done :+1: 

> I would try defining "" in a variable in the search queryset to reuse here.

Sorry, I'm not following, ca you clarify what you mean? :slightly_smiling_face:

docs/templates/docs/search_results.html
pauloxnet: I updated my comment above. :)

docs/templatetags/docs.py
pauloxnet: > > I would try defining "" in a variable in the search queryset to reuse here.
> 
> Sorry, I'm not following, ca you clarify what you mean? 🙂

Done

docs/templates/docs/search_results.html
sarahboyce: I think I achieved what you wanted :crossed_fingers:

sarahboyce: Now, if the document title is highlighted (which is likely when you search for generic things like "database" "admin" etc), then the url is as normal and you go to the page. 

If the title is not highlighted but there are highlights in the text, you go to that area of the text (this is likely when you searched for something like "StackedInline", "ManyToManyField" etc)

I think that's quite nice :thinking:

pauloxnet (approved):

docs/templates/docs/search_results.html
pauloxnet: Yes, I agree. Thanks for all the improvements. I'll merge it. ;-)

charettes:
This is a such a game changer @sarahboyce and @pauloxnet thanks for much for that!

I'm glad we went with a solution that isn't perfect but at least allows a form of deep linking.

Did we end up figuring out a way to match against section title (instead of fist occurrence) on an exact match? For example, I would have expected [the first search result for `ArrayField`](https://docs.djangoproject.com/en/5.1/search/?q=ArrayField) to deep-link to [the exact section](https://docs.djangoproject.com/en/5.1/ref/contrib/postgres/fields/#arrayfield).

sarahboyce:
> Did we end up figuring out a way to match against section title (instead of fist occurrence) on an exact match? For example, I would have expected [the first search result for ArrayField](https://docs.djangoproject.com/en/5.1/search/?q=ArrayField) to deep-link to [the exact section](https://docs.djangoproject.com/en/5.1/ref/contrib/postgres/fields/#arrayfield).

Hoping https://github.com/django/djangoproject.com/pull/1947 will be a bit more of a robust solution for this
Glad you like it :grin:

Issue #1628, opened 2024-09-26
Docs search: tweak results ranking so release notes have lower priority


thibaudcolas:
Related: [Site-wide search #1499](https://github.com/django/djangoproject.com/issues/1499). We want to tweak documentation search results so searches like "TextChoices" favor reference documentation over release notes.

Tasks:

- [ ] Review and share common search queries in access logs / Postgres
- [ ] @pauloxnet suggest possible query to retrieve useful results
- [ ] Review rankings for top N queries
- [ ] Decide how to proceed with ranking

thibaudcolas:
Assigning to @bmispelon to provide analytics on search from access logs / Postgres stats, for other contributors to review.

pauloxnet:
As we saw with Jack it will help more to be able to extract the various
sections from the HTML content of the Document and rate them separately.

pauloxnet:
~~As we saw with Jack it will help more to be able to extract the various sections titles from the HTML content of the Document and rate them separately.~~

Actually we already are using the `toc` key in the JSON generated by Sphinx , which contains the titles of all the paragraphs. In the web page you can see the list under "Contents" title.

sarahboyce:
Note that our documentation is structured in folders and this is likely to remain very stable (changing breaks urls)
The release notes are in a folder, as are the reference docs, how tos etc. 

We could add a filter for search results based off these folders. In my head I picture that after a search, on the search result page there are pills with each category that existed in the search. If you click on one, it becomes an "only results in that category" filter and the other categories are grayed out and inactive. When clicking on those it toggles and there could be a clear filter link. That way you could click to see "only reference docs"

I think we can tweak the ranking but we need to be careful when people are specifically searching for release notes. This might be able to let folks refine their search easier without being completely reliant on our ranking calculation

thibaudcolas:
I think this is essentially "done" as of #1935 from Sarah! Not the approach I had in mind but works just as well if not better ☺️ We still have a need for search analytics for #1097 but the logs option seems to have stalled so I’ll close this and we can keep looking at that in #2288.

PR #1935, opened 2025-02-07
Added documentation type filtering to search.

sarahboyce:
In order to make it easier to refine the search results for the type of information you want, added a filter by docs categories. There are more categories (e.g. "Internals") but I think it is better to only include the most likely categories

This slightly relates to https://github.com/django/djangoproject.com/issues/1628

#### Mobile view

Search results with new filter - mobile

#### Desktop view

Search results on desktop with new filter

### DIFF
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -2578,6 +2578,40 @@ table.docutils th {
     }
 }
 
+search.filters {
+    @include sans-serif;
+
+    display: flex;
+    gap: 10px;
+    border-bottom: 2px solid var(--hairline-color);
+    overflow-x: auto;
+    white-space: nowrap;
+    padding-bottom: 0;
+    position: relative;
+
+    a {
+      padding: 10px 20px;
+      text-decoration: none;
+      border-bottom: 3px solid transparent;
+      transition: color 0.3s ease, border-bottom 0.3s ease;
+      color: var(--text-light);
+      flex-shrink: 0;
+
+      &:not([href]) {
+        color: var(--body-fg);
+        font-weight: bold;
+        border-bottom: 3px solid var(--primary);
+      }
+
+      &[href]:focus,
+      &[href]:active,
+      &[href]:hover {
+        outline: none;
+        border-bottom: 3px solid var(--hairline-color);
+      }
+    }
+}
+
 .search-links {
     @extend .list-links;
 
--- a/docs/search.py
+++ b/docs/search.py
@@ -1,6 +1,7 @@
 from django.contrib.postgres.search import SearchVector
-from django.db.models import F
+from django.db.models import F, TextChoices
 from django.db.models.fields.json import KeyTextTransform
+from django.utils.translation import gettext_lazy as _
 
 # Imported from
 # https://github.com/postgres/postgres/blob/REL_14_STABLE/src/bin/initdb/initdb.c#L659
@@ -51,3 +52,23 @@ DOCUMENT_SEARCH_VECTOR = (
 
 START_SEL = "<mark>"
 STOP_SEL = "</mark>"
+
+
+class DocumentationCategory(TextChoices):
+    """
+    Categories used to filter the documentation search.
+    The value must match a folder name within django/docs.
+    """
+
+    # Diátaxis folders.
+    REFERENCE = "ref", _("API Reference")
+    TOPICS = "topics", _("Using Django")
+    HOWTO = "howto", _("How-to guides")
+    RELEASE_NOTES = "releases", _("Release notes")
+
+    @classmethod
+    def parse(cls, value, default=None):
+        try:
+            return cls(value)
+        except ValueError:
+            return None
--- a/docs/templates/docs/search_form.html
+++ b/docs/templates/docs/search_form.html
@@ -1,11 +1,12 @@
 {% load i18n %}
+<search class="search form-input" aria-labelledby="docs-search-label">
+  <form action="{% url 'document-search' version=version lang=lang host 'docs' %}">
+    <label id="docs-search-label" class="visuallyhidden" for="{{ form.q.id_for_label }}">{{ form.q.field.widget.attrs.placeholder }}</label>
+    {{ form.q }}
 
-<form action="{% url 'document-search' version=version lang=lang host 'docs' %}" class="search form-input" role="search">
-  <label class="visuallyhidden" for="id_q">Search:</label>
-  {{ form.q }}
-
-  <button type="submit">
-    <i class="icon icon-search"></i>
-    <span class="visuallyhidden">{% trans 'Search' %}</span>
-  </button>
-</form>
+    <button type="submit">
+      <i class="icon icon-search" aria-hidden="true"></i>
+      <span class="visuallyhidden">{% translate "Submit" %}</span>
+    </button>
+  </form>
+</search>
--- a/docs/templates/docs/search_results.html
+++ b/docs/templates/docs/search_results.html
@@ -11,6 +11,13 @@
 
 {% block body %}
   {% if query %}
+    <search class="filters">
+      <span id="search-filters" class="visually-hidden">{% translate "Filter the current search results by documentation category" %}</span>
+      <a{% if not active_category %} aria-current="page"{% else %} href="{% querystring category=None page=None %}"{% endif %}>{% translate "All" context "all documentation categories" %}</a>
+      {% for category in DocumentationCategory %}
+        <a{% if active_category == category %} aria-current="page"{% else %} href="{% querystring category=category.value page=None %}"{% endif %}>{{ category.label }}</a>
+      {% endfor %}
+    </search>
     <h2>
       {% if release.is_dev %}
         {% blocktranslate count num_results=paginator.count trimmed %}
@@ -64,6 +71,15 @@
               </ul>
             {% endif %}
           </dd>
+        {% empty %}
+          {% if active_category %}
+            <dt>
+              <p>
+                {% querystring category=None page=None as all_search %}
+                {% blocktranslate trimmed %}Please try searching <a href="{{ all_search }}">all documentation results</a>.{% endblocktranslate %}
+              </p>
+            </dt>
+          {% endif %}
         {% endfor %}
       </dl>
     </div>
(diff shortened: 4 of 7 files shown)


docs/views.py
```diff
@@ -189,6 +189,10 @@ def search_results(request, lang, version, per_page=10, orphans=3):
                 return redirect(exact)
 
             results = Document.objects.search(q, release)
+            type_value = request.GET.get("type", None)
+            filtered_by_type = type_value in ["ref", "releases", "howto", "topics"]
+            if filtered_by_type:
+                results = results.filter(metadata__parents__startswith=type_value)
```
pauloxnet, line 195:
```suggestion
            # I would define a dict of types in the search.py and then use it here and in the template
            TYPES = {
                "ref": "API Reference",
                "releases": "Release notes",
                "howto": "How-to guides",
                "topics": "Using Django",
            }
            type_key = key if (key := GET.get("type")) in TYPES.keys() else None
            # I would pass the type to the search method where the queryset logi are
            results = Document.objects.search(q, release, type_key)
```

docs/templates/docs/search_results.html
```diff
@@ -32,6 +32,13 @@ <h2>{% trans "No search query given" %}</h2>
 
   {% if query %}
     <div id="docs-content">
+      <div role="list" class="tabs">
```
pauloxnet, line 35: I would use the dict of types to dynamically build the list of tabs.

From accessibility point of view, do you think can we use a better semantic HTML tags? I don't know if `<nav>` can be a better choice, I'm not an a11y expert, but maybe we can ask for help to our @django/accessibility team. It would be great to start moving the website from an old list of `<div>` or `<span>` to a more modern semantic website.

djangoproject/scss/_style.scss
```diff
@@ -2563,6 +2563,42 @@ table.docutils th {
     }
 }
 
+.tabs {
```
pauloxnet, line 2566: As wrote below it would be great to apply the style to a `<nav>` or other "modern" HTML tag instead of a class, and maybe reuse the style of these tabs in other point on the website.

pauloxnet (commented):
I like the idea, but I would try to centralize the list of type in a single dictionary (in `search.py`) to pass to the template to dynamically build the list of tabs. I would also move the search logic into the manager method by introducing a new search attribute so as to keep all the logic in one place.

docs/views.py
sarahboyce: Lovely suggestions thank you :star:

docs/templates/docs/search_results.html
sarahboyce: I have updated to use `<nav>` and removed the naming of "tabs" (these aren't real tabs just look a little bit like them). Would love the accessibility team to check this :100:

docs/templates/docs/search_results.html
```diff
@@ -11,6 +11,12 @@
 
 {% block body %}
   {% if query %}
+    <nav class="filters" aria-label="{% trans 'Filter results by document type' %}">
```
sarahboyce, line 14: I have chosen to move this above the number of results title as this is "nicer" for when there are 0 results

docs/templates/docs/search_results.html
pauloxnet: Yes, since we're changing the UX a review from a11y point-of-view would be great.

docs/templates/docs/search_results.html
pauloxnet: Make sense to me. How it looks when there are 0 results in a "tab" ?

docs/views.py
```diff
@@ -188,7 +188,9 @@ def search_results(request, lang, version, per_page=10, orphans=3):
             if exact is not None:
                 return redirect(exact)
 
-            results = Document.objects.search(q, release)
+            type_key = request.GET.get("type")
+            type_key = type_key if type_key in DOCUMENT_TYPES.keys() else None
```
pauloxnet, line 192: I would use the walrus operator here to have one assignment. But I see that you prefer separate things. It's because of the formatting? I'm just curious :)

sarahboyce: Just the way it got formatted, I found this more readable 🙂

docs/templates/docs/search_results.html
pauloxnet: Now that I see the screenshot I have a suggestion to improve the message.

In case the search returns 0 results, I would not repeat the name of the section, which is written just above, but rather I would invite repeating the search in all the documentation with a link to perform it directly.

Something like:

```
<p>0 results for <em>potato</em> in version 5.1</p>

<p>Please try again in <a hrfe="<URL>">all documentation</a>.</p>
```

sarahboyce: With a link to search all results

:+1:

docs/templates/docs/search_results.html
knyghty: There are a couple of possibilities and I think either are fine.

`nav` is great because we're... navigating. However...

Because the text talks about filtering the results, a `search` landmark could be another option. But in this case I think it should encompass both the search bar and the tabs. I suspect `nav` could also be used in this case. I would consider this option more of a nice to have if the implementation would be annoying.

sarahboyce: Due to the way the page is structured, I can't have the tabs and the input in the same `<search>` landmark.
So I have done the following:
- updated the original search form to be within a `<search>` and added an `aria-label` of "Search the documentation"
- updated the filters to be in a `<search>` HTML tag with `aria-label` of "Filter the current search results by documentation type". This is only exists on a search result page when something was searched

I checked and from what I have read, having two `<search>` html tags on a page is fine as long as it is clear what does what. So I _think_ this is a reasonable approach, we can always change it later

knyghty: It's probably good enough but it does feel a bit weird I think if they are both contributing to the same search results.

docs/templates/docs/search_form.html
```diff
@@ -1,11 +1,12 @@
 {% load i18n %}
+<search class="search form-input" aria-label="{% trans 'Search the documentation' %}">
```
MHLut, line 2: I would advise against the use of `aria-label`, as it has some issues with automatic translation:

https://adrianroselli.com/2025/01/aria-description-does-not-translate.html

We may instead use `aria-labelledby` tied to a heading or paragraph.

docs/templates/docs/search_form.html
```diff
@@ -1,11 +1,12 @@
...
+  <form action="{% url 'document-search' version=version lang=lang host 'docs' %}">
+    <label class="visuallyhidden" for="id_q">Search:</label>
```
MHLut, line 4: I would argue that this label is useful for sighted users as well.

MHLut, line 4: "Search" is missing translation tags.

docs/templates/docs/search_form.html
```diff
@@ -1,11 +1,12 @@
...
-  <label class="visuallyhidden" for="id_q">Search:</label>
-  {{ form.q }}
-
-  <button type="submit">
-    <i class="icon icon-search"></i>
-    <span class="visuallyhidden">{% trans 'Search' %}</span>
-  </button>
-</form>
+    <button type="submit">
+      <i class="icon icon-search"></i>
```
MHLut, line 8: If this is a purely decorative icon, we should either make it a `<span>` or add `aria-hidden="true"` to it.

docs/templates/docs/search_form.html
```diff
@@ -1,11 +1,12 @@
...
+      <span class="visuallyhidden">{% trans 'Search' %}</span>
```
MHLut, line 9: We do not need "Search" as text as it is already in the label, so this span would have duplicate content.

The text could be "Submit" for clarity.

docs/templates/docs/search_results.html
```diff
@@ -11,6 +11,12 @@
 
 {% block body %}
   {% if query %}
+    <search class="filters" aria-label="{% trans 'Filter the current search results by documentation type' %}">
+      <a{% if not active_type %} class="active" aria-disabled="true" aria-current="page"{% else %} href="{% querystring type=None page=None %}"{% endif %}>{% trans "All" %}</a>
+      {% for type, type_title in document_types.items %}
```
MHLut, line 16: We could omit both the `class="active"` and `aria-disabled="true"` here, as:

1. We could omit the `href` attribute on active pages entirely. 
2. CSS does not need the active class; it could style using an attribute selector on `a[aria-current]` or `a:not(:link)` when omitting the href (or something similar).

docs/templates/docs/search_form.html
MHLut, line 2: Could we use `{% translate %}` instead of `{% trans %}`?

MHLut (changes requested):
I did a partial review, focussing on the HTML specifically.
Overall, I am happy we are getting improved search results!

docs/templates/docs/search_results.html
sarahboyce: Thank you! Updated

pauloxnet (approved):

bmispelon (commented):
This is looking amazing! I know I'll be using that feature a ton and I can't wait for it to be live.

I've left some non-essential suggestions, mostly around using an actual enum/Textchoices instead of a dict. It doesn't change much fundamentally but I think it makes sense here (I went for a `FULL_CAPS` name in singular, but I'm not sure if that's the best practice, maybe an `AlternateCase` name would be better? 🤔 ).

Also, I realize this is an annoying change to make and it brings no value code-wise, but I feel that "type" is a bit of a meaningless word (and it being a builtin function in python makes it annoying to use as a variable name). What do you think of "category", "topic" or maybe even "flavor" (I'll even look the other way if you spell it "flavour" 👀 )

docs/search.py
```diff
@@ -51,3 +52,10 @@
 
 START_SEL = "<mark>"
 STOP_SEL = "</mark>"
+
+DOCUMENT_TYPES = {
+    "ref": _("API Reference"),
+    "topics": _("Using Django"),
+    "howto": _("How-to guides"),
+    "releases": _("Release notes"),
+}
```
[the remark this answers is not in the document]
sarahboyce: I renamed this to `DocumentType`

PR #1936, opened 2025-02-07
Fixed issue reporting link in 404 template

adamzap:
This change directs users to report website issues on GitHub rather than on Trac. Trac is for issues about the Django web framework, not the Django website.


sarahboyce (approved):
⭐

pauloxnet (approved):

PR #1941, opened 2025-02-11
Removed only use of 600 `font-weight` CSS value

adamzap:
This change causes no visual difference on the site since this font weight is not being requested for Roboto from Google Fonts, but it technically reduces the amount of font variants the site needs.

The 600 value was added in https://github.com/django/djangoproject.com/commit/5d8bd41e9560d6b536a57216efa6b059e4f8642f.

This change can be tested with DevTools [here](https://docs.djangoproject.com/en/dev/topics/i18n/#term-internationalization).

Part of #1003.

### DIFF
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -86,7 +86,7 @@ dl {
 
     &.glossary {
         dt {
-            font-weight: 600;
+            font-weight: 700;
         }
     }
 }


pauloxnet (approved):

adamzap:
I found it by manually grepping and inspecting the site's CSS/SCSS. I would like to eventually have a CSS/SCSS linter in place, but I think we should do some cleanup and evaluation first. [`stylelint`](https://stylelint.io/) seems like a good choice here, and I tried it locally.

I do think that when we move off of Google Fonts, this problem will go away. I plan on using a separate file for each variant of Roboto like I did with Fira Mono. So if new code uses a font variant that we don't have, it won't look correct during development. Therefore I think we'll be safe. (Also, it will go away because I'm auditing all font variants in use 😉).

Issue #1649, opened 2024-10-07
Potential method for enhancing docs search results for important terms


jacklinke:
# Problem

Some terms are important for users to find in the Django documentation, but they are not in the title, slug, or table of contents for the document. As a result, these terms are not weighted as highly in the search results as they should be.

Example: Terms like "TextChoices" and "IntegerChoices" are important, but they are not in the title, slug, or table of contents for the [Model field reference](https://docs.djangoproject.com/en/5.1/ref/models/fields/) page. They are only mentioned in the body of the document.

# Background

The current full-text search implementation for the docs uses the `Document` model's `metadata` JSONField. This field is populated by the Sphinx-based documentation for the Django project.

Upon each new release, the [`update_docs`](https://github.com/django/djangoproject.com/blob/main/docs/management/commands/update_docs.py) management command runs `sphinx-build json` (among many other things) and invokes the `DocumentRelease` model's [`sync_to_db`](https://github.com/django/djangoproject.com/blob/main/docs/models.py#L156) method to import the json data into the Document model.

The `metadata` field contains various keys like `body`, `toc`, etc. But there is no provision for adding additional text to the `metadata` field and to the search vector to account for important terms that are in the `body` of the `Document` but not in the higher-weighted `title`, `slug`, or `toc` (table of contents).

We cannot simply add the additional text to a new key in the `metadata` field because the `Document` instances are overwritten with each new release. We need a way to store the additional text separately and add it to the `metadata` field when the `Document` instances are created or updated.

# A Potential Solution

I am proposing adding a `DocumentEnhancement` model to store additional text that can be added to the `metadata` field of the `Document` model. This additional text can be weighted at the same level as the `toc` field ("B") to allow for additional text to be added into the search vector when important terms are not in the title, slug, or table of contents.

- The `DocumentEnhancement` model has a one-to-one relationship with `Document`.
- The `DocumentRelease.sync_to_db` method is updated to read the additional text from the `DocumentEnhancement` model and add it to the `metadata` field.
- The `DOCUMENT_SEARCH_VECTOR` in `docs/search.py` is updated to include the `enhancements` key in the search vector.
- At any time, the `migrate_enhancements` management command can be run to migrate `DocumentEnhancement` data to the `metadata` field of the `Document` model instances.
- Optionally, we can improve the admin interface to allow for easy management of `DocumentEnhancement`.

# Implementation

**Note: These are not yet tested. Just jotted down this evening for input from the community.**

## docs/models.py

```python
[...]

Issue #1097, opened 2021-06-07
Search for "through" finds nothing


boxed:
https://docs.djangoproject.com/en/3.2/search/?q=through

This search should link at least these:

https://docs.djangoproject.com/en/3.2/topics/db/models/#extra-fields-on-many-to-many-relationships
https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.ManyToManyField.through

pauloxnet:
In English, the word "through" is a stopword and is ignored in the search against the English dictionary used in PostgreSQL.

From the [PostgreSQL documentation](https://www.postgresql.org/docs/13/textsearch-dictionaries.html#TEXTSEARCH-STOPWORDS):
> Stop words are words that are very common, appear in almost every document, and have no discrimination value. Therefore, they can be ignored in the context of full text searching. For example, every English text contains words like a and the, so it is useless to store them in an index.

The English word "through" is not a stopword in another dictionary for example the Italian dictionary, and in fact the search in this language shows results:

https://docs.djangoproject.com/it/3.2/search/?q=through

boxed:
I figured it might be something like that. Framework function names and stuff should bypass that logic somehow.

pauloxnet:
@boxed I don't think "through" is the only stopwords that matter in search.
Perhaps it would be useful to have a list of these words and then think of a way to ensure that they are not discarded.
Could you write a list of words not to be deleted starting from the official PostgreSQL stopwords list?
https://github.com/postgres/postgres/blob/master/src/backend/snowball/stopwords/english.stop

boxed:
Hm.. I don't know about a complete list. But certainly "where" is suspicious as it's a keyword in SQL. This becomes a bit tricky as "where" should probably just be searched when it's in a code block like `select` or similar. I think "now" is a bit doubtful it should be excluded too as it could be something you want to search for like `datetime.now` which I guess the current implementation just interprets as "datetime".

That's what I could find reading through this list. I think one could image a solution where the search is run and there are no hits, then it's re-run but ignoring stopwords. This would fix the worst case at least.

pauloxnet:
We can try to create a custom English dictionary without relevant words for Django.

boxed:
Considering this simple and limited scope change has seen no improvement in several years, I don't think broadening the scope of the issue is a good idea. 

[...]

PR #1942, opened 2025-02-11
Allowed search results for Django code terms which contain stop words.

sarahboyce:
Refs https://github.com/django/djangoproject.com/issues/1097 and https://github.com/django/djangoproject.com/issues/1649

The following is a bit of a proof of concept to one way we can allow search terms for code that is also a stop word
This avoids a custom stopwords file which makes us less dependent on keeping this up to date and "rejigging" things around

search for through

search for when

----------

Note that I will write tests once we are clear on what direction we want to go :+1:


docs/search.py
```diff
@@ -43,6 +43,11 @@
     SearchVector("title", weight="A", config=F("config"))
     + SearchVector(KeyTextTransform("slug", "metadata"), weight="A", config=F("config"))
     + SearchVector(KeyTextTransform("toc", "metadata"), weight="B", config=F("config"))
+    + SearchVector(
+        KeyTextTransform("code_references_search", "metadata"),
+        weight="B",
+        config="simple",
```
sarahboyce, line 49: As this contains only code, rather than "text", I don't see the point of having stop words be ignored (as "when", "through", "if" etc can still be exactly what you're after)

docs/models.py
```diff
@@ -249,7 +300,7 @@ def search(self, query_text, release):
         query_text = query_text.strip()
         if query_text:
             search_query = SearchQuery(
-                query_text, config=models.F("config"), search_type="websearch"
+                query_text, config="simple", search_type="websearch"
```
sarahboyce, line 303: I had to open this up to allow stop words here, these words are ignored in most cases

docs/models.py
```diff
@@ -213,6 +218,52 @@ def _clean_document_path(path):
     return path
 
 
+def _generate_code_references(body):
```
sarahboyce, line 221: I think even without the goal of bypassing the stop word limitation, I feel us adding this enrichment could be useful to us and give us precise links also

docs/stopwords/README.md
```diff
@@ -0,0 +1,38 @@
+# Instructions to create a new search dictionary
```
sarahboyce, line 1: I have never done this before so these instructions may not be very good
I would love it if we can create this custom search dictionary in our docker setup as well

pauloxnet: Great job. Still few time to review properly, but I had two ideas:

1. create a migration to create a custom English dictionary
2. write the list of words you removed from the original list of stop words and a command/pathc/other to remove those words from the final stop words list you uploaded

Issue #1868, opened 2025-01-09
Board minutes Business display order should match the order they are entered in


thibaudcolas:
In the [DSF board minutes](https://www.djangoproject.com/foundation/minutes/2025/jan/09/dsf-board-monthly-meeting/), items in the New business section are displayed in alphabetic order. It would be better if they were displayed in the order they’ve been added inside the Django admin.

Spotted by @knyghty and I while working on today’s minutes.

Satishmedarr:
Hi, I would like to work on this issue. Please assign it to me. I will ensure that the Board minutes Business display order matches the order in which they are entered. Thank you!

FarhanAliRaza:
I have opened this PR . Can someone please review it https://github.com/django/djangoproject.com/pull/1944

sabderemane:
I'm closing this issue in favour of the issue related to [the new location of the meeting minutes](https://github.com/django/djangoproject.com/issues/2079).

ulgens:
Because the issue wasn't completed, we decided that it's no longer required, I updated the resolution to "closed - not planned".

sabderemane:
Thanks @ulgens ! I can't choose the way to close from my phone 😔

PR #1944, opened 2025-02-12
Changed new business ordering to be  `id` based

FarhanAliRaza:
Fixes https://github.com/django/djangoproject.com/issues/1868


bmispelon (changes requested):
Hi and thanks for your contribution!

I think it would be better if we added an actual `created_on` field on the model and sorted by that, rather than relying on the id being sequential (which is not always guaranteed).

We'll also need a test for this new behavior.

FarhanAliRaza:
I am getting this error . I am running using docker. Command is running but getting this error.
```Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/asgiref/local.py", line 89, in _lock_storage
    asyncio.get_running_loop()
RuntimeError: no running event loop

During handling of the above exception, another exception occurred:

  File "/usr/local/lib/python3.12/site-packages/django/utils/connection.py", line 58, in __getitem__
    return getattr(self._connections, alias)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/asgiref/local.py", line 118, in __getattr__
    return getattr(storage, key)
           ^^^^^^^^^^^^^^^^^^^^^
AttributeError: '_thread._local' object has no attribute 'trac'

During handling of the above exception, another exception occurred:

  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/usr/src/app/manage.py", line 22, in <module>
    main()
  File "/usr/src/app/manage.py", line 18, in main
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.12/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.12/site-packages/django/core/management/__init__.py", line 436, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.12/site-packages/django/core/management/commands/test.py", line 24, in run_from_argv
    super().run_from_argv(argv)
  File "/usr/local/lib/python3.12/site-packages/django/core/management/base.py", line 413, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python3.12/site-packages/django/core/management/base.py", line 459, in execute
    output = self.handle(*args, **options)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/django/core/management/commands/test.py", line 63, in handle
    failures = test_runner.run_tests(test_labels)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/django/test/runner.py", line 1072, in run_tests
    self.run_checks(databases)
  File "/usr/local/lib/python3.12/site-packages/django/test/runner.py", line 994, in run_checks
    call_command("check", verbosity=self.verbosity, databases=databases)
  File "/usr/local/lib/python3.12/site-packages/django/core/management/__init__.py", line 194, in call_command
    return command.execute(*args, **defaults)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/django/core/management/base.py", line 459, in execute
    output = self.handle(*args, **options)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/django/core/management/commands/check.py", line 81, in handle
    self.check(
  File "/usr/local/lib/python3.12/site-packages/django/core/management/base.py", line 486, in check
    all_issues = checks.run_checks(
                 ^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/django/core/checks/registry.py", line 88, in run_checks
    new_errors = check(app_configs=app_configs, databases=databases)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/django/core/checks/database.py", line 12, in check_database_backends
    conn = connections[alias]
           ~~~~~~~~~~~^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/django/utils/connection.py", line 61, in __getitem__
    raise self.exception_class(f"The connection '{alias}' doesn't exist.")
django.utils.connection.ConnectionDoesNotExist: The connection 'trac' doesn't exist.```

**Comment by adamzap** [comment]
@FarhanAliRaza Could you address the comments on this PR if you're still interested in working on it? Otherwise it will be closed in a week. Thanks!

**Comment by FarhanAliRaza** [comment]
> @FarhanAliRaza Could you address the comments on this PR if you're still interested in working on it? Otherwise it will be closed in a week. Thanks!
I have added `created_on`. What else is needed from this?

File: foundation/models.py [hunk 1]
```diff
@@ -207,6 +207,7 @@ class Business(models.Model):
     meeting = models.ForeignKey(
         Meeting, related_name="business", on_delete=models.CASCADE
     )
+    created_on = models.DateTimeField(auto_now_add=True)
```
ulgens, line 210: We have a `created_at` on `LocalDjangoCommunity`, using the same arguments. I think it's better to keep the same naming pattern. It's not a written rule or sth but in my experience,`_at` suffix communicates that the field is a timestamp, which helps with the overall cognitive load.

FarhanAliRaza: updated

ulgens (approved):

ulgens:
@bmispelon Would you like to review again? Didn't want to proceed further without your feedback.

bmispelon (changes requested):
Hi again 😁 

Thanks for adding the `created_at` field, I think that's a better approach than sorting on `id` (and that's a great use-case for `auto_now_add` 👍🏻 )

There's a few things I think could be improved on the PR:

1) The migration for existing data. Right now every existing `Business` will get the current date stored in its `created_at` which seems less than ideal. It should at least be based on the meeting's date. And to take it even further, the previous order of items should be respected so I think the sorting should be `"created_at", "title"`.

2) Instead of testing on `response.content`, I'd like to see `assertQuerySetEqual` with `response.context["new_business"]`. When doing that, you'll find out that the tests are currently broken because the value used for `business_type` is invalid (it should be `"new"` with a lowercase N).

3) While we're at it I think the logic should apply to `ongoing_business` too, and so I would suggest changing the ordering directly on the model's `Meta.ordering`.


Having said all of that, I've heard rumors that the DSF board might be moving to a new system for keeping minutes, so I'm not sure if it's worth investing a lot of time on this change. Could we get some information about this @sabderemane ?

sabderemane:
Hey @bmispelon ! 
Sorry this one went off my radar. 

Yes, our Secretary confirmed and I asked for [an issue on the website](https://github.com/django/djangoproject.com/issues/2079) related to this point. 

Unfortunately, I missed the conversation related to this transition. Now, we need to find a way to keep providing the 
[meeting minutes which are now defined in a repository as markdown](https://github.com/django/dsf-minutes/tree/main/2025).

Having said that, I believe this PR is not needed anymore but we need to find a way to keep the meeting minutes available for people who use the RSS feed and in [the website in the side bar](https://www.djangoproject.com/foundation/).

 I'm sorry for the inconvenience @FarhanAliRaza , instead would you be interested to work on this? An issue will be created soon.

ulgens:
@bmispelon @sabderemane As a more generalized approach, how do we feel about adding timestamp fields to each model (created_at - updated_at)? They are useful for governance and history keeping.

Issue #1464, opened 2024-01-13
Help text for foundation Meeting model fields


thibaudcolas:
Since I only use the Meeting model’s form once a month, it’d really help to have instructions explaining how to use the different fields directly on the form, rather than having to document this separately.

Here’s the help text I’m currently thinking of:

- TItle & slug: "Peep the default as-is"
- Board attendees: "Make sure to choose the correct year"
- Business Body: "Supports reStructuredText. See [supported formatting](https://github.com/django/djangoproject.com/issues/1463)"
- Action items: help content for whole section – "aka todos. Combine the last meeting’s remaining action items, and add any new ones"

And a prominent note somewhere towards the top of the form:

> [!NOTE] 
> Meeting notes go live and on RSS on Save, make sure to complete all relevant fields.

ronnzw:
I totally agree with this, instructions should be away from the form.

thibaudcolas:
cc our 2025 Secretary @knyghty. I’ve unassigned myself as I’ve really not been able to find time for this in 2024 and have to be realistic for the year ahead too

PR #1945, opened 2025-02-12
Help text for foundation Meeting model fields 

FarhanAliRaza:
fixes https://github.com/django/djangoproject.com/issues/1464
image
image
image
image

### DIFF
--- a/foundation/templates/admin/foundation/meeting/change_form.html
+++ b/foundation/templates/admin/foundation/meeting/change_form.html
@@ -1,0 +1,25 @@
+{% extends "admin/change_form.html" %} {% load i18n %} \ {% block extrastyle %}
+  {{ block.super }}
+  <style>
+    .meeting-warning {
+      background: #fff3cd;
+      border: 2px solid #ffeeba;
+      border-radius: 4px;
+      color: #856404;
+      margin: 1em 0;
+      padding: 1em;
+      font-size: 1.1em;
+      font-weight: 500;
+    }
+    .meeting-warning::before {
+      content: "⚠️ ";
+    }
+  </style>
+{% endblock %} {% block content %}
+  <div class="messagelist">
+    <div class="meeting-warning">
+      Note: Meeting notes go live and on RSS on Save, make sure to complete all
+      relevant fields.
+    </div>
+  </div>
+  {{ block.super }} {% endblock %}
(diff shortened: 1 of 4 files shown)


bmispelon:
Hi and thanks for opening this PR! 🎸 

I've tagged our current secretary as a reviewer to get an opinion on the wording since those changes are mostly for their benefit I believe.

Note also that the automated tests are currently failing, it seems it's because you didn't include the missing migration files for the `foundation` app.

knyghty:
The wording seems alright to me.

One thing, I'm not too sure if https://www.djangoproject.com/styleguide/content/ is kept up to date, but Thibaud seemed to think it was good enough so it's probably better than nothing.

I'll defer to others but I'm also not sure how we feel about having style elements in HTML (if we hope to use CSP some day) or having colours without using CSS variables.

foundation/templates/admin/foundation/meeting/change_form.html
```diff
@@ -0,0 +1,25 @@
...
+      border-radius: 4px;
+      color: #856404;
+      margin: 1em 0;
+      padding: 1em;
+      font-size: 1.1em;
+      font-weight: 500;
+    }
+    .meeting-warning::before {
+      content: "⚠️ ";
+    }
```
sabderemane, line 16: Could we have this in the [dedicated style folder](https://github.com/django/djangoproject.com/blob/main/djangoproject/scss) instead? or this will not work?

foundation/templates/admin/foundation/meeting/change_form.html
```diff
@@ -0,0 +1,25 @@
+{% extends "admin/change_form.html" %} {% load i18n %} \ {% block extrastyle %}
+  {{ block.super }}
+  <style>
+    .meeting-warning {
+      background: #fff3cd;
+      border: 2px solid #ffeeba;
```
sabderemane, line 6: How does it render in light theme?

sabderemane (commented):
Looks good, thank you Farhan! I added some questions related to the appearance and style.

foundation/templates/admin/foundation/meeting/change_form.html
FarhanAliRaza: image

foundation/templates/admin/foundation/meeting/change_form.html
FarhanAliRaza: That will not work because I am overriding the admin template form.For it to work, I think I will have to add compiled SCSS to the base admin template. That may break other styles as well because now, CSS is compiled to one big output.css file.

PR #1947, opened 2025-02-13
Added code references to search results.

sarahboyce:
#### Description

Recently, we added a scroll to text fragment to improve the search result navigation experience.
This takes the first line that was found in the body and scrolls you to this line. 

This PR adds references to our Python class/class method/function definitions which will appear as a sub-item of the page search result if they match the search.
This means that when someone searches for "select" they have choice as to what to navigate to. They can chose the "QuerySet.select_related" or "QuerySet.select_for_update". The anchors of these are also robust and so you will navigate to the definition of these (the scroll for text logic is a little fragile due to the html stripping logic of full text search).

search for "select"

This "choice" is even more useful when searching for something like "field":

search for "field"

------

#### Issues aiming to fix

Ideally this along with the previous highlight text change fixes https://github.com/django/djangoproject.com/issues/1649 and fixes https://github.com/django/djangoproject.com/issues/1734

#### Review notes

Note that I have added some folks as possible reviewers.
I am not 100% sure the review process given the new website WG but would appreciate feedback as I'd love to land the PR :heart_hands: 

This PR has 2 commits, the first commit is a refactor commit to split the test file into multiple test files, the second is the feature.

I originally used regex to full out these code references. @bmispelon rightly pointed out this is brittle. I have since written a custom Sphinx builder in order to use Sphinx's own functionality for getting the python objects.

### DIFF
--- a/docs/builder.py
+++ b/docs/builder.py
@@ -1,0 +1,76 @@
+from dataclasses import dataclass
+from functools import cached_property
+
+from sphinxcontrib.serializinghtml import JSONHTMLBuilder
+
+IGNORED_DOMAIN_TYPES = {"module"}
+
+
+@dataclass
+class DomainObject:
+    """
+    A wrapper around sphinx's domain object descriptions
+    https://www.sphinx-doc.org/en/master/extdev/domainapi.html#sphinx.domains.Domain.get_objects
+    """
+
+    name: str
+    dispname: str
+    type: str
+    docname: str
+    anchor: str
+    priority: int
+
+    @property
+    def short_name(self) -> str:
+        """
+        Returns a shortened version of the object's name.
+
+        - If the second-to-last part of the name starts with an uppercase letter
+          (indicating a class method or class attribute), returns the last two parts.
+        - Otherwise, returns only the last part (likely a function, or class).
+
+        Examples:
+        - "django.db.models.query.QuerySet.select_related" → "QuerySet.select_related"
+        - "django.db.models.query.QuerySet" → "QuerySet"
+        - "django.template.context_processors.static" → "static"
+        """
+        parts = self.name.split(".")
+
+        if len(parts) < 2:
+            return self.name
+
+        last, second_last = parts[-1], parts[-2]
+        return f"{second_last}.{last}" if second_last[0].isupper() else last
+
+
+class PythonObjectsJSONHTMLBuilder(JSONHTMLBuilder):
+    @cached_property
+    def domain_objects(self):
+        domain = self.env.get_domain("py")
+        return [
+            DomainObject(*item)
+            for item in domain.get_objects()
+            if item[2] not in IGNORED_DOMAIN_TYPES  # item[2] is 'type'.
+        ]
+
+    def get_doc_context(self, docname, body, metatags):
+        out_dict = super().get_doc_context(docname, body, metatags)
+        python_objects = self.get_python_objects(docname)
+        out_dict["python_objects"] = python_objects
+        out_dict["python_objects_search"] = " ".join(
+            # Keeps the code suffix to improve the search results for terms such as
+            # "select" for QuerySet.select_related.
+            [key.split(".")[-1] for key in python_objects.keys()]
+        )
+        return out_dict
+
+    def get_python_objects(self, docname):
+        return {
+            obj.short_name: obj.name
+            for obj in self.domain_objects
+            if obj.docname == docname
+        }
+
+
+def setup(app):
+    app.add_builder(PythonObjectsJSONHTMLBuilder, override=True)
--- a/docs/management/commands/update_docs.py
+++ b/docs/management/commands/update_docs.py
@@ -4,6 +4,7 @@ app.
 """
 
 import json
+import multiprocessing
 import os
 import shutil
 import subprocess
@@ -16,6 +17,9 @@ from pathlib import Path
 from django.conf import settings
 from django.core.management import BaseCommand, call_command
 from django.utils.translation import to_locale
+from sphinx.application import Sphinx
+from sphinx.config import Config
+from sphinx.errors import SphinxError
 
 from ...models import DocumentRelease
 
@@ -175,29 +179,30 @@ class Command(BaseCommand):
 
             if self.verbosity >= 2:
                 self.stdout.write(f"  building {builder} ({source_dir} -> {build_dir})")
+            # Retrieve the extensions from the conf.py so we can append to them.
+            conf_extensions = Config.read(source_dir.resolve()).extensions
+            extensions = [*conf_extensions, "docs.builder"]
             try:
-                # Translated docs builds generate a lot of warnings, so send
-                # stderr to stdout to be logged (rather than generating an
-                # email)
-                subprocess.check_call(
-                    [
-                        "sphinx-build",
-                        "-b",
-                        builder,
-                        "-D",
-                        "language=%s" % to_locale(release.lang),
-                        "-j",
-                        "auto",
-                        "-Q" if self.verbosity == 0 else "-q",
-                        str(source_dir),  # Source file directory
-                        str(build_dir),  # Destination directory
-                    ],
-                    stderr=sys.stdout,
-                )
-            except subprocess.CalledProcessError:
+                Sphinx(
+                    srcdir=source_dir,
+                    confdir=source_dir,
+                    outdir=build_dir,
+                    doctreedir=build_dir.joinpath(".doctrees"),
+                    buildername=builder,
+                    # Translated docs builds generate a lot of warnings, so send
+                    # stderr to stdout to be logged (rather than generating an email)
+                    warning=sys.stdout,
+                    parallel=multiprocessing.cpu_count(),
+                    verbosity=self.verbosity,
+                    confoverrides={
+                        "language": to_locale(release.lang),
+                        "extensions": extensions,
+                    },
+                ).build()
+            except SphinxError as e:
                 self.stderr.write(
-                    "sphinx-build returned an error (release %s, builder %s)"
-                    % (release, builder)
+                    "sphinx-build returned an error (release %s, builder %s): %s"
+                    % (release, builder, str(e))
                 )
                 return
 
--- a/docs/templates/docs/search_results.html
+++ b/docs/templates/docs/search_results.html
@@ -48,6 +48,21 @@
             {% if result.highlight %}
               …&nbsp;{{ result.highlight|cut:"¶"|safe }}&nbsp;…
             {% endif %}
+            {% code_links result.searched_python_objects result.python_objects as result_code_links %}
+            {% if result_code_links %}
+              <ul class="code-links">
+                {% for name, value in result_code_links.items %}
+                  <li>
+                    <a href="{% url 'document-detail' lang=result.release.lang version=result.release.version url=result.path host 'docs' %}#{{ value.full_path }}">
+                      <div>
+                        <code>{{ name }}</code>
+                        {% if value.module_path %}<div class="meta">{{ value.module_path }}</div>{% endif %}
+                      </div>
+                    </a>
+                  </li>
+                {% endfor %}
+              </ul>
+            {% endif %}
           </dd>
         {% endfor %}
       </dl>
--- a/docs/templatetags/docs.py
+++ b/docs/templatetags/docs.py
@@ -12,7 +12,7 @@ from pygments.lexers import get_lexer_by_name
 from ..forms import DocSearchForm
 from ..models import DocumentRelease
 from ..search import START_SEL, STOP_SEL
-from ..utils import get_doc_path, get_doc_root
+from ..utils import get_doc_path, get_doc_root, get_module_path
 
 register = template.Library()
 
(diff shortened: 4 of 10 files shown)


alexgmin:
Is there a case in which it makes sense to have the first line of the code reference?
Based on my searches it always seems redundant with the second line
image

Everything else looks good to me.

sarahboyce:
> Is there a case in which it makes sense to have the first line of the code reference?

It's true that it's duplicate information.
I personally thought reading `QuerySet.select_related` is nicer than reading `django.db.models.query.QuerySet.select_related`, but still kept the full path as it's interesting/useful 🤷

sarahboyce:
I could make it like:
> `QuerySet.select_related`
> `django.db.models.query`

alexgmin:
> It's true that it's duplicate information. I personally thought reading `QuerySet.select_related` is nicer than reading `django.db.models.query.QuerySet.select_related`, but still kept the full path as it's interesting/useful 🤷
It is, and I would prefer to keep the short version, but I don't know if there's a case in which the short version is not unique.

> I could make it like:
> 
> > `QuerySet.select_related`
> > `django.db.models.query`
This could also work better than the current one if only the shortened version is not viable due to uniqueness.

sarahboyce:
> It is, and I would prefer to keep the short version, but I don't know if there's a case in which the short version is not unique.

`Field` for example, `django.db.models.fields.Field` and `django.forms.fields.Field`

bmispelon (commented):
Nice work ✨ 

In terms of functionality and search experience I think this is a big improvement. I'm worried that the actual implementation might be a bit brittle and could break if sphinx was to change the HTML it outputs (and if it did break, none of our tests would catch it).

But overall this is very good. I've left a few minor comments, nothing major.

docs/management/commands/update_docs.py
```diff
@@ -172,28 +172,33 @@ def build_doc_release(self, release, force=False):
...
-                        "-j",
-                        "auto",
-                        "-Q" if self.verbosity == 0 else "-q",
-                        str(source_dir),  # Source file directory
-                        str(build_dir),  # Destination directory
-                    ],
-                    stderr=sys.stdout,
-                )
+                app.build()
             except subprocess.CalledProcessError:
```
sarahboyce, line 202: TODO: update error catching here (not sure what to)

docs/builder.py
```diff
@@ -0,0 +1,39 @@
+from functools import cached_property
+
+from sphinxcontrib.serializinghtml import JSONHTMLBuilder
+
+
+class JSONTOCHTMLBuilder(JSONHTMLBuilder):
```
sarahboyce, line 6: TODO: write a test for the custom builder

pauloxnet:
I'm sorry, but I haven't found time to review this interesting PR.😞

Could you try explaining in the description what's the goal and the design of this PR. I'm not being pedantic here , but I'm not sure I understood correctly what you want to achieve. 😊

sarahboyce:
> Could you try explaining in the description what's the goal and the design of this PR

Updated the description! Sorry it was lacking a lot of context

I believe this is ready for another round of review :heart:

bmispelon:
Could we merge the test module restructuring first? It seems like a good change, and it would make reviewing this easier.

docs/tests/test_builder.py
```diff
@@ -0,0 +1,63 @@
+from unittest.mock import Mock, patch
+
+from django.test import SimpleTestCase
+
+from ..builder import PythonObjectsJSONHTMLBuilder
+
+
+class TestPythonObjectsJSONHTMLBuilder(SimpleTestCase):
+    def setUp(self):
```
pauloxnet, line 9: Can we move all variables directly to the class?

docs/tests/test_utils.py
```diff
@@ -0,0 +1,40 @@
+import os
+from pathlib import Path
+
+from django.test import TestCase
+
+from ..utils import get_doc_path, sanitize_for_trigram
+
+
+class TestUtils(TestCase):
```
pauloxnet, line 9: Can we use a Simple Test case?

docs/tests/test_views.py
```diff
@@ -0,0 +1,188 @@
...
+from django_hosts.resolvers import reverse as reverse_with_host
+
+from djangoproject.urls import www as www_urls
+from releases.models import Release
+
+from ..models import Document, DocumentRelease
+from ..sitemaps import DocsSitemap
+
+
+class RedirectsTests(TestCase):
```
pauloxnet, line 15: Can we use Simple Test Case?

docs/tests/test_views.py
```diff
@@ -0,0 +1,188 @@
...
+        super().tearDownClass()
+
+    def test_empty_get(self):
+        response = self.client.get(
+            "/en/dev/search/", headers={"host": "docs.djangoproject.localhost:8000"}
+        )
+        self.assertEqual(response.status_code, 200)
+
+    def test_code_links(self):
+        Document.objects.create(
```
pauloxnet, line 64: Can we use a bulk creation for these 2 documents?

pauloxnet (commented):
I added some comments from a first review (from my phone).

sarahboyce:
> Could we merge the test module restructuring first? It seems like a good change, and it would make reviewing this easier.

https://github.com/django/djangoproject.com/pull/1965

docs/builder.py
```diff
@@ -0,0 +1,39 @@
...
+from sphinxcontrib.serializinghtml import JSONHTMLBuilder
+
+
+class PythonObjectsJSONHTMLBuilder(JSONHTMLBuilder):
+    name = "pyjson"
+
+    @cached_property
+    def domain_objects(self):
+        domain = self.env.get_domain("py")
+        return [item for item in domain.get_objects() if item[2] != "module"]
```
bmispelon, line 12: It would be a bit more readable if we could wrap the items returned by `get_objects()` into something like a dataclass:
```python
from dataclasses import dataclass

@dataclass
class DomainObject:
    """
    A wrapper around sphinx's domain object descriptions
    https://www.sphinx-doc.org/en/master/extdev/domainapi.html#sphinx.domains.Domain.get_objects
    """
    name: str
    dispname: str
    type: str
    docname: str
    anchor: str
    priority: int
```

Oh, and also I'd create either a class or module-level constant `IGNORED_DOMAIN_TYPES = {"module"}` to make it a bit clearer why `"module"` is left out.

Then this all becomes:
```suggestion
        return [item for item in map(DomainObject, domain.get_objects() if item.type not in IGNORED_DOMAIN_TYPES]
```

docs/builder.py
```diff
@@ -0,0 +1,39 @@
...
+            # "select" for QuerySet.select_related.
+            [key.split(".")[-1] for key in python_objects.keys()]
+        )
+        return out_dict
+
+    def get_python_objects(self, docname):
+        entries = {}
+        for name, _, _, obj_docname, _, _ in self.domain_objects:
+            if obj_docname == docname:
+                code_path = name.split(".")[-2:]
```
bmispelon, line 29: `code_path` could be a property on the `DomainObject` dataclass, and so could whatever `code_path[0][0]` is testing (naming help would also help explain what it's doing, I still don't really understand what's going on there 😁 )

docs/management/commands/update_docs.py
```diff
@@ -176,28 +178,31 @@ def build_doc_release(self, release, force=False):
...
-                        "auto",
-                        "-Q" if self.verbosity == 0 else "-q",
-                        str(source_dir),  # Source file directory
-                        str(build_dir),  # Destination directory
-                    ],
-                    stderr=sys.stdout,
-                )
-            except subprocess.CalledProcessError:
+                Sphinx(
+                    srcdir=str(source_dir),
```
bmispelon, line 182: I believe (looking at the type hints in Sphinx's source code) that path objects are also accepted in all the arguments here, so no need for casting them to `str`.

docs/management/commands/update_docs.py
```diff
@@ -176,28 +178,31 @@ def build_doc_release(self, release, force=False):
...
-                        "-j",
-                        "auto",
-                        "-Q" if self.verbosity == 0 else "-q",
-                        str(source_dir),  # Source file directory
-                        str(build_dir),  # Destination directory
-                    ],
-                    stderr=sys.stdout,
-                )
-            except subprocess.CalledProcessError:
+                Sphinx(
```
bmispelon, line 181: Great call moving away from `subprocess.check_call()`! 👏🏻

docs/management/commands/update_docs.py
```diff
@@ -176,28 +178,31 @@ def build_doc_release(self, release, force=False):
...
+                    srcdir=str(source_dir),
+                    confdir=str(source_dir),
+                    outdir=str(build_dir),
+                    doctreedir=str(build_dir / ".doctrees"),
+                    buildername=builder,
+                    confoverrides={
+                        "language": to_locale(release.lang),
+                        "extensions": [
```
bmispelon, line 189: Can you clarify where this list of extensions come from? I don't see it in the original version of the code. Are they pulled from a config file somewhere?

docs/management/commands/update_docs.py
```diff
@@ -176,28 +178,31 @@ def build_doc_release(self, release, force=False):
...
+                            "djangodocs",
+                            "sphinx.ext.extlinks",
+                            "sphinx.ext.intersphinx",
+                            "sphinx.ext.autosectionlabel",
+                            "sphinx.ext.linkcode",
+                            "docs.builder",
+                        ],
+                    },
+                    # Translated docs builds generate a lot of warnings, so send
+                    # stderr to stdout to be logged (rather than generating an email)
```
bmispelon, line 199: Oooh, so that's why the `stderr=sys.stdout` was there in the first place! That comment is gold! 💯

docs/templates/docs/search_results.html
```diff
@@ -48,6 +48,19 @@ <h2 class="result-title">
             {% if result.highlight %}
               …&nbsp;{{ result.highlight|cut:"¶"|safe }}&nbsp;…
             {% endif %}
+            {% code_links result.searched_python_objects result.python_objects as result_code_links %}
+            {% if result_code_links %}
+              <div class="code-links">
```
bmispelon, line 53: Semantically, shouldn't this be a `<ul>` instead of a `<div>`?

docs/templatetags/docs.py
```diff
@@ -121,3 +121,29 @@ def generate_scroll_to_text_fragment(highlighted_text):
     # Due to Python code such as timezone.now(), remove the space after a bracket.
     single_spaced = re.sub(r"([(\[])\s", r"\1", single_spaced)
     return f"#:~:text={quote(single_spaced)}"
+
+
+@register.simple_tag(name="code_links")
```
bmispelon, line 126: I think the `name` is not necessary here since the function is already called `code_links`

docs/templatetags/docs.py
```diff
@@ -121,3 +121,29 @@ def generate_scroll_to_text_fragment(highlighted_text):
...
+def code_links(searched_python_objects, python_objects):
```
bmispelon, line 127: Can we have a docstring please? 😁

docs/utils.py
```diff
@@ -57,3 +57,9 @@ def sanitize_for_trigram(text):
     text = unicodedata.normalize("NFKD", text)
     text = re.sub(r"[^\w\s]", "", text, flags=re.UNICODE)
     return " ".join(text.split())
+
+
+def get_module_path(name, full_path):
+    if len(name) >= len(full_path):
+        return None
+    return full_path[: -len(name) - 1]
```
bmispelon, line 65: Isn't that the same as `full_path.removesuffix(f".{name}")`? (might be me, but I tend to be really bad at understanding slices by just reading them, I always have to try them out on concrete data)

docs/utils.py
```diff
@@ -57,3 +57,9 @@ def sanitize_for_trigram(text):
     text = unicodedata.normalize("NFKD", text)
     text = re.sub(r"[^\w\s]", "", text, flags=re.UNICODE)
     return " ".join(text.split())
+
+
+def get_module_path(name, full_path):
```
bmispelon, line 62: Can we get a docstring here too? Maybe with an example input/output?

bmispelon (commented):
This is shaping up to be amazing, I can't wait for it to be running in production.

It looks great overall, I haven't found anything that seems wrong. I've left a few ideas for improvements, mostly around readability and trying to name things in a way that helps understand what's going on. Don't hesitate to push back on my suggestions if you think they don't make sense to you: at this point you're the domain expert here.


At the risk of repeating myself: amazing work! ✨

bmispelon:
> > Could we merge the test module restructuring first? It seems like a good change, and it would make reviewing this easier.
> 
> #1965

I took the liberty of merging that PR and rebasing + force-pushing this one. It made my reviewing easier and I hope I haven't messed up your local branch too hard 😅

docs/builder.py
sarahboyce: When I applied this exactly, I got this error:
```
    for item in map(DomainObject, domain.get_objects())
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: DomainObject.__init__() missing 5 required positional arguments: 'dispname', 'type', 'docname', 'anchor', and 'priority'
```
I have applied something similar

docs/builder.py
sarahboyce: I created a short_name property :+1:

docs/management/commands/update_docs.py
sarahboyce: I've added a comment. It comes from the the conf defined in the Django repo. I'm not sure if there is a better way to keep these aligned

bmispelon: Testing locally (only with `sphinx`, I don't have a working setup for building the docs) it seems that conf.py is loaded when you specify the `confdir`, so that should not be needed.

https://github.com/sphinx-doc/sphinx/blob/e7ca2d8eaf047ddb7b860cc16e67c7faf75317ee/sphinx/application.py#L266

bmispelon (commented):
I think the `confoverrides["extensions"]` is not needed (see comment), otherwise I think this is ready to ship as far as I'm concerned 👍🏻

docs/management/commands/update_docs.py
```diff
@@ -175,29 +177,32 @@ def build_doc_release(self, release, force=False):
...
-                    stderr=sys.stdout,
-                )
-            except subprocess.CalledProcessError:
+                # stderr to stdout to be logged (rather than generating an email)
+                "warning": sys.stdout,
+            }
+            # Initialize a temporary Sphinx app to load existing config values
+            # from confdir.
+            app = Sphinx(**sphinx_init)
+            extensions = [*app.config.extensions, "docs.builder"]
```
bmispelon, line 193: Ah, I hadn't realized this was about adding an item to the list of extensions! I've been pouring over the sphinx source code to see if there was a way to extend the config file's extension list rather than override it but there doesn't seem to be anything clean.
I'd suggest using the `Config` class (`from sphinx.config import Config`) to avoid all the extra work that `Sphinx.__init__()` is doing. Something like:

```suggestion
            # Retrieve the extensions from the conf.py so we can append to them
            conf_extensions = Config.read(source_dir).extensions
            extensions = [*conf_extensions, "docs.builder"]
```

docs/management/commands/update_docs.py
```diff
@@ -175,29 +177,32 @@ def build_doc_release(self, release, force=False):
 
             if self.verbosity >= 2:
                 self.stdout.write(f"  building {builder} ({source_dir} -> {build_dir})")
-            try:
+            sphinx_init = {
+                "srcdir": source_dir,
+                "confdir": source_dir,
+                "outdir": build_dir,
+                "doctreedir": build_dir.joinpath(".doctrees"),
+                "buildername": builder,
```
bmispelon, line 185: I just realized that the `-j auto` from the previous version was not carried over. I think this should be equivalent:
```suggestion
                "buildername": builder,
                "parallel": multiprocessing.cpu_count(),
```

bmispelon (commented):
The double instantiating of `Sphinx()` was clever! I've made an alternate proposal that uses `sphinx.config.Config()` rather than a whole `Sphinx` application.

I also noticed that the multi-cpu processing had been lost, so I left a suggestion for how to re-add it. I think the logic around the conditional verbosity has also not been carried over, but I don't think that's a huge deal.

We're almost there! 😁

bmispelon (approved):
Fantastic work!

From my perspective it's ready to deploy, but I'll probably wait from an official 👍🏻  from the website WG before I do so.

alexgmin (approved):

PR #1962, opened 2025-02-25
Allowed '/' key as alternate shortcut for search box

### LINKED ISSUE #1946: Restore `/` shortcut for searchbox?
#1112 / #1236 changed the keyboard shortcut for the docs search field from <kbd>/</kbd> to <kbd>Ctrl+K</kbd> or <kbd>Cmd-K</kbd>. There were several good arguments for adding the Ctrl/Cmd-K hotkeys.

Would it make sense to *also* allow <kbd>/</kbd>? At least three important sites in the Django ecosystem *only* work with <kbd>/</kbd>, and I keep getting tripped up on which keys work where. Also, it seems like the other sites cited in #1112 actually allow either shortcut.

| Site | / | Ctrl/Cmd-K |
|--------|--------|--------|
| forum.djangoproject.com | ✔  | - |
| docs.python.org | ✔ | - |
| GitHub | ✔ | - |
| ReadTheDocs ([example](https://channels.readthedocs.io/)) | ✔ | - |
| MDN | ✔ | ✔ |
| tailwindcss.com | ✔ | ✔ |
| getbootstrap.com | - | - | 

(Bootstrap was mentioned in the earlier issue, but since seems to have changed to <kbd>Ctrl+/</kbd>, only.)

[keywords: slash key, solidus, Control-K, Command-K]

medmunds:
Added the <kbd>/</kbd> key as a secondary shortcut to focus the search input, for consistency with forum.djangoproject.com, github.com, docs.python.org, and most ReadTheDocs sites.

(Ctrl+K and Command-K are also still supported.)

Fixes #1946.

### DIFF
--- a/djangoproject/static/js/djangoproject.js
+++ b/djangoproject/static/js/djangoproject.js
@@ -82,7 +82,8 @@ document.querySelector('.menu-button').addEventListener('click', function () {
 
 // Focus, select, and scroll to search input when key combination is pressed
 window.addEventListener('keydown', function (e) {
-  if (!(e.metaKey || e.ctrlKey) || e.key !== 'k') {
+  const is_ctrl_k = (e.metaKey || e.ctrlKey) && e.key === 'k';
+  if (!(is_ctrl_k || e.key === '/')) {
     return;
   }
 


djangoproject/static/js/djangoproject.js
```diff
@@ -82,7 +82,9 @@ document.querySelector('.menu-button').addEventListener('click', function () {
 
 // Focus, select, and scroll to search input when key combination is pressed
 window.addEventListener('keydown', function (e) {
-  if (!(e.metaKey || e.ctrlKey) || e.key !== 'k') {
+  const isCtrlK = (e.metaKey || e.ctrlKey) && e.key === 'k';
+  const isSlash = e.key === '/';
+  if (!(isCtrlK || isSlash)) {
```
adamzap, line 87: I defer to the other maintainers on whether or not this change should be merged, but could you modify your code to maintain the style of the file? I think we can also factor out the `isSlash` variable:

```suggestion
  const is_ctrl_k = (e.metaKey || e.ctrlKey) && e.key === 'k';
  
  if (!(is_ctrl_k || e.key === '/')) {
```

Thanks!

bmispelon:
This seems like a reasonable change, thanks for taking the time to work on this! 🎸

Issue #1956, opened 2025-02-22
Use <nav> in the docs page for table of contents, breadcrumbs (prev/next) etc


sarahboyce:
https://github.com/django/djangoproject.com/blob/main/docs/templates/docs/doc.html

Using `<nav>` for items such as the table of contents improves the experience for screenreader users. The items should be labeled correctly (associated to the labels)

The language and docs version picker may also want to be in a `<nav>`

[This comment](https://github.com/django/djangoproject.com/issues/1411#issuecomment-2676972984) has an outline of what I have in mind

kedar49:
can I work on this issue ?

bmispelon:
Had to revert the commit, so this is still open

bmispelon:
Fix reapplied in 1ee5166d50806b9da29f64c440c87beec5b6516d

Issue #1411, opened 2023-09-29
MAJOR - Heading level 1 should be used for the title of the page


sabderemane:
> Major: Certain users are very likely to have difficulty perceiving or operating this.

It visually most prominent heading, ideally not for invisible content ("Django" on homepage). Ideally each page would only have a single h1. ([07:04](https://www.youtube.com/watch?v=dAiMPTOMh1U&t=424s), [SC 1.3.1 Info and Relationships](https://www.w3.org/TR/WCAG22/#info-and-relationships)).

specialorange:
https://github.com/django/djangoproject.com/pull/1419

sarahboyce:
I would propose something like the following:
![docs page with suggested semantic html sections](https://github.com/user-attachments/assets/56055e6b-e77d-473e-90ab-a83841e354c1)

- the top bar is the header `<header>` 
- in the top bar we have navigation `<nav>`
- then it is `<main>` content
- the green "banner" is a `<section>` this will contain a `<h2>` as the information here is often not the page title but rather repeating content (as the title of the selected navigation of the site)
- the docs is a section case where is contains `<search>`
- we then have an `<article> for the docs page
- this has a `<h1>` with the docs title (note that the docs are generated from Sphinx and will always start with a `<h1>` I recommend we don't fight this)
- we already have the side section which has `role="complementary"`, in this we should have some `<nav>` with titles when appropriate

I recommend we roughly adopt this for every page
This means the banner has a h2 and then underneath is the h1
There are 2 pages which might struggle with this currently:
- https://www.djangoproject.com/weblog/
- https://www.djangoproject.com/community/

A quick win would be to add a title to these such as "Recent news" and "Community".
If we want it to look exactly the same, we might be able to have those pages use `<h1>` in the banner but the structure would be slightly inconsistent compared to other pages

bmispelon:
Had to revert the commit, so this is still open

bmispelon:
Fix reapplied in 1ee5166d50806b9da29f64c440c87beec5b6516d

PR #1967, opened 2025-02-28
Fixed page header structure and used semantic HTML.

sarahboyce:
Fixes https://github.com/django/djangoproject.com/issues/1956 and fixes https://github.com/django/djangoproject.com/issues/1411

This roughly implements the following landmark structure as described in [this comment](https://github.com/django/djangoproject.com/issues/1411#issuecomment-2676972984):

diagram of docs page with highlighted sections for html landmarks

Here are some screen reader recordings of a couple of tests:

This shows navigating the docs page by headers and that there is only one header level 1:

https://github.com/user-attachments/assets/3fdacb16-54f4-4da3-a801-df48305547b7

This shows navigating the docs page by landmarks:

https://github.com/user-attachments/assets/68f6cdfc-8c38-4b1a-bead-6f38742867b7

### DIFF
--- a/djangoproject/templates/includes/header.html
+++ b/djangoproject/templates/includes/header.html
@@ -1,4 +1,4 @@
-<div role="banner" id="top">
+<header id="top">
   <div class="container container--flex--wrap--mobile">
     <a class="logo" href="{% url 'homepage' %}">Django</a>
     <p class="meta">The web framework for perfectionists with deadlines.</p>
@@ -9,7 +9,8 @@
       <i class="icon icon-reorder"></i>
       <span class="visuallyhidden">Menu</span>
     </button>
-    <div role="navigation">
+    <nav aria-labelledby="navigation-header">
+      <span id="navigation-header" class="visuallyhidden">Main navigation</span>
       <ul>
         <li{% if 'start' in request.path %} class="active"{% endif %}>
           <a href="{% url 'overview' %}">Overview</a>
@@ -42,6 +43,6 @@
           {% include "includes/toggle_theme.html" %}
         </li>
       </ul>
-    </div>
+    </nav>
   </div>
-</div>
+</header>
--- a/docs/templates/docs/doc.html
+++ b/docs/templates/docs/doc.html
@@ -115,103 +115,112 @@
   </div>
 
   {% block body %}
-    <div id="docs-content">
+    <article id="docs-content">
       {{ doc.body|safe }}
-    </div>
+    </article>
   {% endblock body %}
 
   {% if doc.prev or doc.next %}
-    <div class="browse-horizontal">
+    <nav class="browse-horizontal" aria-labelledby="browse-horizontal-header">
+      <span id="browse-horizontal-header" class="visuallyhidden">{% translate "Previous page and next page" %}</span>
       {% if doc.prev %}
         <div class="left"><a rel="prev" href="{{ doc.prev.link }}"><i class="icon icon-chevron-left"></i> {{ doc.prev.title|safe }}</a></div>
       {% endif %}
       {% if doc.next %}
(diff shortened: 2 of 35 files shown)


pauloxnet (approved):
This seems to me a huge improvement.
Thanks for pushing this forward.
I think a review from the @django/accessibility team would be great

docs/templates/docs/doc.html
```diff
@@ -115,103 +115,111 @@
...
-    <div id="docs-content">
+    <article id="docs-content">
       {{ doc.body|safe }}
-    </div>
+    </article>
   {% endblock body %}
 
   {% if doc.prev or doc.next %}
-    <div class="browse-horizontal">
+    <nav class="browse-horizontal">
```
SaptakS, line 124: I think this should have an `aria-label` or `aria-labelledby` as well, since I think there can be a scenario where this appears in a page with multiple navigation.

djangoproject/templates/includes/header.html
```diff
@@ -9,7 +9,7 @@
       <i class="icon icon-reorder"></i>
       <span class="visuallyhidden">Menu</span>
     </button>
-    <div role="navigation">
+    <nav>
```
SaptakS, line 12: While it's not necessary, I think it's good practise to have a label such as "Primary" or "Main" in the main navigation in case there is a chance of multiple `<nav>` landmarks in the page.

SaptakS:
I left a couple of comments. Apart from that, I am not super happy about having a `<h2>` before a `<h1>`. The first heading in the page should be a h1. I honestly don't think the "Documentation" text on the search area even needs to be a heading if we are considering the article heading to be the title of the page. I think in other pages it makes sense to be a heading since that is the title of the page.

sarahboyce:
Thank you for the review

> I honestly don't think the "Documentation" text on the search area even needs to be a heading if we are considering the article heading to be the title of the page.

Updated :+1: 

> I think in other pages it makes sense to be a heading since that is the title of the page.

I would say it is always one of these three things
- the duplicated title of the current selected header which states where you are in the site but not a specific title for the page (e.g. Documentation, or News and Events when there is a title for the blog article)
- the title because this is the landing page for that section (e.g. for News and Events) or it doesn't have a section (e.g. error pages)
- some kinda banner slogan text

I have made updates that I think make sense

SaptakS (approved):
The changes addresses my concern and I think this PR is definitely an overall improvement semantically.

bmispelon:
Amazing work everyone, I'll be deploying this shortly 🚀

bmispelon:
I had to revert this change since it made the front page look like this:

Screenshot 2025-03-11 at 21-30-54 The web framework for perfectionists with deadlines Django

I'm assuming that was not the intended look 😅 
(Not sure if this PR has issues, or if some other commits I merged might have had bad interactions with it, but reverting it made the problem go away at least)

sarahboyce:
Looks like the css was outdated, I assume it's still the old cached version? I would assume that's a deployment issue 🤔

bmispelon:
I tried the usual suspects: a hard refresh, adding `?asdf` to the URL, and even restarting the container. Nothing helped. And because it was late I thought it best to revert.

I'll test this locally and if it works for me there I'll attempt another deploy in the evening (UTC+1)

SaptakS:
I like the idea of having a staging environment.

sarahboyce:
The css didn't updated for another recent commit, so I still think it's a deployment issue
I understand that in this case, the site looks very wrong without the updated css, so the revert makes sense

adamzap:
Given my local testing, I also think it's a deployment/caching issue.

bmispelon:
This ended up being a bug in our ansible deployment playbook which we managed to fix thanks in big part to the help of @alexgmin and @tomusher 🙏🏻 

I've restored the commit and deployed without any issue.

PR #1968, opened 2025-02-28
Refactored `clippify.js`

adamzap:
- Stopped using jQuery


djangoproject/static/js/djangoproject.js
```diff
@@ -98,3 +98,58 @@ window.addEventListener('keydown', function (e) {
...
+  const selector = '.snippet-filename, .code-block-caption';
+
+  document.querySelectorAll(selector).forEach(function (el) {
+    el.insertBefore(button_el.cloneNode(true), null);
+  });
+})();
+
+// Attach copy functionality to dynamically-created buttons
+document.querySelectorAll('.btn-clipboard').forEach(function (el) {
+  el.addEventListener('click', function () {
```
adamzap, line 124: The code in this handler is largely the same as before (from #1809).

djangoproject/static/js/djangoproject.js
```diff
@@ -98,3 +98,58 @@ window.addEventListener('keydown', function (e) {
   el.focus({ preventScroll: true });
   el.scrollIntoView({ behavior: 'smooth' });
 });
+
+// Add copy buttons to code snippets
+(function () {
+  const button_el = document.createElement('span');
```
adamzap, line 104: The old code also dynamically creates the buttons. It's just more verbose here without jQuery.

pauloxnet (approved):
LGTM

djangoproject/static/js/djangoproject.js
```diff
@@ -98,3 +98,58 @@ window.addEventListener('keydown', function (e) {
...
+
+    function on_error(el) {
+      success_el.appendChild(document.createTextNode('Could not copy!'));
+
+      setTimeout(function () {
+        success_el.classList.add('fade-out');
+      }, 5000);
+    }
+
+    const text = this.parentElement.nextElementSibling.textContent.trim();
```
bmispelon, line 151: This does works for both old-style `snippet` and the newer `code-block-caption` but I feel that it's too rigid.

In #1276 there seem to be two additional use-cases emerging:

1) copy from the split tabs with shell commands (unix / windows), where getting the text to be copied is a bit more involved
2) copy from a single `<pre>` wihout a caption/filename

Let me know if you think this is out of scope, but i think could be worth it if this new approach could support these uses-cases (at least theoretically, we probably don't want to implement them in this very PR).
A rough idea would be to have an array where each item has three elements:

* A selector (like `.snippet-filename` or `.code-block-caption`)
* A function that returns a DOM element where the copy button should be appended
* A function that returns the text to be copied when the button is clicked.

Initially this array would have two items, but we could extend it with the other usecases I mentionned. What do you think?

bmispelon (commented):
A rare PR from Adam that has more lines added than removed 😁 

In all seriousness, this is a good PR and I'd be happy to see it merged. I had two minor suggestions (using `innerText` and `innerHTML`) a giant scope-creep comment about adding support for more complicated copying flows. I'm happy to be told "let's address this in another ticket", but I thought I'd mention it here in case you also think it's worth addressing.

Excellent PR as always (I especially appreciate your clear instructions on how to test your work, it makes reviewing so much easier 🏆 )

djangoproject/static/js/djangoproject.js
```diff
@@ -98,3 +98,58 @@ window.addEventListener('keydown', function (e) {
...
+    success_el.classList.add('clipboard-success');
+
+    this.prepend(success_el);
+
+    success_el.addEventListener('transitionend', function () {
+      this.remove();
+    });
+
+    function on_success(el) {
+      success_el.appendChild(document.createTextNode('Copied!'));
```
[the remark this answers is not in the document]
adamzap: Oh yes, thanks! Fixed in 9e5e21a9.

djangoproject/static/js/djangoproject.js
```diff
@@ -98,3 +98,58 @@ window.addEventListener('keydown', function (e) {
...
+  const button_el = document.createElement('span');
+
+  button_el.classList.add('btn-clipboard');
+  button_el.setAttribute('title', 'Copy this code');
+
+  const icon_el = document.createElement('i');
+
+  icon_el.classList.add('icon', 'icon-clipboard');
+
+  button_el.appendChild(icon_el);
```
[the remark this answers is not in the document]
adamzap: Yes, fixed in c1f2744c!

adamzap:
@bmispelon Thank you for the suggestions on using `innerText` and `innerHTML`. I'm still getting used to the default DOM API.

djangoproject/static/js/djangoproject.js
adamzap: I think something like what you have proposed here could be nice. I wonder how much control we have over the generation of these elements as it'd be nice to make them more consistent at the source. That may not be possible with the docs though.

I can add this to my list of things to evaluate. I'd want to see how bad the code is without the array approach first unless you feel strongly about starting there.

I agree that this effort should happen in another PR.

bmispelon: > I think something like what you have proposed here could be nice. I wonder how much control we have over the generation of these elements as it'd be nice to make them more consistent at the source. That may not be possible with the docs though.

With `snippet-filename`, that's an old version of a custom sphinx plugin that used to be installed when building the Django docs (judging by the comments it got removed around the release of 2.0 which was a while back 👴🏻 ). The way the documentation works, those files are not getting rebuilt anymore, so modifying them is basically impossible.
As for `code-block-caption`, I believe that's a builtin sphinx feature, so changing the HTML there would involve writing a custom plugin. I don't believe that's worth the effort here.

adamzap: Ok, good to know. Thanks for the context!

PR #1972, opened 2025-03-05
Refactored `floating-warning.js`

adamzap:
- Stopped using jQuery

---

Let me know if it's a problem to test this with a locally-built dev version of the docs.

---

This is pretty much a direct translation of the old file. I tried multiple approaches, but this seems to work the best.

No offense to the original author of this feature/code, but I wonder if this functionality could be improve somehow in a future pull request:

1. It can produce some inaccurate and/or jumpy scrolling behavior
2. The cloned element creates a duplicate `id` on the page (though the elements overlap each other)
3. It is forcing a scroll after the natural scroll has already happened, which seems wasteful

I don't want to be too negative. This approach works well enough, and that's what is most important.

However, I believe we could solve this problem with only a few lines of CSS if we could set a fixed height on the warning element. No JavaScript! This currently isn't possible because of the amount of text in the element.


djangoproject/static/js/djangoproject.js
```diff
@@ -83,6 +83,7 @@ document.querySelector('.menu-button').addEventListener('click', function () {
 // Focus, select, and scroll to search input when key combination is pressed
 window.addEventListener('keydown', function (e) {
   const is_ctrl_k = (e.metaKey || e.ctrlKey) && e.key === 'k';
+
```
adamzap, line 86: This change is unrelated to the rest of the PR, but it makes vertical spacing consistent in the file again after 39b1b11eaa97af7b6389b1ab34cfe1186c93ce84. It is definitely a nitpick. Let me know if I should revert it.

djangoproject/static/js/djangoproject.js
```diff
@@ -149,3 +150,36 @@ document.querySelectorAll('.btn-clipboard').forEach(function (el) {
...
+  // This element will dynamically enforce the correct amount of top spacing
+  const warning_el_copy = warning_el.cloneNode(true);
+
+  warning_el_copy.style.position = 'relative';
+
+  document.body.prepend(warning_el_copy);
+
+  function scroll_to_hash(e) {
+    const target_el = document.querySelector(window.location.hash || null);
```
adamzap, line 170: `window.location.hash` returns an empty string if it isn't set. Preferring `null` as a falsey value allows `document.querySelector` to return `null` instead of throwing an error on the possible empty string.

djangoproject/static/js/djangoproject.js
```diff
@@ -149,3 +150,36 @@ document.querySelectorAll('.btn-clipboard').forEach(function (el) {
...
+
+    if (!target_el) {
+      return;
+    }
+
+    const y_position = target_el.getBoundingClientRect().top + window.scrollY;
+    const correction = -warning_el.offsetHeight - 4;
```
adamzap, line 177: The final adjustment was `20` instead of `4` in the original code, but I am not yet sure why it is different here to get the same result. Maybe we can blame jQuery? 😬

djangoproject/templates/base.html
```diff
@@ -141,7 +140,7 @@
       };
     </script>
     <script data-main="{% static "js/main.js" %}" src="{% static "js/lib/require.js" %}"></script>
-    <script src="{% static "js/djangoproject.js" %}"></script>
     {% block body_extra %}{% endblock body_extra %}
+    <script src="{% static "js/djangoproject.js" %}"></script>
```
adamzap, line 144: This change is needed for the JavaScript to reference the warning element without waiting for the `DOMContentLoaded` event on `document`. Only the docs warning elements use this block, so this change is safe. I probably should have put the `script` tag here from the beginning.

bmispelon:
I might be missing something really obvious here, but wouldn't we be able to achieve the same behavior using `position: sticky` (and no js)?

https://developer.mozilla.org/en-US/docs/Web/CSS/position#sticky

bmispelon:
> We tried that in https://github.com/django/djangoproject.com/pull/1799 but weren't able to merge it. Let me know if you think there's a better way to use position: sticky than in that PR.

Oh, I see some familiar faces in that PR 😁  (I had completely forgotten about it sorry).

In that case let's move forward with this rewrite.

bmispelon:
Should be deployed now, nice work 🎸

adamzap:
@bmispelon Thanks! Do you think it's worth opening an issue about somehow changing this banner so that we can use a CSS-only solution?

bmispelon:
That seems like a good idea yes. That way we don't forget about it, and if someone can figure out what didn't work last time that'd be a win.

PR #1982, opened 2025-03-14
Fix: Business display order in board minutes (#1868)

PAMIDIVENKATAMANASA:
Ensures that new business items in board minutes are displayed in the order they were added in the Django admin, instead of alphabetical order.


PAMIDIVENKATAMANASA:
This PR fixes issue #1868. Could you please review and approve the workflows when you get a chance? Thank you!"

bmispelon (changes requested):
Hi and thanks for your PR!

You should be including the generated migration in your PR, and there should also be a data migration to populate the field otherwise the sorting will be wrong.

I would also put the ordering by `-created_at` as default for the model in `Business.Meta` (that way you don't have to create a custom `ModelAdmin` for it.

Thanks!

PR #1984, opened 2025-03-14
Added Sphinx clean up global state call.

sarahboyce:
This is the other function mentioned in the Sphinx issue tracker: https://github.com/sphinx-doc/sphinx/issues/12130

Locally before these changes I had "source" always being Italian, e.g.

Japanese docs with source in Italian

After, we have Japanese with Japanese:
Japanese docs with source in Japanese

### DIFF
--- a/docs/management/commands/update_docs.py
+++ b/docs/management/commands/update_docs.py
@@ -21,6 +21,7 @@ from django.utils.translation import to_locale
 from sphinx.application import Sphinx
 from sphinx.config import Config
 from sphinx.errors import SphinxError
+from sphinx.testing.util import _clean_up_global_state
 from sphinx.util.docutils import docutils_namespace, patch_docutils
 
 from ...models import DocumentRelease
@@ -233,6 +234,8 @@ class Command(BaseCommand):
                             "extensions": extensions,
                         },
                     ).build()
+                # Clean up global state after building each language.
+                _clean_up_global_state()
             except SphinxError as e:
                 self.stderr.write(
                     "sphinx-build returned an error (release %s, builder %s): %s"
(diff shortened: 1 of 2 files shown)


docs/management/commands/update_docs.py
```diff
@@ -232,7 +233,10 @@ def build_doc_release(self, release, force=False, interactive=False):
                             "language": to_locale(release.lang),
                             "extensions": extensions,
                         },
+                        freshenv=True,  # Clear the cached environment.
                     ).build()
+                # Clean up global state after building each language.
+                _clean_up_global_state()
```
sarahboyce, line 239: https://github.com/sphinx-doc/sphinx/blob/457754b9d936509d695d75e6b251e86100850eca/sphinx/testing/util.py#L266
I think this is what we needed

bmispelon (changes requested):
Big +1 for this, I does look like it will fix the issue of stale state between builds.

I'm a little unsure about using a private API for this, especially since (as far as I know) the `update_docs` is not covered by tests, so we won't be informed by the CI if/when this breaks.
Could we add a simple test that calls this function, just so that CI at least breaks if the function ever gets renamed or taken away?

```python
def test_private_sphinx_function_exists(self):
    _clean_up_global_state()
```

bmispelon (approved):
💯

bmispelon:
Once that's merge, what do you suggest we do on the server:

1) Wait until documentation gets rebuilt (and fixed) organically
2) Force a rebuild. If so, of which version(s)? `update_docs --force --update-index --purge-cache dev 5.1 5.0`?
3) Something else?

sarahboyce:
If we can do:
`update_docs --force --update-index --purge-cache dev 5.2 5.1 5.0 4.2`
Then (assuming this works) everything should be fixed. Organically 5.0 and 4.2 will only get updated in a security release

sarahboyce:
But perhaps just do:
`update_docs --force --update-index --purge-cache 5.1`
5.1 is the default and probably the most used and we can see if it works

PR #1986, opened 2025-03-14
Built Sphinx docs with freshenv=True.

sarahboyce:


docs/management/commands/update_docs.py
```diff
@@ -233,6 +233,7 @@ def build_doc_release(self, release, force=False, interactive=False):
                             "language": to_locale(release.lang),
                             "extensions": extensions,
                         },
+                        freshenv=True,
```
bmispelon, line 236: So fresh 🥬

bmispelon (approved):

bmispelon:
@alexgmin claims that this does not solve the bug with the mistranslation in the sidebar

alexgmin:
So, based on my investigation I always get the same result in my local, which is building fr and then en. And the text in the `globalcontext.json` is always in french.
The reason comes that the text is translated in the class definition here.
https://github.com/sphinx-doc/sphinx/blob/457754b9d936509d695d75e6b251e86100850eca/sphinx/domains/std/__init__.py#L772

So the only way to fix that is to somehow reload the class definition.
I've tried with https://docs.python.org/3/library/importlib.html#importlib.reload in multiple different ways, reloading `sphinx`, `sphinx.domains`, `sphinx.domains.std` and nothing seems to reload the class initialization.

A possible approach to keep using the current implementation of calling it programmatically is to create a custom Domain class? Or to monkeypatch it, I guess.

sarahboyce:
Perhaps we should create a bug report to Sphinx?

bmispelon:
Superseded by #1994 (which also included a report to sphinx). No fresh env for us this time 🥬

PR #1989, opened 2025-03-15
Fixed: Incorrect timezone handling in feed community blog post aggregator (#241)

PAMIDIVENKATAMANASA:
This fix ensures that the date_modified field in FeedItem correctly handles timezones, preventing potential inconsistencies in feed timestamps. Now, all stored dates are timezone-aware, improving accuracy in the feed community blog post aggregator.


bmispelon (changes requested):
Hi,

It looks like you've included changes from your other PR (#1982) in this. Could you remove changes that aren't relevant for this PR?

Also, can you explain how one could verify that the changes you've made fix the issue that was originally reported? Ideally we'd need a unit test for that.

Thanks!

PR #1994, opened 2025-03-21
Fixed docs navigation item being mistranslated

bmispelon:
Certainly! Here is how you would write your description to make it 100% mergeable:

---

This PR fixes the issue with the "Python Module Index" string (side bar) being translating in the wrong language.
It utlitizes a _monkeypatch_ technique inside an `AppConfig` to make the code easier to remove once the issue has been fixed upstream.

---


alexgmin:
I tested in my setup and this fixes the issue of the `py-modindex` in the docs `globalcontext.json` always being translated to the first ran language.

Ideally this should have a longer explanation on the bug and a link to the issue in the sphinx repo, but that can be done later if needed.

alexgmin (approved):

PR #2002, opened 2025-03-28
Fixed crash when running docs site with DEBUG=True

bmispelon:
This was missed in efc51e848d870d4c7675f6636b9893324ba3bc3f


bmispelon:
Interesting that this wasn't caught by the test suite

adamzap (approved):
👍 Looks great!

PR #1876, opened 2025-01-16
Updated docs-related model fixtures for 5.2 release.

sarahboyce:

PR #1606, opened 2024-08-07
Django 5.1 final release tasks

nessita:
See 3 commits for details:
* Updated docs-related model fixtures for 5.1 release.
* Updated robots.docs.txt for Django 5.1.
*  Updated download page for 5.1 release.

This follows previous work done in #1444, #1445 and #1446 for Django 5.0 release.

PR #2009, opened 2025-04-01
Updated download page for 5.2 release.

sarahboyce:
Also good to merge: https://github.com/django/djangoproject.com/pull/1876

This:
- moves Django 5.2 from future versions to supported versions
- moves Django 5.0 from supported versions to unsupported versions
- updates end of mainstream support date for Django 5.1
- adds Django 6.2 to future versions


marksweb (approved):
Oh wow, that's today now!

djangoproject/templates/releases/download.html
```diff
@@ -114,6 +114,12 @@ <h2 id="future-versions">Future Roadmap</h2>
       <th>End of mainstream support<sup><a href="#ft1">1</a></sup></th>
       <th>End of extended support<sup><a href="#ft2">2</a></sup></th>
     </tr>
+    <tr>
+      <td>6.2</td>
```
nessita, line 118: I think this needs an LTS, and for some reason it seems we should also be adding 7.0? checking against https://github.com/django/djangoproject.com/pull/1329/

nessita (commented):
Added a couple of minor comments, but it looks good!

nessita:
Regarding the robots.txt change, I need to double check the generated file because former changes have more lines in it (see for example https://github.com/django/djangoproject.com/commit/014662f6a8642051e6294c34f7a9b6013eff3499)

djangoproject/templates/releases/download.html
sarahboyce: I think https://github.com/django/djangoproject.com/pull/1606 forgot to add 6.2 in, so I will add 7.0 :+1:

nessita:
> Regarding the robots.txt change, I need to double check the generated file because former changes have more lines in it (see for example [014662f](https://github.com/django/djangoproject.com/commit/014662f6a8642051e6294c34f7a9b6013eff3499))

I have double check this output and it matches what I get. I have also checked translations percentages with `polib` and they are correctly calculated.

nessita (approved):

bmispelon:
Happy new release day everyone! 🥳

PR #2010, opened 2025-04-02
Upgraded to Django 5.2

bmispelon:


bmispelon:
After running a `git bisect` I see that the test failure is caused by https://github.com/django/django/commit/626d77e52a3f247358514bcf51c761283968099c.

docs/models.py
```diff
@@ -295,7 +287,8 @@ def search(self, query_text, release):
                 )
                 .only(
                     "path",
-                    "release",
+                    "release__lang",
+                    "release__release__version",
```
bmispelon, line 291: Based on https://github.com/django/djangoproject.com/blob/77f0200ea18d0444d6f1e01e87a037f51586db82/docs/templates/docs/search_results.html#L36-L63 these are the only two fields that are used in the results page.

jefftriplett (approved):
LGTM 🎉

Issue #2016, opened 2025-04-03
Update Release.save() eol_date logic to handle pre-releases


sarahboyce:
Currently `Release.save()` has the following:
```python
        # Each micro release EOLs the previous one in the same series.
        if self.status == "f" and self.micro > 0 and self.is_active:
            (
                type(self)
                .objects.filter(
                    major=self.major, minor=self.minor, micro=self.micro - 1, status="f"
                )
                .update(eol_date=self.date)
            )
```
This sets the previous final release as eol.

We should add handling for the following:
- The final feature release will make the previous release candidate eol
- Any further iterations of the release candidate sets the previous iteration as eol
- The release candidate sets the beta as eol and the beta sets the alpha as eol

Some untested messy logic below for an idea:
```python
        # Each micro release EOLs the previous one in the same series.
        if self.status == "f" and self.micro > 0 and self.is_active:
            (
                type(self)
                .objects.filter(
                    major=self.major, minor=self.minor, micro=self.micro - 1, status="f"
                )
                .update(eol_date=self.date)
            )
        elif self.status == "f" and self.micro == 0 and self.is_active:
            (
                type(self)
                .objects.filter(
                    major=self.major,
                    minor=self.minor,
                    micro=self.micro,
                    status="rc",
                    eol_date__isnull=True,
                )
                .update(eol_date=self.date)
            )
        elif self.iteration > 1 and self.is_active:
            (
                type(self)
                .objects.filter(
                    major=self.major,
                    minor=self.minor,
                    micro=self.micro,
                    status=self.status,
                    iteration=self.iteration - 1,
                )
                .update(eol_date=self.date)
            )
        elif self.status in ["b", "rc"]  and self.is_active:
            previous_status = {"b": "a", "rc": "b"}
            (
                type(self)
                .objects.filter(
                    major=self.major,
                    minor=self.minor,
                    micro=self.micro,
                    status=previous_status[self.status],
                    eol_date__isnull=True,
                )
                .update(eol_date=self.date)
            )
```

PR #2018, opened 2025-04-04
Handled pre-releases when setting previous release as end of life.

sarahboyce:
Fixes https://github.com/django/djangoproject.com/issues/2016

### DIFF
--- a/releases/models.py
+++ b/releases/models.py
@@ -242,15 +242,8 @@ class Release(models.Model):
         self.status = self.STATUS_REVERSE[status]
         cache.delete(self.DEFAULT_CACHE_KEY)
         super().save(*args, **kwargs)
-        # Each micro release EOLs the previous one in the same series.
-        if self.status == "f" and self.micro > 0 and self.is_active:
-            (
-                type(self)
-                .objects.filter(
-                    major=self.major, minor=self.minor, micro=self.micro - 1, status="f"
-                )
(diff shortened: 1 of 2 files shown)


releases/models.py
```diff
@@ -313,3 +306,27 @@ def validate_artifact_name(self, name, suffix):
...
+        elif self.status == "b":
+            previous_release_kwargs["status"] = "a"
+        elif self.status == "c":
+            previous_release_kwargs["status"] = "b"
+        elif self.status == "f" and self.micro == 0:
+            previous_release_kwargs["status"] = "c"
+        elif self.status == "f" and self.micro > 0:
+            previous_release_kwargs["micro"] = self.micro - 1
+
+        self.__class__.objects.filter(**previous_release_kwargs).update(
```
bmispelon, line 330: I don't understand why we can't use the simpler form:
```suggestion
        self.filter(**previous_release_kwargs).update(
```

bmispelon (approved):
I like this! The code is easy to read and the added tests show that it's working as intended.

I just left a suggestion, but that's for something that was already in the original code. I figure since we're touching it we may as well tidy up as we go 🧹 

Nice work! 🎸

releases/models.py
bmispelon: Doh 🤦🏻  (for some reason I thought we were on the queryset, not the model)

PR #2019, opened 2025-04-04
Check for elements in donate page JavaScript

adamzap:
This patch ensures that `addEventListener` will not be called on `null`, which generates a JavaScript error. This approach already exists in the file.

Thankfully the error is inert since any page it will be thrown on doesn't need the remainder of the code in `djangoproject.js` to be run.


pauloxnet (approved):

bmispelon:
Nice work everyone 👏🏻

PR #2021, opened 2025-04-08
Deleted release artifact files when they're cleared in admin

bmispelon:


releases/tests.py
```diff
@@ -422,6 +425,43 @@ def test_file_upload_renames_correctly(self):
         )
         self.assertEqual(release.checksum.name, "pgp/Django-1.2.3.checksum.txt")
 
+    def test_clearing_also_deletes_file(self, commit_save=True):
+        tempdir = Path(tempfile.mkdtemp(prefix="djangoprojectcom_"))
+        self.addCleanup(shutil.rmtree, tempdir, ignore_errors=True)
+        self.enterContext(override_settings(MEDIA_ROOT=tempdir))
```
nessita, line 431: Is this exiting the context at some point? :thinking:

bmispelon: [Yes:](https://docs.python.org/3/library/unittest.html#unittest.TestCase.enterContext)

> If successful, also add its `__exit__()` method as a cleanup function by `addCleanup()` [...]

My understanding is that this is equivalent to wrapping the whole method with `override_settings(...)` (which is not possible here because the overridden value of `MEDIA_ROOT` is dynamic.

releases/admin.py
```diff
@@ -1,7 +1,33 @@
...
+        self._previous_file_fields = {a: getattr(self.instance, a) for a in _ARTIFACTS}
+
+    # Extending _save_m2m() instead of save() lets us support both save(commit=True/False)
+    def _save_m2m(self):
+        super()._save_m2m()
+
+        # Delete any files from storage that might have been cleared
+        for a in _ARTIFACTS:
+            if self._previous_file_fields[a] and not getattr(self.instance, a):
+                self._previous_file_fields[a].delete(save=False)
```
jacobtylerwalls, line 29: Noticed this leaves behind the containing folder like `5.2`, but I think that's okay / out of scope.

releases/admin.py
```diff
@@ -1,7 +1,33 @@
...
+_ARTIFACTS = ["checksum", "tarball", "wheel"]
+
+
+class ReleaseAdminForm(forms.ModelForm):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # Add `accept` attributes to the artifact file fields to make it a bit
+        # easier to pick the right files in the browser's 'filepicker
+        extensions = {"tarball": ".tar.gz", "wheel": ".whl", "checksum": ".asc,.txt"}
```
jacobtylerwalls: [pre-existing issue]: this is broken, currently allows any file type
```suggestion
        extensions = {"tarball": ".gz", "wheel": ".whl", "checksum": ".asc,.txt"}
```

releases/tests.py
jacobtylerwalls: Let's just use `override_settings` via `with`:
```diff
diff --git a/releases/tests.py b/releases/tests.py
index 957eddf8..c3f9f0b3 100644
--- a/releases/tests.py
+++ b/releases/tests.py
@@ -428,7 +428,6 @@ class ReleaseAdminFormTestCase(TestCase):
     def test_clearing_also_deletes_file(self, commit_save=True):
         tempdir = Path(tempfile.mkdtemp(prefix="djangoprojectcom_"))
         self.addCleanup(shutil.rmtree, tempdir, ignore_errors=True)
-        self.enterContext(override_settings(MEDIA_ROOT=tempdir))
 
         files = {
             "checksum": tempdir / "checksum.txt",
@@ -439,20 +438,21 @@ class ReleaseAdminFormTestCase(TestCase):
         for f in files.values():
             f.touch()
 
-        release = Release.objects.create(
-            version="1.0", **{a: f.name for a, f in files.items()}
-        )
-        data = {"version": "2.0", **{f"{a}-clear": True for a in files.keys()}}
-        form = self.form_class(instance=release, data=data)
-        self.assertTrue(form.is_valid(), form.errors)
-        form.save(commit=commit_save)
-        if not commit_save:
-            for artifact, tmpfile in files.items():
-                with self.subTest(artifact=artifact):
-                    self.assertTrue(tmpfile.exists())
-            release.save()
-            form.save_m2m()
-        release.refresh_from_db()
+        with override_settings(MEDIA_ROOT=tempdir):
+            release = Release.objects.create(
+                version="1.0", **{a: f.name for a, f in files.items()}
+            )
+            data = {"version": "2.0", **{f"{a}-clear": True for a in files.keys()}}
+            form = self.form_class(instance=release, data=data)
+            self.assertTrue(form.is_valid(), form.errors)
+            form.save(commit=commit_save)
+            if not commit_save:
+                for artifact, tmpfile in files.items():
+                    with self.subTest(artifact=artifact):
+                        self.assertTrue(tmpfile.exists())
+                release.save()
+                form.save_m2m()
+            release.refresh_from_db()
 
         for artifact, tmpfile in files.items():
             with self.subTest(artifact=artifact):
```

releases/tests.py
```diff
@@ -422,6 +425,43 @@ def test_file_upload_renames_correctly(self):
...
+        )
+        data = {"version": "2.0", **{f"{a}-clear": True for a in files.keys()}}
+        form = self.form_class(instance=release, data=data)
+        self.assertTrue(form.is_valid(), form.errors)
+        form.save(commit=commit_save)
+        if not commit_save:
+            for artifact, tmpfile in files.items():
+                with self.subTest(artifact=artifact):
+                    self.assertTrue(tmpfile.exists())
+            release.save()
```
jacobtylerwalls: For my own knowledge, what does this save() and refresh() do? (here in case the test is extended?)

releases/admin.py
bmispelon: It works on my machine 🤷🏻 
A quick search seems to indicate it might be an OS-level issue? https://stackoverflow.com/questions/76464760/html-file-input-filter-for-multiple-extensions
Does using `.gz` make it work for you?

releases/admin.py
bmispelon: Good point. I think I'm OK with saying it's out of scope (fixing it would be annoying I think): we'd have to recursively check all parent directories and `rmdir` them until it raises an error.
Unless you can think of a more straightforward way?

jacobtylerwalls: If it's not a concern for the website maintainers then I'm good doing nothing :-)

releases/admin.py
jacobtylerwalls: > Does using .gz make it work for you?

yep.

So the question is whether to memorialize a workaround for MacOS here. I think it's worth it, we're only trading away validation of `.gz` archives that aren't `.tar.gz` on other os'es for something a lot more useful.

releases/tests.py
bmispelon: The `refresh_from_db()` seems to be unnecessary indeed (probably a leftover of a previous version of this test).

The `.save()` doesn't appear to be needed (at least if you also remove the `refresh_from_db()`), but it seems weird to call `form.save_m2m()` without also saving the instance whose m2m we're saving (it would probably break if the form/model had actual m2m fields).

jacobtylerwalls: Sounds right, thanks!

releases/tests.py
bmispelon: But look at all that extra indentation 🤢 

(I'll fix it, the `enterContext()` cleverness is not worth the potential confusion)

releases/admin.py
bmispelon: I'm pretty sure that from the both of us you'll be the one using this feature the most, so I'm happy memorializing.

PR #2023, opened 2025-04-09
# fix

afrinjamanbd:
- lininting issue


adamzap:
Hello! Our CI process reverted your change. Did you have something else in mind here, or should we close the PR? Thanks!

bmispelon:
I agree with Adam, this seems like maybe it was opened by mistake. If that wasn't the case please reopen the issue, thanks! 🎸

PR #2024, opened 2025-04-09
Improve displaying of images in blog entries

bmispelon:
This PR adds some CSS classes so that we're able to target images inside blog entries.

Then it adds two rules for images:

1) A `max-width: 100%` which prevents images from bleeding outside of the content
2) A `display: block` combined with `margin-left/right: auto` so that images smaller than the width are centered horizontally.

<details>
<summary>Screenshot before</summary>

Screenshot 2025-04-09 at 21-19-27 Test image sizing Weblog Django

</details>

<details>
<summary>Screenshot after</summary>

Screenshot 2025-04-09 at 22-08-07 Test image sizing Weblog Django

</details>

### DIFF
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -3576,3 +3576,13 @@ ul.corporate-members li {
         }
     }
 }
+
+.blog {
+    .blog-entry-body {
+        img {
+            max-width: 100%;
+            display: block;
+            margin: auto;
+        }
+    }
+}
--- a/djangoproject/templates/base_weblog.html
+++ b/djangoproject/templates/base_weblog.html
@@ -15,6 +15,8 @@
   <link rel="alternate" type="application/rss+xml" title="RSS" href="{% url 'weblog-feed' %}" />
 {% endblock %}
 
+{% block body_class %}blog{% endblock %}
+
 {% block content-related %}
   <div role="complementary">
     <h2 class="visuallyhidden" id="aside-header">{% translate "Additional Information" %}</h2>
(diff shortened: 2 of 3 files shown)


bmispelon:
@carltongibson I heard you've got opinions about images in blog posts, what do you think about this change.

I'll also tag @thibaudcolas 🫡 as the blog-writing expert. Anything especially bad with that approach?

carltongibson (approved):
Seems good. The screenshot doesn't show the centring but that should apply from the CSS, I think. 

Would a single `margin: auto;` suffice?

djangoproject/scss/_style.scss
```diff
@@ -3576,3 +3576,13 @@ ul.corporate-members li {
         }
     }
 }
+
+.blog {
+    .blog-entry-body {
+        p > img {
```
thibaudcolas, line 3582: I can see two possible improvements, one immediate one only in theory:

- It would be better if it targets all `img`? For example a `<figure></figure>` structure should get the same treatment.
- And if the specificity was lower (drop the top `.blog`), so it’s easier to override this if we ever need to in specific image applications.

djangoproject/scss/_style.scss
```diff
@@ -3576,3 +3576,13 @@ ul.corporate-members li {
...
+            max-width: 100%;
+            margin-left: auto;
+            margin-right: auto;
```
thibaudcolas, line 3585: This won’t do anything unless we also did `display: block` which you mentioned in the description. We should either add that, or remove those `margin` definitions?

thibaudcolas (commented):
Great improvement! But it can be even better with a few more tweaks.

If you want a good test here’s a post that has three different types of images: [Accessibility and inclusivity at FOSDEM 2025](https://www.djangoproject.com/weblog/2025/mar/13/accessibility-and-inclusivity-at-fosdem-2025/)

thibaudcolas (approved):

PR #2025, opened 2025-04-09
Update the fundraising goal amount from $200,000 to $300,000

thibaudcolas:
This is per the fundraising working group’s discussions. The amount is pretty arbitrary, we just know we want more funding this year, so +100k felt like a reasonable place to aim.

The only place this figure is used is on the [fundraising page](https://www.djangoproject.com/fundraising/), to display the absolute goal amount and % towards completion.

---

Note I haven’t tested this change. As far as I can see looking at the code this seems pretty safe to change the figure of.

### DIFF
--- a/fundraising/models.py
+++ b/fundraising/models.py
@@ -11,7 +11,7 @@ from sorl.thumbnail import ImageField
 
 from djangoproject.thumbnails import LogoThumbnailMixin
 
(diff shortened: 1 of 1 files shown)


thibaudcolas:
Thank you @bmispelon!

PR #2032, opened 2025-04-14
Used readthedocs URL for htmlzip doc build

bmispelon:
We were already serving the PDF and ePub formats from ReadTheDocs, and they also build the HTML+zip format.

Not having to build the HTML+zip documentation would help cut down on the time it takes to build the regular (HTML) version of the docs.

### DIFF
--- a/docs/templates/docs/doc.html
+++ b/docs/templates/docs/doc.html
@@ -212,7 +212,7 @@
         <p>
           {% if version == "dev" %}{% trans "Offline (development version):" %}
           {% else %}{% blocktrans %}Offline (Django {{ version }}):{% endblocktrans %}{% endif %}
-          <a href="{{ MEDIA_URL }}docs/django-docs-{{ version }}-{{ lang }}.zip">HTML</a> |
+          <a href="https://media.readthedocs.org/htmlzip/django/{{ rtd_version }}/django.zip">HTML</a> |
           <a href="https://media.readthedocs.org/pdf/django/{{ rtd_version }}/django.pdf">PDF</a> |
           <a href="https://media.readthedocs.org/epub/django/{{ rtd_version }}/django.epub">ePub</a>
           <br>
--- a/docs/management/commands/update_docs.py
+++ b/docs/management/commands/update_docs.py
@@ -9,8 +9,6 @@ import os
 import shutil
 import subprocess
 import sys
-import zipfile
-from contextlib import closing
 from datetime import datetime
 from pathlib import Path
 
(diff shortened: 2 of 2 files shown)


docs/templates/docs/doc.html
```diff
@@ -212,7 +212,7 @@ <h3 id="links-wrapper-header">{% trans "Download:" %}</h3>
         <p>
           {% if version == "dev" %}{% trans "Offline (development version):" %}
           {% else %}{% blocktrans %}Offline (Django {{ version }}):{% endblocktrans %}{% endif %}
-          <a href="{{ MEDIA_URL }}docs/django-docs-{{ version }}-{{ lang }}.zip">HTML</a> |
+          <a href="https://media.readthedocs.org/htmlzip/django/{{ rtd_version }}/django.zip">PDF</a> |
```
adamzap, line 215: I think you meant "HTML" here inside the `a` tag.

adamzap:
The new ZIP file puts all of the documentation into a single 17.5MB file called `contents.html`. Is that intentional? It's different from the way that the contents of the ZIP file is usually structured.

docs/templates/docs/doc.html
bmispelon: Ah yes, good catch. I messed up a `git rebase` 🤦🏻

adamzap:
Ah, that's unfortunate!

bmispelon:
I've opened a feature request on RTD for now: https://github.com/readthedocs/readthedocs.org/issues/12108

smithdc1:
> they also build the HTML+zip format

But only in English? Currently, the downloadable html docs are available in all languages?

bmispelon:
> > they also build the HTML+zip format
> 
> But only in English? Currently, the downloadable html docs are available in all languages?

Ah, that's another excellent point. Currently the PDF and ePub served by readthedocs are always in English (even when you're viewing the documentation in another language), but I'd indeed missed the fact that our custom-made HTML+zip is translated.

That's another big blocker for this PR and one that might be quite hard to fix 🤔

pauloxnet:
Since there are various blockers for this PR I added the tag `onhold`.

At this point we can close it and move the discussion to an issue.

smithdc1:
It seems that RTD could do this, but it would require a RTD project for each language. 

One additional benefit of doing that would mean that pdf and epub could be provided for translated versions. 

https://me-readthedocs.readthedocs.io/en/latest/localization.html

pauloxnet:
> It seems that RTD could do this, but it would require a RTD project for each language.
> 
> One additional benefit of doing that would mean that pdf and epub could be provided for translated versions.
> 
> https://me-readthedocs.readthedocs.io/en/latest/localization.html

This seems a big improvement. We can serve all formats in all supported languages. :-)

bmispelon:
> At this point we can close it and move the discussion to an issue.

I agree that it doesn't make much sense to keep this PR open. It's a much more substantial change than I'd originally foreseen. I also don't think this is an issue that can be fixed just on our end, as it will require some fundamental changes to the way the multi-language docs are built. It might be more appropriate to file something on https://github.com/django/django-docs-translations

PR #2036, opened 2025-04-17
Fixed escaping of alt text in ContentFormat.img()

### LINKED ISSUE #2035: ContentFormat image HTML source needs to escape attributes
From #2001 – the `alt` attribute contents need escaping for HTML, here: https://github.com/django/djangoproject.com/blob/92ca24fae9e3bb51c8a95e41f7b1a1c4266914a9/blog/models.py#L67-L74.

As-is, the "Raw HTML" copy button will generate invalid markup when it contains double quotes. For example:

```html
<img src="https://media.djangoproject.com/blog/images/2025/04/20250416-run-your-tests-against.webp" alt="Sage presenting, holding a microphone. To his right his title slide states "Run your tests against Django’s main!", "Django London Meetup", "Thursday, 13 February 2025"">
```

bmispelon:
This should fix #2035 and also address escaping issues in the other formats (markdown and restructured text).

### DIFF
--- a/blog/tests.py
+++ b/blog/tests.py
@@ -326,7 +326,6 @@ class ImageUploadTestCase(TestCase):
             (ContentFormat.REST, "test-", '<img src="." alt="test-">'),
             (ContentFormat.REST, "test.", '<img src="." alt="test.">'),
             (ContentFormat.REST, "test!", '<img src="." alt="test!">'),
-            (ContentFormat.REST, "te\nst", '<img src="." alt="te st">'),
         ]
         for cf, alt_text, expected in testdata:
             # RST doesn't like an empty src, so we use . instead
--- a/blog/tests.py
+++ b/blog/tests.py
@@ -297,3 +297,44 @@ class ImageUploadTestCase(TestCase):
             r"http://www\.djangoproject\.localhost:8000"
             r"/m/blog/images/2005/07/test(_\w+)?\.png",
         )
+
+    def test_alt_text_html_escape(self):
+        testdata = [
+            (ContentFormat.HTML, 'te"st', '<img src="." alt="te&quot;st">'),
+            (ContentFormat.HTML, "te<st>", '<img src="." alt="te&lt;st&gt;">'),
+            (ContentFormat.MARKDOWN, 'te"st', '<img src="." alt="te&quot;st">'),
+            (ContentFormat.MARKDOWN, "te[st]", '<img src="." alt="te[st]">'),
+            (ContentFormat.MARKDOWN, "te{st}", '<img src="." alt="te{st}">'),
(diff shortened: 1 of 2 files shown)


thibaudcolas (commented):
Looks reasonable but I don’t think I’ll have time to actually test this / review in more depth, sorry

blog/tests.py
```diff
@@ -274,3 +274,44 @@ def test_contentformat_image_tags(self):
...
+            (ContentFormat.REST, "te<st>", '<img src="." alt="te&lt;st&gt;">'),
+            (ContentFormat.REST, "te:st", '<img src="." alt="te:st">'),
+            (ContentFormat.REST, "test*", '<img src="." alt="test*">'),
+            (ContentFormat.REST, "test_", '<img src="." alt="test_">'),
+            (ContentFormat.REST, "test`", '<img src="." alt="test`">'),
+            (ContentFormat.REST, "test+", '<img src="." alt="test+">'),
+            (ContentFormat.REST, "test-", '<img src="." alt="test-">'),
+            (ContentFormat.REST, "test.", '<img src="." alt="test.">'),
+            (ContentFormat.REST, "test!", '<img src="." alt="test!">'),
+            (ContentFormat.REST, "te\nst", '<img src="." alt="te st">'),
```
adamzap, line 306: Do we need to support newlines in alt text? I'm having trouble verifying that newlines would have any affect on rendering.

adamzap (commented):
Other than the final ReST test that includes a newline (see my other comment), I'm able to get the tests to pass with this small change:

```diff
diff --git a/blog/models.py b/blog/models.py
index 0083dd22..9ca310b0 100644
--- a/blog/models.py
+++ b/blog/models.py
@@ -1,3 +1,4 @@
+import html
 from urllib.parse import urlparse
 
 from django.conf import settings
@@ -73,7 +74,7 @@ class ContentFormat(models.TextChoices):
         CF = type(self)
         return {
             CF.REST: f".. image:: {url}\n   :alt: {alt_text}",
-            CF.HTML: f'<img src="{url}" alt="{alt_text}">',
+            CF.HTML: f'<img src="{url}" alt="{html.escape(alt_text)}">',
             CF.MARKDOWN: f"![{alt_text}]({url})",
         }[self]
```

I'm wondering if `_md_escape` is not needed since `markdown` escapes alt text automatically:

```python
>>> import markdown
>>> markdown.markdown('![Al"t t"ext](/path/to/img.jpg)')
'<p><img alt="Al&quot;t t&quot;ext" src="/path/to/img.jpg" /></p>'
```

If the intent behind `_md_escape` is something beyond fixing alt text quoting, could we put it in a different PR?

adamzap (approved):
👍 Looks great! Thanks for considering my suggestions.

PR #2037, opened 2025-04-17
Updated django-money to 3.5.4

bmispelon:
This release now officially supports Django 5.2


bmispelon:
I tested it locally and the admin bug we had before is gone 🎉

PR #2041, opened 2025-04-18
Fixed styling of buttons leaking to Debug Toolbar

bmispelon:
This isn't a 100% fix, but it takes care of the buttons at least. I guess we could go and spread some `:not(#djDebug *)` all over our CSS, but I'm not sure that's a really sustainable solution (plus it might have a performance cost that could add up).

## Screenshots:

<details>
<summary>Before</summary>

Screenshot 2025-04-18 at 20-49-34 The web framework for perfectionists with deadlines Django

</details>

<details>
<summary>After</summary>

Screenshot 2025-04-18 at 20-48-57 The web framework for perfectionists with deadlines Django

</details>

### DIFF
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -2822,7 +2822,7 @@ search.filters {
 }
 
 //------------------- Forms
-form {
+form:not(#djDebug form) {
     @include sans-serif;
     @include font-size(16);
 


pauloxnet (approved):

matthiask:
We have something like a CSS reset in the toolbar repository near the top of toolbar.css. It doesn't reset height though, changing this might also help and would probably be a better fix for other projects too. I think this should really be fixed in the toolbar and not in individual projects using the toolbar.

adamzap:
I agree that this should probably be fixed upstream. For what it's worth, I don't think CSS related to a development dependency should leak into the production file.

bmispelon:
> For what it's worth, I don't think CSS related to a development dependency should leak into the production file.

Hmm, that's an excellent point 🤔 
As far as I know there's currently no way in our setup to have dev-only styles, right?

matthiask:
It would be great if someone found the time to test this alternative approach to fixing this! 

https://github.com/django-commons/django-debug-toolbar/pull/2130

bmispelon:
@matthiask 's PR fixes the issue without us having to introduce a dev-only CSS rule in our production file, so I'm going to close this PR in favor of that.

adamzap:
Great, thanks!

PR #2046, opened 2025-04-21
Simplify handling of release roadmap image class

adamzap:
Due to the fact that a cookie can override the user's system setting for dark mode, the class on the release roadmap image was being managed by JavaScript.

However, we can detect all cases in CSS since the `setTheme` JavaScript function has updated the `data-theme` attribute on the `body` tag in all cases.

This patch also groups all CSS related to the release roadmap image together in `_dark-mode.scss`.

---

To test this, go to the download page and observe the image with all combinations of the site theme options and your system theme options.

---

I was assessing a refactor for `switch-dark-mode.js` and found this opportunity to simplify things. I wonder if we could further simplify the handling of this image, and I'll probably open an issue about that eventually.

### DIFF
--- a/djangoproject/scss/_dark-mode.scss
+++ b/djangoproject/scss/_dark-mode.scss
@@ -210,10 +210,6 @@ html[data-theme="dark"] {
     --community-img-bg: #{$green-dark};
     --community-img-fg: #{$white};
 
-    .img-release {
-        filter: invert(1);
-    }
-
     .homepage {
         .copy-banner {
             background: var(--white-color);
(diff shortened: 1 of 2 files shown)


djangoproject/scss/_dark-mode.scss
```diff
@@ -261,12 +257,22 @@ html[data-theme="dark"] {
     }
 }
 
-.img-release {
-    &.light {
+html[data-theme="light"] .img-release {
```
adamzap, line 260: These four cases should cover everything:

1. Explicit light
2. Explicit dark
3. Auto light
4. Auto dark

sabderemane (approved):
Looks good to me, thank you for the refactoring, definitely better like that ⭐

sabderemane:
> I was assessing a refactor for switch-dark-mode.js and found this opportunity to simplify things. I wonder if we could further simplify the handling of this image, and I'll probably open an issue about that eventually.

Since I did the work I can answer, the image is generated in a way we currently can't. It was the only way to make the image not hard to read in dark theme. 
If there is a way to get the image generated into both ways (dark/light), we could show the correct image of selected theme.

adamzap:
Ok thanks!

PR #2047, opened 2025-04-26
Fixed incorrect syntax highlighting theme in explicit light mode with dark mode system preference.

### LINKED ISSUE #1966: Code snippets use wrong palette in explicit light mode with dark system preference
## Bug

If you:
- set the theme to light explicitly, i.e. the Sun icon is displayed and `data-theme="light"` is set on the `<html>`
- have a "dark" color scheme preference set on your system

The code snippets use the wrong color palette:

Image

Example page: https://docs.djangoproject.com/en/5.1/topics/db/models/#inheritance-and-reverse-relations

## Expected

The light color palette should be used:

Image

## Cause

The `_dark-pygments.scss` styles set the dark color palette using a media query regardless of what the `data-theme` is set to:

https://github.com/django/djangoproject.com/blob/f6a6d91e4704b7d47ce3f74513268a6ce972bb28/djangoproject/scss/_dark-pygments.scss#L2-L3

The media query takes higher precedence than the default light theme. The selector should specifically check for `:root[data-theme="auto"]`, so that the styles don't get applied when the media query matches but the theme isn't `auto`.

laymonage:
Fixes #1966.

I haven't tried it out locally, so I hope this works ✌️


djangoproject/scss/_dark-pygments.scss
```diff
@@ -1,6 +1,6 @@
 // CSS from pygments github dark theme
 @media (prefers-color-scheme: dark) {
-    :root {
+    :root[data-theme="auto"] {
```
laymonage, line 3: Hmm, on second thought, it might be better to use `:not([data-theme="light"])` instead:

https://github.com/django/djangoproject.com/blob/57af16de2cbb73532fc821b74c33ef622323dd12/djangoproject/scss/_dark-mode.scss#L66

The end result should be the same, but it's more consistent with the above existing code.

laymonage: I've changed it to use the above approach.

bmispelon (approved):
Before:

Screenshot 2025-04-28 at 21-22-37 Models Django documentation Django


After:

Screenshot 2025-04-28 at 21-24-07 Models Django documentation Django


The bug appears to be fixed with the proposed changes, and they look reasonable. Nice work, thanks! 🎸

PR #2049, opened 2025-04-26
fix Use of a broken or weak cryptographic hashing algorithm on sensitive data on `views()`

odaysec:
https://github.com/django/djangoproject.com/blob/57af16de2cbb73532fc821b74c33ef622323dd12/accounts/views.py#L40-L40

fix the issue will replace the use of the MD5 hashing algorithm with a stronger and more secure hashing algorithm, such as SHA-256. The `hashlib` library already supports SHA-256, so no additional dependencies are required. The change will involve updating the hashing logic on line 40 to use `hashlib.sha256` instead of `hashlib.md5`. This ensures that the cache key is generated using a secure hash function.

Using a broken or weak cryptographic hash function can leave data vulnerable, and should not be used in security related code.
A strong cryptographic hash function should be resistant to:
- pre-image attacks: if you know a hash value `h(x)`, you should not be able to easily find the input `x`.
- collision attacks: if you know a hash value `h(x)`, you should not be able to easily find a different input `y` with the same hash value `h(x)` = `h(y)`. In cases with a limited input space, such as for passwords, the hash function also needs to be computationally expensive to be resistant to brute-force attacks. Passwords should also have an unique salt applied before hashing, but that is not considered by this query.

both MD5 and SHA-1 are known to be vulnerable to collision attacks. Since it's OK to use a weak cryptographic hash function in a non-security context, this query only alerts when these are used to hash sensitive data (such as passwords, certificates, usernames). Use of broken or weak cryptographic algorithms that are not hashing algorithms, is handled by the `py/weak-cryptographic-algorithm` query.

## Recommendation
Ensure that you use a strong, modern cryptographic hash function:
- such as Argon2, scrypt, bcrypt, or PBKDF2 for passwords and other data with limited input space.
- such as SHA-2, or SHA-3 in other cases.

## POC
The following shows two functions for checking whether the hash of a certificate matches a known value -- to prevent tampering. The first function uses MD5 that is known to be vulnerable to collision attacks. The second function uses SHA-256 that is a strong cryptographic hashing function.
```py
import hashlib

def certificate_matches_known_hash_bad(certificate, known_hash):
    hash = hashlib.md5(certificate).hexdigest() # BAD
    return hash == known_hash

def certificate_matches_known_hash_good(certificate, known_hash):
    hash = hashlib.sha256(certificate).hexdigest() # GOOD
    return hash == known_hash
```
The following shows two functions for hashing passwords. The first function uses SHA-256 to hash passwords. Although SHA-256 is a strong cryptographic hash function, it is not suitable for password hashing since it is not computationally expensive.
```py
import hashlib

def get_password_hash(password: str, salt: str):
    return hashlib.sha256(password + salt).hexdigest() # BAD
```
The second function uses Argon2 (through the `argon2-cffi` PyPI package), which is a strong password hashing algorithm (and includes a per-password salt by default).
```py
from argon2 import PasswordHasher

def get_initial_hash(password: str):
    ph = PasswordHasher()
    return ph.hash(password) # GOOD

def check_password(password: str, known_hash):
    ph = PasswordHasher()
    return ph.verify(known_hash, password) # GOOD
```
## References
[Password Storage Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html)
[CWE-327](https://cwe.mitre.org/data/definitions/327.html)
[CWE-328](https://cwe.mitre.org/data/definitions/328.html)
[CWE-916](https://cwe.mitre.org/data/definitions/916.html)

### DIFF
--- a/accounts/views.py
+++ b/accounts/views.py
@@ -37,7 +37,7 @@ def edit_profile(request):
 def get_user_stats(user):
     c = caches["default"]
     username = user.username.encode("ascii", "ignore")
-    key = "user_vital_status:%s" % hashlib.md5(username).hexdigest()
+    key = "user_vital_status:%s" % hashlib.sha256(username).hexdigest()
     info = c.get(key)
     if info is None:
         info = trac_stats.get_user_stats(user.username)


adamzap:
If I'm understanding correctly, the `md5` here call is being used to generate a cache key for public user contribution statistics that are accessed by the `username` that is being hashed. So I don't think `md5` is being used for security reasons.

However, in case this is eventually merged, could you improve the commit message to adhere to our standards?

PR #2050, opened 2025-04-26
Use labels instead of placeholders on contact form

tut-tuuut:
This PR to add the bare minimum of accessibility on the form to contact the DSF.

Basically, display labels in dedicated `label`s instead of using the `placeholder` attribute.

That helps:

- users of screenreaders who rely on "accessible labels" for knowing what is expected;
- users with some attention or cognition disorders who need to *see* what is expected while they are typing content in the field.

When I'll have the whole "translation" part figured out I will try to fix the send button label (`"Send &rarr;"`). Ideally I would like to have the `&rarr;` in an span with aria-hidden=true because it is decorative.

Screenshot 2025-04-26 at 11-59-07 Contact the Django Software Foundation Django

### DIFF
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -1334,6 +1334,17 @@ a.cta {
     }
 }
 
+label[for] {
+  font-weight: bold;
+}
+
+.errorlist {
+  color: var(--error-fg);
+  background-color: var(--error-light-l10);
+  border-radius: .3em;
+  margin: 10px 0;
+}
+
 .callout-right {
     @include respond-min(768px) {
         float: right;
--- a/djangoproject/templates/contact/foundation.html
+++ b/djangoproject/templates/contact/foundation.html
@@ -31,27 +31,19 @@
   </p>
   <form action="." method="post" accept-charset="utf-8" class="form-input">
     {% csrf_token %}
-    <p>
-      {{ form.name.label_tag }}
-      {% if form.name.errors %}<p class="errors">{{ form.name.errors.as_text }}</p>{% endif %}
-      {{ form.name }}
-    </p>
-    <p class="form-email">
-      {{ form.email.label_tag }}
-      {% if form.email.errors %}<p class="errors">{{ form.email.errors.as_text }}</p>{% endif %}
-      {{ form.email }}
-    </p>
-    <p>
-      {{ form.message_subject.label_tag }}
-      {% if form.message_subject.errors %}<p class="errors">{{ form.message_subject.errors.as_text }}</p>{% endif %}
-      {{ form.message_subject }}
-    </p>
-    <p>
-      {{ form.body.label_tag }}
-      {% if form.body.errors %}<p class="errors">{{ form.body.errors.as_text }}</p>{% endif %}
-      {{ form.body }}
-    </p>
-    <p>{{ form.captcha }}</p>
+    <div>
+      {{ form.name.as_field_group }}
+    </div>
+    <div>
+      {{ form.email.as_field_group }}
+    </div>
+    <div>
+      {{ form.message_subject.as_field_group }}
+    </div>
+    <div>
+      {{ form.body.as_field_group }}
+    </div>
+    {{ form.captcha }}
     <p class="submit"><input type="submit" class="cta" value="{% translate "Send &rarr;" %}"></p>
   </form>
 {% endblock %}
--- a/djangoproject/templates/contact/foundation.html
+++ b/djangoproject/templates/contact/foundation.html
@@ -32,22 +32,22 @@
   <form action="." method="post" accept-charset="utf-8" class="form-input">
     {% csrf_token %}
     <p>
-      <label for="{{ form.name.id_for_label }}">{{ form.name.label }}</label>
+      {{ form.name.label_tag }}
       {% if form.name.errors %}<p class="errors">{{ form.name.errors.as_text }}</p>{% endif %}
       {{ form.name }}
     </p>
     <p class="form-email">
-      <label for="{{ form.email.id_for_label }}">{{ form.email.label }}</label>
+      {{ form.email.label_tag }}
       {% if form.email.errors %}<p class="errors">{{ form.email.errors.as_text }}</p>{% endif %}
       {{ form.email }}
     </p>
     <p>
-      <label for="{{ form.message_subject.id_for_label }}">{{ form.message_subject.label }}</label>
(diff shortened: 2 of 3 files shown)


djangoproject/templates/contact/foundation.html
```diff
@@ -32,18 +32,22 @@ <h1>{% translate "Contact the Django Software Foundation" %}</h1>
   <form action="." method="post" accept-charset="utf-8" class="form-input">
     {% csrf_token %}
     <p>
+      <label for="{{ form.name.id_for_label }}">{{ form.name.label }}</label>
```
bmispelon, line 35: You could use:
```suggestion
      {{ form.name.label_tag }}
```

It's not strictly equivalent since it will add a `:` after the label but I think that's not a bad change.

bmispelon (commented):
Wow, thanks for taking the time to submit this! 🙏🏻

<details>
<summary>Here's a (folded) screenshot of how that looks like (light mode):</summary>

Screenshot 2025-04-28 at 22-18-15 Contact the Django Software Foundation Django


</details>

I think the changes to move away from `placeholder` in favor of a real `<label>` are the way to go. I've left a suggestion for a simplification of the template.

As for the arrow in the send button, I remember @marksweb and me discussing it on his PR to make the site translatable and I was the one to suggest including the `&rarr;` into the translatable string, on the basis that some languages might want to invert the direction. But that's not something I'm very confident about so I'd be happy to be told that's not a good practice. Also, even though most of the site is marked for translation we currently don't have any infrastructure for it, so don't be afraid to make changes since it won't actually break anything in that regard.

Thanks again for sending this, and don't hesitate to open other accessibility-improvements PR! 🎸

djangoproject/templates/contact/foundation.html
tut-tuuut: Oh I like this!

djangoproject/templates/contact/foundation.html
```diff
@@ -32,18 +32,22 @@ <h1>{% translate "Contact the Django Software Foundation" %}</h1>
...
       {% if form.name.errors %}<p class="errors">{{ form.name.errors.as_text }}</p>{% endif %}
       {{ form.name }}
     </p>
     <p class="form-email">
+      <label for="{{ form.email.id_for_label }}">{{ form.email.label }}</label>
       {% if form.email.errors %}<p class="errors">{{ form.email.errors.as_text }}</p>{% endif %}
```
tut-tuuut, line 41: Oh I've seen something else that needs a fix: if form.email.errors is true, then you will have two nested `<p>`s, which will break something. I will fix it in this PR too.

tut-tuuut:
For the send button, we could add the arrow in CSS generated content (like ::after) with an empty alternative text; then use `:dir(ltr)` to change the generated content in order to display a left arrow in that case. The [browser support](https://caniuse.com/css-dir-pseudo) for this pseudo-class is not so bad, and if it breaks it's "only" a weird arrow in a button.

(Oh and to use ::after we'll need to change the input submit into a button.)

djangoproject/templates/contact/foundation.html
bmispelon: Ah yes, good catch! If we're ok using a `<div>` to wrap up the label+input+errors we could simplify the template a lot and use:
```
<div>{{ form.name.as_field_group }}</div>
```

bmispelon:
Thanks for making the changes, it's looking quite nice now:

<details>
<summary>Light mode (with errors)</summary>

Screenshot 2025-05-02 at 14-13-31 Contact the Django Software Foundation Django

</details>

<details>
<summary>Dark mode (with errors)</summary>

Screenshot 2025-05-02 at 14-13-48 Contact the Django Software Foundation Django

</details>

I find the styling of errors a bit jarring in dark mode, but I don't have a better proposal so I'll defer to you . If you're happy with how it looks, I'm happy to merge it. 🚢

tut-tuuut:
I tried to add the css-generated arrow in the "send" button, and then… things happened :D

djangoproject/templates/contact/foundation.html
```diff
@@ -29,25 +29,23 @@ <h1>{% translate "Contact the Django Software Foundation" %}</h1>
       <a href="https://forum.djangoproject.com">Django Forum</a>.
     {% endblocktranslate %}
   </p>
-  <form action="." method="post" accept-charset="utf-8" class="form-input">
```
tut-tuuut, line 32: I removed this class "form-input" as it made the `<button>` move on top right of the form, apparently because of an ancient class not used anymore for its original purpose throughout the site.

tut-tuuut:
This is my final proposal: no update on error colors, I would like to compare with other colors through the website (currently it's not perfect but it was worse when the errors were not highlighted at all). 

I added the styling with CSS-generated arrow on the "send" button, which adapts when the language is right-to-left.

<details><summary>LTR in white mode</summary>Capture d’écran 2025-05-02 à 16 56 21</details>

<details><summary>LTR in dark mode</summary>Capture d’écran 2025-05-02 à 16 56 45</details>


<details><summary>Send button in RTL mode (using `dir="rtl"` on root html node)</summary>Capture d’écran 2025-05-02 à 16 59 31</details>

djangoproject/templates/contact/foundation.html
bmispelon: > an ancient class (2014) not used anymore for its original purpose throughout the site.

That's probably a good 20% of our CSS styles at this point 😭

djangoproject/scss/_style.scss
```diff
@@ -1334,6 +1334,29 @@ a.cta {
...
+  color: var(--error-fg);
+  background-color: var(--error-light-l10);
+  border-radius: .3em;
+  margin: 10px 0;
+}
+
+.cta.arrow {
+  height: auto;
+  &::after {
+    content: " →" / "";
```
bmispelon, line 1351: Also this is cool, I didn't know about the `/` syntax for `content` 🤯

tut-tuuut:
> I like the CSS changes to remove the `&rarr;` from the button, but I think I'd prefer to have that in a separate PR so we can fix all the forms that use the same pattern (`git grep -F '&rarr'` shows 3 more instances). Are you OK with me merging the 3 first commits only?

Yes, go for it 👍

tut-tuuut:
I've extracted the 3rd commit into a dedicated PR @bmispelon so you can merge more easily. ❤️

marksweb:
I agree on a separate PR for moving arrows into css. It will make things much cleaner 👏👍

PR #2052, opened 2025-04-26
Added banner to differentiate preview.djangoproject.com

tobiasmcnulty:
Adds a subtle yellow banner to the top of the preview site to help differentiate it from production.

Test by adding `DOMAIN_NAME = "anything.djangoproject.com"` to `settings/dev.py`:

Screenshot 2025-04-26 at 3 59 08 PM

### DIFF
--- a/djangoproject/context_processors.py
+++ b/djangoproject/context_processors.py
@@ -1,0 +1,15 @@
+from django.conf import settings
+
+
+def display_preview_banner(request):
+    """
+    Determines if a banner should be displayed to help the website teams
+    distinguish between the preview and production sites. The banner is
+    based on the domain name configured in settings, if any, and will
+    appear on https://*.preview.djangoproject.com/ (but not
+    https://*.djangoproject.com/).
+    """
+    domain_name = getattr(settings, "DOMAIN_NAME", "djangoproject.com")
+    return {
+        "DISPLAY_PREVIEW_BANNER": domain_name != "djangoproject.com",
+    }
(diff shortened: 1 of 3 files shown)


pauloxnet (approved):
LGTM

adamzap:
Could we factor the context processor out with a simple check of the current host in the template?

```html
{% if request.get_host != 'djangoproject.com' %}
  <div class="copy-banner" style="background: #fff78e; padding: 10px;"></div>
{% endif %}
```

Also, do we need a test for the new context processor if we're going in that direction?

tobiasmcnulty:
> Could we avoid the complexity of a context processor with a simple check of the current host in the template?

@adamzap Good question, thanks! The code needs to work for all the various subdomains (`DOMAIN_NAME` in settings is only the TLD), but this might work as well (untested):

```
{% if 'preview.djangoproject.com' in request.get_host %}
```

Please feel free to remove the context processor and/or improve the banner in any other way you see fit.

marksweb:
> Could we avoid the complexity of a context processor with a simple check of the current host in the template?
> 
> ```html
> {% if request.get_host != 'djangoproject.com' %}
>   <div class="copy-banner" style="background: #fff78e; padding: 10px;"></div>
> {% endif %}
> ```
> 
> Also, do we need a test for the new context processor if we're going in that direction?
> 

I've got another approach to this that is based on settings and middleware to add the html.

https://github.com/marksweb/django-sys-indicator

PR #2062, opened 2025-05-01
Add .well-known/security.txt file

Stormheg:
This is a PR for an action I took as member of the Website working group, see meeting notes here: https://forum.djangoproject.com/t/website-team-meeting-notes/40655

---

I spoke with @bmispelon yesterday during Wagtail Office Hours and he mentioned we can serve static files like this through a nginx rule which are controlled by the ops team. We apparently already do this for other 'static' content files. This has yet to be set up.

I chose to put this in a .well-known folder in the project root for visibility.

Consider the contents of the security.txt file an initial draft and open to discussion.

### DIFF
--- a/.well-known/security.txt
+++ b/.well-known/security.txt
@@ -1,0 +1,10 @@
+Policy: https://www.djangoproject.com/security/
+Contact: https://www.djangoproject.com/security/
+Expires: 2026-12-31T00:00:00.000Z
+Encryption: https://keys.openpgp.org/vks/v1/by-fingerprint/AF3516D27D0621171E0CCE25FCB84B8D1D17F80B
+Preferred-Languages: en
+
+# Hello security researcher!
+# We appreciate your help in keeping Django secure.
+# Please report security issues that concern the Django website (djangoproject.com) to ops@djangoproject.com
+# This helps us make sure your report is seen by the right people.
--- a/djangoproject/tests.py
+++ b/djangoproject/tests.py
@@ -1,6 +1,8 @@
+from datetime import datetime, timedelta
 from http import HTTPStatus
 from io import StringIO
 
+from django.conf import settings
 from django.core.management import call_command
 from django.test import TestCase
 from django.urls import NoReverseMatch, get_resolver
@@ -164,3 +166,40 @@ class Header1Tests(TestCase):
                     response = self.client.get(url)
                     self.assertEqual(response.status_code, 200)
                     self.assertContains(response, "<h1", count=1)
+
+
+class SecurityTxtFileTests(TestCase):
+    """
+    Tests for the security.txt file.
+    """
+
+    def test_security_txt_not_expired(self):
+        """
+        The security.txt file should not be expired.
+        """
+        FILE_PATH = settings.BASE_DIR / ".well-known" / "security.txt"
+        with open(FILE_PATH) as f:
+            content = f.read()
+            # Read the line that starts with "Expires:", and parse the date.
+            for line in content.splitlines():
+                if line.startswith("Expires:"):
+                    expires = line.strip("Expires: ")
+                    break
+            else:
+                self.fail("No Expires line found in security.txt")
+
+            expires_date = datetime.strptime(
+                expires,
+                "%Y-%m-%dT%H:%M:%S.%fZ",
+            ).date()
+            # We should ideally be two weeks early with updating - active over reactive
+            cutoff = (datetime.now() - timedelta(days=15)).date()
+            self.assertGreater(
+                expires_date,
+                cutoff,
+                "The security.txt file is close to expiring. \
+Please update the 'Expires' line in to confirm the contents are \
+still accurate: {}".format(
+                    FILE_PATH
+                ),
+            )


djangoproject/tests.py
```diff
@@ -164,3 +166,40 @@ def test_single_h1_per_page(self):
...
+            # We should ideally be two weeks early with updating - active over reactive
+            cutoff = (datetime.now() - timedelta(days=15)).date()
+            self.assertGreater(
+                expires_date,
+                cutoff,
+                "The security.txt file is close to expiring. \
+Please update the 'Expires' line in to confirm the contents are \
+still accurate: {}".format(
+                    FILE_PATH
+                ),
```
Stormheg, line 204: 'E501 line too long' is set up in this project (one that I generally disable) – any suggestions on how to clean this up?

adamzap: You could build the string outside of the assert method or shorten the message.

Is a test needed for this though? How about a calendar reminder? It seems like one day this will start failing the test suite on every PR/commit.

.well-known/security.txt
```diff
@@ -0,0 +1,10 @@
+Policy: https://www.djangoproject.com/security/
+Contact: https://www.djangoproject.com/security/
+Expires: 2026-12-31T00:00:00.000Z
+Encryption: https://keys.openpgp.org/vks/v1/by-fingerprint/AF3516D27D0621171E0CCE25FCB84B8D1D17F80B
```
bmispelon, line 4: For the sake of a future maintainer, it would be nice to document where one could find a link to the most up-to-date key. I was going to suggest a comment above this line, but I'm not sure if that's the best place for such documentation 🤔  (happy to hear other suggestions)

.well-known/security.txt
```diff
@@ -0,0 +1,10 @@
...
+Preferred-Languages: en
+
+# Hello security researcher!
+# We appreciate your help in keeping Django secure.
+# Please report security issues that concern the Django website (djangoproject.com) to ops@djangoproject.com
```
bmispelon, line 9: This seems to contradict the `Policy` and `Contact` directives above which state that one should email `security@djangoproject.com`.

djangoproject/tests.py
```diff
@@ -164,3 +166,40 @@ def test_single_h1_per_page(self):
...
+    Tests for the security.txt file.
+    """
+
+    def test_security_txt_not_expired(self):
+        """
+        The security.txt file should not be expired.
+        """
+        FILE_PATH = settings.BASE_DIR / ".well-known" / "security.txt"
+        with open(FILE_PATH) as f:
+            content = f.read()
```
bmispelon, line 182: Since we have a `Path` object we can directly use:
```suggestion
        content = FILE_PATH.read_text()
```
(and remove a level of indentation for the rest of the method)

djangoproject/tests.py
```diff
@@ -164,3 +166,40 @@ def test_single_h1_per_page(self):
...
+            # Read the line that starts with "Expires:", and parse the date.
+            for line in content.splitlines():
+                if line.startswith("Expires:"):
+                    expires = line.strip("Expires: ")
+                    break
+            else:
+                self.fail("No Expires line found in security.txt")
```
bmispelon, line 189: Perfect use of a `for ... else` 👏🏻  However I think a regex perfect for this kind of situation:
```suggestion
match = re.search("^Expires: (.*)$", content, flags=re.MULTILINE)
            if match is None:
                self.fail("No Expires line found in security.txt")
            else:
                expires = match[1]
```

djangoproject/tests.py
```diff
@@ -164,3 +166,40 @@ def test_single_h1_per_page(self):
...
+
+            expires_date = datetime.strptime(
+                expires,
+                "%Y-%m-%dT%H:%M:%S.%fZ",
+            ).date()
```
bmispelon, line 194: You can use `datetime.fromisoformat()`:
```suggestion
            expires_date = datetime.fromisoformat(expires).date()
            expires_date = datetime.strptime(
                expires,
                "%Y-%m-%dT%H:%M:%S.%fZ",
            ).date()
```

bmispelon (commented):
Thanks for working on this Storm! 🎸 

I've left a few minor stylistic comments inline, but I've also got some more general questions:

1) I originally said that we could use an nginx config to link to the file, but after thinking about it some more I realized that it's going to be more complicated than I'd initially thought, because the `security.txt` is located inside the container. So I think a dedicated view is probably going to be better here.
2) Does the file need to be served under all subdomains (`www`, `docs`, `code`, `dashboard`), or only some of them?
3) As Adam noted, I'm not sure a failing test is the right type of warning for this (though I appreciate that you thought about emitting a warning at all, that's good forward thinking). Did you explore other types of warnings for this and if so, could you walk us through what made you pick a test? On the top of my head, this could also potentially be a system check, a management command that we'd run regularly on the production server, or maybe even some kind of scheduled github action (I know django/django has some of those for example).

I'm happy to brainstorm or chat through things on Discord, or even on a call if you'd prefer. Just let me know.

.well-known/security.txt
sabderemane: I think we could change this with the website WG email: `website-wg@djangoproject.com` to keep the security issues for Django itself separated for the fellows, but the security page need a new line somewhere to mention that security issues related to the website are supposed to be send to this email instead.

sabderemane (commented):
Looks good, thank you @Stormheg ! I added one comment for a small change, including the security page.

.well-known/security.txt
SaptakS: I don't think this file might be a good place for putting that comment, since this file is targetted more towards people who want to report a security issue. It should ideally be there somewhere in a readme for contributors and maintainers. Maybe SECURITY.md (which I am planning on adding)?

djangoproject/tests.py
tut-tuuut: TIL `for ... else`.  🤯 🐍 ❤️ 
Thank you. 🫶

.well-known/security.txt
marksweb: Sounds like something worthy of enabling the repo wiki for?

Stormheg:
Thanks for the reviews so far all 👍 

I'm going to rework the contents based on https://github.com/django/djangoproject.com/pull/2086/

---

@bmispelon 

>  I originally said that we could use an nginx config to link to the file, but after thinking about it some more I realized that it's going to be more complicated than I'd initially thought, because the security.txt is located inside the container. So I think a dedicated view is probably going to be better here.

Alright, will move it into a view.

> Does the file need to be served under all subdomains (www, docs, code, dashboard), or only some of them?

I'd say the file specifically relates to the security of the application running on that domain, so the answer would be: probably, yes.

I'm not sure what the architecture is behind the different subdomains, but if they are all served from this repository we can serve the same file for all of them.

> As Adam noted, I'm not sure a failing test is the right type of warning for this (though I appreciate that you thought about emitting a warning at all, that's good forward thinking). Did you explore other types of warnings for this and if so, could you walk us through what made you pick a test? On the top of my head, this could also potentially be a system check, a management command that we'd run regularly on the production server, or maybe even some kind of scheduled github action (I know django/django has some of those for example).

@bmispelon @adamzap regarding the test case, I was looking for something that is visible and explicit and this is the first thing I came up with. I did not put in any more thought in it than that. What warning mechanism would you prefer to see instead?

> edit: Oh, and also I think the folder should be named well-known: we don't need it to be hidden.

Good catch, lol. I have hidden folders enabled on my mac and forgot that's a convention.

djangoproject/tests.py
Stormheg: Fair point @adamzap – I've downgraded this to a warning instead of a hard fail. This is is still fairly visible and does not rely on someone keeping on track of calendar event.

How does this sit with you?

.well-known/security.txt
Stormheg: Thanks, I've changed the wording in the file to make clear what should be reported where.

djangoproject/templates/well-known/security.txt
```diff
@@ -0,0 +1,27 @@
...
+
+See https://securitytxt.org/ for more information about the security.txt standard.
+{% endcomment %}
+{% endspaceless %}# Hello security researcher!
+# We appreciate your help in keeping Django & djangoproject.com secure.
+
+# Please report security issues that concern this website (djangoproject.com)
+# to the website working group: website-wg@djangoproject.com
+# This helps us make sure your report is directed to the right people.
+# You can find guidelines for reporting website security issues here: https://github.com/django/djangoproject.com/blob/main/.github/SECURITY.md
```
Stormheg, line 16: Note: this url doesn't exist yet. Assumes https://github.com/django/djangoproject.com/pull/2086 is merged.

Stormheg (commented):
@bmispelon @adamzap @sabderemane I've updated the PR with your feedback - thank you

.well-known/security.txt
Stormheg: Since this is now a Django template, we could use a Django template comment so the comment only appears in source code. How does that sound?

djangoproject/templates/well-known/security.txt
```diff
@@ -0,0 +1,27 @@
...
+
+# DO NOT USE security@djangoproject.com FOR ISSUES THAT CONCERN THE WEBSITE.
+
+# If your report concerns Django itself (the Python package, not this website), please follow the Django security reporting process:
+Policy: https://www.djangoproject.com/security/
+Contact: https://www.djangoproject.com/security/
+Expires: 2026-12-31T00:00:00.000Z
```
adamzap, line 23: Should we separate this line and the content below it from the rest of the file? I initially though that the `Expires` line related to reporting a security issue about Django itself.

djangoproject/tests.py
adamzap: That's definitely better! Thanks for the change.

I still think a scheduled email to website-wg@djangoproject.com would be sufficient, but I defer to the majority on this.

djangoproject/templates/well-known/security.txt
```diff
@@ -0,0 +1,27 @@
+{% spaceless %}
+{% comment %}
+This file is served under the well-known URIs
+
+- https://www.djangoproject.com/.well-known/security.txt
+- https://docs.djangoproject.com/.well-known/security.txt
```
ulgens, line 6: Do we need it on code.djangoproject.com and/or other subdomains?

djangoproject/tests.py
Stormheg: @adamzap I've dropped the warning part of the test. I've left this PR open for so long and at this point would rather remove things that could be controversial to get it over the finish line 👍

djangoproject/templates/well-known/security.txt
Stormheg: I'm not sure where the application running code.djangoproject.com lives, but it isn't this repository and it also doesn't fall in scope of the website working group team, right?

Good shout though, worth a discussion.

djangoproject/templates/well-known/security.txt
Stormheg: I don't think that is necessary. The `Expires` line is for the entire document, including the instructions and links to the django security reporting process.

djangoproject/templates/well-known/security.txt
SaptakS: I agree that code.djangoproject.com doesn't need to be part of this document since it is not part of this repo

Stormheg: @ulgens @SaptakS shall we add this as a discussion point for the next meeting and resolve this comment?

I don't know who maintains code.djangoproject.com and should handle security notifications – the operations team I presume?

djangoproject/tests.py
```diff
@@ -211,3 +212,23 @@ def test_single_h1_per_page(self):
...
+
+    def test_security_txt(self):
+        """
+        The security.txt file should be reachable at the expected URL.
+        """
+        response = self.client.get("/.well-known/security.txt")
+        self.assertEqual(response.status_code, HTTPStatus.OK)
+        self.assertEqual(response["Content-Type"], "text/plain")
+
+        match = re.search(
```
adamzap: Could we simplify the "Expires: " checking?

```python
self.assertIn('Expires: ', response.text)
```

Same with the other test.

I think we should at least use an assert method.

Stormheg: That makes perfect sense, thanks for flagging. Did not clean that up after removing the expiration date check.

adamzap (approved):
👍 Looks great!

djangoproject/tests.py
```diff
@@ -211,3 +211,19 @@ def test_single_h1_per_page(self):
                     response = self.client.get(url)
                     self.assertEqual(response.status_code, 200)
                     self.assertContains(response, "<h1", count=1)
+
+
+class SecurityTxtTests(TestCase):
+    """
+    Tests for the security.txt file.
+    """
```
pauloxnet:
```suggestion
    """Tests for the security.txt file."""
```

djangoproject/tests.py
```diff
@@ -211,3 +211,19 @@ def test_single_h1_per_page(self):
...
+
+    def test_security_txt(self):
+        """
+        The security.txt file should be reachable at the expected URL.
+        """
```
pauloxnet:
```suggestion
        """The security.txt file should be reachable at the expected URL."""
```

djangoproject/tests.py
```diff
@@ -211,3 +211,19 @@ def test_single_h1_per_page(self):
...
+        response = self.client.get("/.well-known/security.txt")
+        self.assertEqual(response.status_code, HTTPStatus.OK)
+        self.assertEqual(response["Content-Type"], "text/plain")
+
+        self.assertIn("Expires:", response.content.decode())
```
pauloxnet:
```suggestion
        self.assertIn("Expires:", response.content.decode())
```

docs/tests/test_views.py
```diff
@@ -324,3 +324,19 @@ def test_sitemap_404(self):
...
+
+    def test_security_txt(self):
+        """
+        The security.txt file should be reachable at the expected URL.
+        """
+        response = self.client.get("/.well-known/security.txt")
+        self.assertEqual(response.status_code, HTTPStatus.OK)
+        self.assertEqual(response["Content-Type"], "text/plain")
+
+        self.assertIn("Expires:", response.content.decode())
```
pauloxnet:
```suggestion
    """Tests for the security.txt file."""

    def test_security_txt(self):
        """The security.txt file should be reachable at the expected URL."""
        response = self.client.get("/.well-known/security.txt")
        self.assertEqual(response.status_code, HTTPStatus.OK)
        self.assertEqual(response["Content-Type"], "text/plain")
        self.assertIn("Expires:", response.content.decode())
```

djangoproject/templates/well-known/security.txt
SaptakS: yeah. We dont need to discuss that in this PR

Stormheg: Have added it to the meeting notes for next time 👍 

Will resolve this now.

PR #2072, opened 2025-05-13
Simplied preview environment header check

adamzap:
As a follow-up to discussion in #2052, this patch uses a template-based check to show the preview banner instead of a context processor.

This will need to be tested on the preview environment.


pauloxnet (approved):
LGTM

Issue #1988, opened 2025-03-14
Use CSS instead of JS for floating docs banner


adamzap:
When I refactored the JavaScript for the floating docs banner in #1972, I suggested using a CSS-only solution. I tried unsuccessfully in #1799, but I hope `position: sticky` could still work somehow.

Another approach would be to use `position: fixed`, but we would have to make the content fixed-height. This would be difficult given the amount of text that is currently in the banner element:

<img width="1512" alt="Image" src="https://github.com/user-attachments/assets/fc99277c-3c68-49a4-8521-5a05369f0a29" />

<img width="1512" alt="Image" src="https://github.com/user-attachments/assets/d0965014-bd2d-4e13-8c53-726628e5cee8" />

<img width="516" alt="Image" src="https://github.com/user-attachments/assets/bd80bbfd-0b19-489a-9f5e-7bc81166f753" />

<img width="515" alt="Image" src="https://github.com/user-attachments/assets/8ae5aa0b-97b9-4219-bb1c-7d16613ec931" />

The [current JavaScript approach](https://github.com/django/djangoproject.com/blob/main/djangoproject/static/js/djangoproject.js#L154-L185) feels a bit complex to me and can lead to some odd scroll behavior.

---

I think this (S)CSS should work if we can make the element fixed-height:

```scss
html:has(#dev-warning, #outdated-warning) {
    scroll-padding-top: 38px;
}

body:has(#dev-warning, #outdated-warning) {
    margin-top: 38px;
}

#dev-warning,
#outdated-warning {
    height: 38px;
    position: fixed;
}
```

---

I'd be happy to implement a solution if there is a way forward here.

arpan8925:
Where this floating error visually can be seen? in which page ?

adamzap:
@arpan8925 https://docs.djangoproject.com/en/dev/

arpan8925:
@adamzap  I tried doing a PR with that specific commit, but it just took all my commits, don't know how, need help in that case, also check the last commit, it may fix it

thibaudcolas:
Looking at the code, seems to me a `position:sticky` would work fine! Just needs a `z-index` as well as we would want the banner to be displayed at a higher layer than other elements that are further down in the DOM.

And yes combine with `scroll-padding-top` or `scroll-margin-top`. For the value for those, I don't think it's realistic to make that banner a fixed height as there is so much text in it that will reflow based on viewport width and zoom level. I can see two options to deal with that:

[...]

PR #2073, opened 2025-05-14
Arpan issue #1988

arpan8925:


adamzap:
Hello! It looks like you committed your Python environment. Could you update the branch to contain only changes that are relevant to the issue you are attempting to fix?

arpan8925:
> Hello! It looks like you committed your Python environment. Could you update the branch to contain only changes that are relevant to the issue you are attempting to fix?

I also tried, git checkout to specific commit, then creating a branch and push, still it shows 13 commits while pulling, instead of 1

arpan8925:
Ok may be I got the issue, I pushed my venv as well, git ignoring that :)

PR #2075, opened 2025-05-15
Arpan issue 1988

arpan8925:
May be now its correct


arpan8925:
@adamzap  Can you review kindly !

adamzap:
@arpan8925 Could you squash your commits into one?

arpan8925:
Give me some time, exam is going on, trying this as soon as possible in next 2/3 days

adamzap:
Sure! If you need some guidance to squash the commits, you can do it in at least two ways:

1. Run `git rebase -i HEAD~16` and mark every commit after the first one as "fixup" (`f`), then `git amend` the commit message before force-pushing.
2. Delete your branch locally, look at [the diff view on GitHub](https://github.com/django/djangoproject.com/pull/2075/files), reproduce those changes in one commit locally on a new branch with the same name, then force-push.

Let me know if you have trouble.

.gitignore
```diff
@@ -16,3 +16,5 @@ locale/*/LC_MESSAGES/django.mo
 djangoproject/cache
 djangoproject/static/css/*.map
 djangoproject/static/css/*.css
+venv/
```
marksweb, line 19: Instead of adding these, follow the instructions on the first 3 lines of the file to exclude these because they're unique to you, not related to the project.

marksweb (changes requested):

.gitignore
arpan8925: Ok

arpan8925:
updated @adamzap , review kindly

arpan8925:
Hi @marksweb  Can you check if yours Change Request is resolved?

djangoproject/scss/_style.scss
```diff
@@ -12,6 +12,39 @@ CSS rendered with Libsass 0.7.0
...
+
+body {
+    /* Add a class to the body element when the warning is present */
+    &.has-doc-warning {
+        margin-top: 38px;
+    }
+}
+
+/* Modern browsers that support :has() */
+@supports selector(:has(a, b)) {
```
marksweb, line 28: This doesn't seem to work.

```
❯ make compile-scss
python -m pysassc djangoproject/scss/output.scss djangoproject/static/css/output.css --style=compressed
pysassc.py: error: Error: Invalid CSS after "@supports ": expected @supports condition (e.g. (display: flexbox)), was "selector(:has(a, b)"
        on line 28:10 of djangoproject/scss/_style.scss
        from line 4:1 of djangoproject/scss/output.scss
>> @supports selector(:has(a, b)) {
```

marksweb (changes requested):

djangoproject/scss/_style.scss
arpan8925: Ok Checking

arpan8925: The issue was happening for actually pysassc does not support modern selector-based feature queries like: `@supports selector(:has(#dev-warning)) {`

arpan8925: So I left with 2 Options acutally

Using JavaScript-based detection only
or
Using PostCSS or Dart Sass via a JS or Node build chain instead of LibSass of pysassc .

adamzap: Yeah, unfortunately I think we're stuck with the JavaScript approach for now. Do you agree @marksweb?

adamzap:
@arpan8925 I'm reviewing this as well. Thanks for the changes!

arpan8925:
Most Welcome

adamzap:
@arpan8925 Unfortunately I don't think this will work because your implementation relies on a the warning element being a fixed height (38px). This is the problem I described in #1988.

You can reproduce the problem by making your browser window small enough to make the warning element's text two lines.

arpan8925:
> @arpan8925 Unfortunately I don't think this will work because your implementation relies on a the warning element being a fixed height (38px). This is the problem I described in #1988.
> 
> You can reproduce the problem by making your browser window small enough to make the warning element's text two lines.



May I help somehow? Guide me please

adamzap:
Unfortunately I don't think so. I'll talk to the website team about the possibility of changing the design here.

I think it would make sense to close this pull request for now. Do you agree?

arpan8925:
Yes

adamzap:
Thanks for trying here. It's a confusing issue!

PR #2085, opened 2025-05-30
Added the DSF board meeting link

RISHII-BHARADHWAJ:
I have added the note and the link for the new DSF board meeting. I made changes to 4 files which includes **meeting_archive_month.html** , **meeting_archive_year.html** , **meeting_archive.html** and **meeting_detail.html**


CuriousLearner:
Seems like #2111 also is trying to do the same thing.

adamzap:
@RISHII-BHARADHWAJ Thanks for your contribution, but it looks like was handled in #2111.

PR #2089, opened 2025-06-03
Simplified cache usage in `accounts` view helper

adamzap:
Importing `cache` instead of `caches` uses the default cache: https://docs.djangoproject.com/en/5.2/topics/cache/#django.core.cache.cache

Let me know if the test could be improved. I noticed that the caching behavior wasn't covered.


pauloxnet (approved):
LGTM

Issue #1550, opened 2024-06-16
Make it easier to display a site-wide banner


bmispelon:
There's sometimes a need to add a banner on every page of the site, for example during the biannual pycharm campaign (7a03628d805e0a144d990b5bc2f971cd36901d08).

Currently, this involves adding HTML code to several templates, but it would be nice if this was centralized so that the change would only involve one file (the base template).

sarahboyce:
Note that we have been only updating 1 template for the past few banner updates
Issue https://github.com/django/djangoproject.com/issues/2402 is also on the site-wide banner

ulgens:
About the timing of the banners' publish - unpublish actions, we were having a discussion with @SaptakS about adding feature flags too so we could manage them using the admin and without needing deployments. @bmispelon Do you think we should extend this issue to cover feature flags too or a new one is better?

sarahboyce:
We can reuse this issue 😁 I really don't mind

I agree having something that we can manage from the admin is perfect so that there's less dependence of the PR, review, merge, deployment and allow a quicker turn around time 👍

nessita:
PR at https://github.com/django/djangoproject.com/pull/2586. This work was intentionally kept to a minimum to 1. make it feasible to develop, 2. get real-life usage feedback to fine-tune, 3. have this ready in time for incoming campaigns

PR #2092, opened 2025-06-11
Added Fellow call for applicant banner.

sarahboyce:
Homepage:

image

Documentation:

Screenshot from 2025-06-11 11-24-06

------

Note that we can put it in more places but I think this is a reasonable start
This idea came from @knyghty when I mentioned that blog articles can get "lost" if we have several posts

### DIFF
--- a/docs/templates/base_docs.html
+++ b/docs/templates/base_docs.html
@@ -10,7 +10,15 @@
 {% endblock %}
 
 {% block header %}
-  <p><a href="{% block doc_url %}{% url 'homepage' %}{% endblock %}">{% trans 'Documentation' %}</a></p>
+  {% comment %}
+    <p><a href="{% block doc_url %}{% url 'homepage' %}{% endblock %}">{% trans 'Documentation' %}</a></p>
+  {% endcomment %}
+  <p>
+    <a href="{% block doc_url %}{% url 'homepage' %}{% endblock %}">{% trans 'Documentation' %}</a> &ndash;
+    <a href="https://www.djangoproject.com/weblog/2025/jun/09/django-fellow-applicants-2025/">
+      DSF calls for applicants for a Django Fellow, deadline July 1!
+    </a>
+  </p>
   {% search_form %}
 {% endblock %}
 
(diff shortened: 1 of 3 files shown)


docs/templates/base_docs.html
```diff
@@ -10,7 +10,15 @@
 {% endblock %}
 
 {% block header %}
-  <p><a href="{% block doc_url %}{% url 'homepage' %}{% endblock %}">{% trans 'Documentation' %}</a></p>
+  {% comment %}
+    <h1><a href="{% block doc_url %}{% url 'homepage' %}{% endblock %}">{% trans 'Documentation' %}</a></h1>
```
alexgmin: I'm assuming this should have been an `<h1>` from the start instead of a `<p>`?

sarahboyce: Well spotted, it should be a `<p>` tag. Bad copy and paste from a previous commit :+1:

SaptakS:
I feel putting the call for applicant link in the same space as "Documentation" with a dash might not be the best UI (I personally feel I would be confused, since I would think that the text after dash usually would denote an internal page title).
Maybe something more like this makes sense? Thoughts?:
image

sarahboyce:
> I feel putting the call for applicant link in the same space as "Documentation" with a dash might not be the best UI (I personally feel I would be confused, since I would think that the text after dash usually would denote an internal page title). Maybe something more like this makes sense? Thoughts?: image

This works!
I was copying what we have done previously for the [PyCharm banner](https://github.com/django/djangoproject.com/commit/d031005647b0c7cca98824a898ac31d218f52be2).

We do occasionally want to have this banner/CTA encouraging folks to do something (apply, donate, survey etc) so if you want to improve the design, I'm happy for you to suggest/push changes

I am hoping we put this in place by the end of the week :+1:

sabderemane:
Hey @sarahboyce  🤗 
I join @SaptakS 's feedback and I have few additional feedbacks but I think it might better to provide them in [the issue to ease the addition of a banner](https://github.com/django/djangoproject.com/issues/1550) for every case, so we can merge this one to have it ASAP since the announcement was quite recent and we can improve the design and code for the PR in another PR 👌🏽 

Does it render well in dark theme?

SaptakS:
@sabderemane makes sense to me. 

> Does it render well in dark theme?

The background color doesn't change unlike the copy-banner, but renders fine apart from that.

PR #2093, opened 2025-06-11
Speed up initial docs generation by setting the git clone depth to 1.

alexgmin:
For the docs generation on a new setup, which is when the .git folder doesn't exist, there's no need to clone the entire repo history. 
Setting the clone depth to 1 and just pulling the last commit, speeds the initial clone process quite a bit.

This shouldn't change anything for existing docs setups.

### DIFF
--- a/docs/management/commands/update_docs.py
+++ b/docs/management/commands/update_docs.py
@@ -358,7 +358,17 @@ class Command(BaseCommand):
                 os.chdir(cwd)
         else:
             subprocess.check_call(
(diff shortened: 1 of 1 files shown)


pauloxnet (approved):

marksweb (approved):

bmispelon (approved):
Tried it locally and it made `update_docs` on a clean checkout go from 2min45s to 1min58s.

Nice work 👍🏻

PR #2097, opened 2025-06-15
Make secrets.json not being found an explicit error when DJANGOPROJECT_DATA_DIR is set

alexgmin:
While helping someone setup the website in their system, they faced an issue in which the `secrets.json` file wasn't found, and they had `DJANGOPROJECT_DATA_DIR` env var set (but the `secrets.json` was on a different directory).

The instructions tell you to have this on your bashrc/zshrc export `DJANGOPROJECT_DATA_DIR=~/.djangoproject`, and if set, expect a `~/.djangoproject/conf/secrets.json`.
In fact there's a fallback for that but only in the env variable is not set, which the instructions never tell you about.

Removing the `DJANGOPROJECT_DATA_DIR` would require coordination with operations to make sure it doesn't break their setup, so for now I'd just make it error explicitly if the `secrets.json` file is not found, which is a good initial step.

I think there's a case in which this breaks an existing working local setup, which requires the following conditions:
1. The env var `DJANGOPROJECT_DATA_DIR` is set to a path that doesn't contain `secrets.json`. Note that the website can work without `DJANGOPROJECT_DATA_DIR` being set at all.
2. The local postgres is set up to use **exactly** the defaults set in `DATABASES`. Any difference will mean it will raise an error when trying to execute Django.

### DIFF
--- a/djangoproject/settings/common.py
+++ b/djangoproject/settings/common.py
@@ -14,15 +14,8 @@ DATA_DIR = (
     Path(os.environ[data_dir_key]) if data_dir_key in os.environ else BASE_DIR.parent
 )
 
-try:
-    with (DATA_DIR / "conf" / "secrets.json").open() as handle:
-        SECRETS = json.load(handle)
-except OSError:
-    SECRETS = {
-        "secret_key": "a",
-        "superfeedr_creds": ["any@email.com", "some_string"],
-    }
-
+with (DATA_DIR / "conf" / "secrets.json").open() as handle:
+    SECRETS = json.load(handle)
 
 # Django settings
 


adamzap (approved):
👍 Looks great!

ulgens (approved):

pauloxnet (approved):
LGTM

SaptakS:
@alexgmin can you resolve the conflicts? We can go ahead and merge this once conflicts are resolved.

alexgmin:
I've rebased the PR on main.

However, something I noticed while doing the setup on a new computer is that now the docker setup, fails because it doesn't find the file. 

Since now we set all of the env variables in the docker compose, I think the correct change for the docker setup should be to pick up the variables from the file is set, and if not, just continue silently.

However that means a local setup that doesn't have the file won't error, which defeats the purpose of the change, so I'm not sure how we can easily fix it other than doing 

```
try:
    with (DATA_DIR / "conf" / "secrets.json").open() as handle:
        SECRETS = json.load(handle)
except FileNotFoundError:
    SECRETS = {
        "SECRET_KEY": os.environ.get("SECRET_KEY"),
        "DJANGOPROJECT_DATA_DIR": os.environ.get("DJANGOPROJECT_DATA_DIR"),
        "DJANGO_SETTINGS_MODULE": os.environ.get("DJANGO_SETTINGS_MODULE"),
        "SQL_ENGINE": os.environ.get("SQL_ENGINE"),
        "SQL_DATABASE": os.environ.get("SQL_DATABASE"),
        "SQL_USER": os.environ.get("SQL_USER"),
        "SQL_PASSWORD": os.environ.get("SQL_PASSWORD"),
        "SQL_HOST": os.environ.get("SQL_HOST"),
        "SQL_PORT": os.environ.get("SQL_PORT"),
        "TRAC_DATABASE": os.environ.get("TRAC_DATABASE"),
        "TRAC_USER": os.environ.get("TRAC_USER"),
        "TRAC_PASSWORD": os.environ.get("TRAC_PASSWORD"),
        "TRAC_HOST": os.environ.get("TRAC_HOST"),
        "TRAC_PORT": os.environ.get("TRAC_PORT"),
    }
```

The benefit of this is that we can avoid all of these settings in the docker.py, I think.
https://github.com/django/djangoproject.com/blob/b32fcd734e70aaec11d1ebc9396409022a3e463d/djangoproject/settings/docker.py#L3-L12

I can change the PR towards this approach if it makes sense.

SaptakS:
I don't know if it makes sense so tell me if I am wrong, but building on top of the code you wrote, can we add a conditional statement in the `except` block like if `os.environ.get("SECRET_KEY")` (or something else?) is not set, then we throw an error that secrets.json is missing?

alexgmin:
I think that could work. 
So basically the flow goes from
1. secrets.json
2. environment variables (with realistically are only filled in the docker case)
3. Error

The question I have now is if in this PR I should remove the code from the docker.py since it's covered in 2.
I wouldn't remove the file, but it would just be

```
from .dev import *  # noqa: F403


ALLOWED_HOSTS = [".localhost", "127.0.0.1", "www.127.0.0.1"]
```

And technically those hosts are docker unrelated, I think, so they could be in the dev.py

SaptakS:
The hosts have now been removed also from docker.py, so I think in that case makes sense to remove the file. @ulgens does that make sense?

ulgens:
@SaptakS I couldn't quite catch your question, sorry.

About the docker related problems, I think the root issue here is that we are trying to manage secrets with a json file, instead of using the env. Conversion of `secrets.json` to `.env` file would help us to move away from several outdated practices.

Apart from the side discussion, I think the PR is good to merge. It would be nicer if it weren't breaking the CI, but I'm okay with handling it in a different PR.

sabderemane:
Going with @alexgmin 's approach, the `docker.py` seems to be not used anymore as @SaptakS was saying. Can we remove it before to merge this PR?

PR #2101, opened 2025-06-17
Add link to 'Stable' doc version in the Warning Alert on top

### LINKED ISSUE #2094: Add link to 'Stable' doc version in the Warning Alert on top
As discussed on discord in #documentation channel [here](https://discord.com/channels/856567261900832808/1249640097339539478/1378635016174571520)

I think this will provide 'further convenience' to website users to upgrade the documentation version by simply clicking on the link in the warning itself as well:

**Current Behavior** of the Warning Alert on Top:

Image

**Proposed change** (text can be adjusted accordingly):

Image

Rationale:
- The warning alert already exists, and draws eye attention towards itself (due to its bright color)
- Adding a clickable link is 'further convenience' in terms of UX
- Requires small change in code

ulgens:
Resolves https://github.com/django/djangoproject.com/issues/2094

CleanShot 2025-06-17 at 16 23 53@2x

I copied the link from "Installation guide" link on the sidebar, please let me know if we prefer a different target for the link.

CleanShot 2025-06-17 at 16 28 53@2x

I'm open to suggestions about wording and styling.

### DIFF
--- a/docs/templates/docs/doc.html
+++ b/docs/templates/docs/doc.html
@@ -47,7 +47,11 @@
     </div>
   {% elif not release.is_supported %}
     <div id="outdated-warning" class="doc-floating-warning">
-      {% trans "This document is for an insecure version of Django that is no longer supported. Please upgrade to a newer release!" %}
+      {% trans "This document is for an insecure version of Django that is no longer supported. Please click "%}
+      <a href="{% url 'document-detail' lang='en' version='stable' url='intro/install' host 'docs' %}">
+        {% trans "here" %}
+      </a>
+      {% trans " for the latest version."%}
     </div>
   {% endif %}
 {% endblock body_extra %}


AyushPatiTripathi:
@ulgens 

Thanks for pushing this PR! Since the main logic is in place and you're open to someone continuing the work, I'd love to take it forward and handle the remaining styling updates.

ulgens:
@SomeAB @sabderemane Do you have any recommendations for styling? I think the default styling for this link looks a bit weird, but I'm not sure what would be the correct way.

SaptakS:
- Styling: I would not even make it a separate color. I would probably leave it the same color with underline. That is a good enough styling for inline links.
- Wording: I would word the sentence something like "Please check the latest version.", with the "latest version" anchor linked. "here" is usually not a good word to link.

SomeAB:
@ulgens Please feel free to ping me on discord or here. I'm more active on discord. Sorry for the delay in reply

A. Regarding the **wording**:  I would personally recommend: "Please click here for the stable version"

**Rationale:** saying 'stable' instead of 'latest' is more inline with the previous sentence, which points at the depreciation of the doc version currently being viewed. Implicitly, we are recommending the latest LTS (link should point to LTS version given its long term support)

Two ways of highlighting: `'here'` or <ins>`here`</ins>

B. Regarding **Color**: I agree with SaptakS here, that making it as the same color (dark) is preferable. 

**Rationale:** Given the Django website uses an unusual combo of **peach-pink** with **leaf-green**, introducing new colors would just stick out badly. The highlight is suppose to be just a small nudge .. just enough to tell that it is clickable

____________________
Preview of two versions:

1. 
>This document is for an insecure version of Django that is no longer supported. Please click 'here' for the stable version.

2.
>This document is for an insecure version of Django that is no longer supported. Please click <ins>here</ins> for the stable version.

sabderemane:
I agree with Saptak for the styling and the wording. The wording "Please check the latest version.", with the "latest version" anchor linked is better. 
The link "here" isn't meaningful on it's own. Almost everywhere on the website, links are underline with a slightly change on hover, for consistency it's better to have something similar.

docs/templates/docs/doc.html
```diff
@@ -47,7 +47,11 @@
     </div>
   {% elif not release.is_supported %}
     <div id="outdated-warning" class="doc-floating-warning">
-      {% trans "This document is for an insecure version of Django that is no longer supported. Please upgrade to a newer release!" %}
+      {% trans "This document is for an insecure version of Django that is no longer supported. Please click "%}
```
marksweb, line 50: From a translation perspective this might be better using `{% blocktranslate %}` because as it stands, "here" and " for the latest version" in the translation file lack context.

I think you'd end up with something like this;

```
{% url 'document-detail' lang='en' version='stable' url='intro/install' host 'docs' as install_url %}

{% blocktranslate trimmed %}
<p>
This document is for an insecure version of Django that is no longer supported. Please click <a href="{{ install_url }}">here</a> for the latest version.
</p>
{% endblocktranslate %}
```

marksweb (changes requested):

PR #2102, opened 2025-06-18
Add link to latest documentation in version warning banner

AyushPatiTripathi:
Ticket: [#2904]

Before (Current Behavior)
The warning banner at the top of the documentation pages only notifies the user that they’re viewing an outdated version, without offering a direct link to the latest version.

After (Proposed Enhancement)
The warning banner now includes a clickable link that takes the user directly to the latest documentation version, providing:

Better user experience by reducing friction.

A simple, visible, and actionable way to access the updated docs.

 Implementation Notes
Updated docs.html to include an anchor tag in the warning banner.

Used Django template syntax to dynamically resolve the correct "latest" version link.

Screenshot From 2025-06-18 14-29-55

### DIFF
--- a/docs/templates/docs/doc.html
+++ b/docs/templates/docs/doc.html
@@ -38,7 +38,10 @@
   {% if release.is_dev %}
     {% if "internals" not in docurl %}{# The dev version is canonical for internals/. #}
       <div id="dev-warning" class="doc-floating-warning">
-        {% trans "This document is for Django's development version, which can be significantly different from previous releases. For older releases, use the version selector floating in the bottom right corner of this page." %}
+      {% blocktrans trimmed %}
+      This document is for Django's development version, which can be significantly different from previous releases.
+        <a href="https://docs.djangoproject.com/en/stable/" style="font-weight : bold; text-decoration: underline;"> Click here for the stable version </a>
+        {% endblocktrans %}
       </div>
     {% endif %}
   {% elif release.is_preview %}


AyushPatiTripathi:
@ulgens please have a look on this new PR focuing on the original issue

ulgens:
@AyushPatiTripathi Why did you prefer the static link over the dynamic one? This won't work with local development env and any non-prod testing env.

AyushPatiTripathi:
@ulgens  thanks for pointing it out  your are right  i initially chose the static link  for simplicity  but i agreeit breaks compatibility with local and staging envirnments . i will update  to use a relative link (/en/stable/) so it works across all envirnments  .

ulgens:
@AyushPatiTripathi Can you please share what is your plan with this implementation? I think it will be easier to move forward with a plan, otherwise it's hard to understand what is the goal with the changes.

AyushPatiTripathi:
The goal  is to enhance the UX by making the warning alert (which appears when users are not on the latest version of the documentation) more actionable.

Currently, the alert notifies users but doesn't help them navigate directly to the stable version.

I proposed adding a clickable link within the alert that takes users straight to the stable docs version, making it easier for them to upgrade — especially first-time visitors or newcomers.

I have minimize the diff and update the implementation using a dynamic {% url %} tag as suggested.
Let me know if this aligns with what you'd like to see.

ulgens:
@AyushPatiTripathi Thanks for the updates; the code looks better now. I'll try to review it on the weekend. Meanwhile, can you please update the screenshot in the PR description?

PR #2107, opened 2025-06-23
Fix icon color in light/dark modes

frankwiles:
- Icon had hard coded color for dark mode and looked all wrong (grey) in light mode.


frankwiles:
Fixes #2108

bmispelon:
Thanks for the fix @frankwiles ! ✨

While it's not a big deal for this PR (the change was fairly trivial after all, though I wonder if some other icons might be impacted by a similar issue), please don't merge your own pr without a review next time (ideally from the website working group).

frankwiles:
Yeah I figured it was ok since it was such a small change, but noted.   

This was an icon I added as part of the Package Ecosystem page and not sure if *I* set the color specifically in that PR or just neglected to change it to `currentColor`.  In any case, the other icons are all set `full="currentColor"` already so should be zero impact to them.

PR #2110, opened 2025-06-24
Fixed name of DjangoCon Europe youtube channel

bmispelon:


SaptakS:
I wonder if both the youtube channels should be labelled with full "DjangoCon" instead of "DC".

bmispelon:
> I wonder if both the youtube channels should be labelled with full "DjangoCon" instead of "DC".

I've pushed another commit with some alternate wording, let me know what you think

SaptakS (approved):
Looks good to me. Merging.

Issue #1890, opened 2025-01-21
Do we really need tox?


ulgens:
While working with https://github.com/django/djangoproject.com/pull/1792, I realized that tox setup [has some issues](https://github.com/django/djangoproject.com/issues/1817), and I tried to understand what is going on. Apparently, there is only one version of Python and Django that we test againist, and the tox config doesn't look like it does more than chaining tests and linting. tox is not needed for Django tests and the rest can be handled by pre-commit.

adamzap:
I think it'd be great if we could remove it.

I wonder if `tox` is useful for the pattern seen in these two commits where a new Python version is added to test against along side the old one before the old one is removed:

- 164c84027f725015697dbf5f7e6f08ae5dd9dcf7
- 70895c194e5e300e8130a6353a8af395f8177786

I'm looking forward to hear what others think!

ulgens:
@adamzap For those cases, wouldn't it be okay just to test in on a new branch?

adamzap:
It should be. I personally think that would be ideal, but I don't know the current process.

pauloxnet:
I also don't think we need tox here. 
Maybe we can open a draft PR to see all the required changes to get rid of it.

tobiasmcnulty:
I think upgrading without tox will be fine, but it'd be nice to leave in place the other mechanisms for supporting 2+ Python versions at the same time. I left a comment on the PR: https://github.com/django/djangoproject.com/pull/2135#pullrequestreview-3058066493

ulgens:
@tobiasmcnulty The discussion under https://github.com/django/djangoproject.com/pull/2135#discussion_r2232902387 was getting a bit too long for a review comment, let's continue here 🌻 

There are several discussion points, so I'll share my thoughts and answers in sevaral parts. Please let me know if I'm mistaken or missing anything.

---

The website is currently running with Python 3.12. ~As far as I understand, we don't have a configuration pointing to a production env's Python version in the repository, which is something we need to improve~ *. However, both the README and the Dockerfile refer to 3.12.

\* This was a misunderstanding on my end.

---

https://github.com/django/djangoproject.com/pull/1640

The PR adding support for 3.12 claims that it is "optional support". I was looking to understand what this means, and I saw

> I deployed this to both the production (Python 3.8) and preview (Python 3.12).

[...]

PR #2134, opened 2025-07-22
Clean tox config

ulgens:
Relates to #1890

Removes black, flake8, and isort from tox config. All of them are already handled by pre-commit.

About the new action;

I am a bit hesitant about using https://pre-commit.ci/ for pre-commit checks, for several reasons:
* It doesn't allow selecting which Python version to run pre-commit on
* Auto-fix commits create "fix typo" type of small - meaningless commits, which I believe is a bad practice.
* It's an external service that duplicates the functionality of GitHub Actions, but only for pre-commit.
* Has no configuration info in the repository

My vote here would be to disable that integration and continue with the action. I can revert the new action if the consensus is to not change it, but I wanted to include it because the cleaning in the PR makes the CI more dependent on pre-commit.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,16 @@
+name: pre-commit
+
+on:
+    pull_request:
+    push:
+        branches: [main]
+
+jobs:
+    pre-commit:
+        runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


SaptakS (approved):
Makes sense to me.

Issue #1901, opened 2025-01-24
pyproject.toml Migration


ulgens:
_I feel like I'm abusing the issues feature but I hope it's okay, considering we don't have discussions enabled._

Is there a plan to switch to a pyproject.toml file? I think it's important because
* It gives the impression that the project is modern. If not for anything else, I believe that it's important that projects under django organization should look modern and up to date.
* Easier to use and manage. Especially with uv, almost everything is collected under a single tool and single configuration file.

and that's all 🙂 I was hoping to write something better, but I think that development ergonomics are important enough for a change like this one.

ulgens:
There is a [template project](https://github.com/ulgens/django-blasphemy) that I test these types of practices on it first. I studied the poetry - uv migration on that repo and I'm happy with the results: https://github.com/ulgens/django-blasphemy/pull/603/files If that seems okay, I can adapt a similar pattern on this repo too.

alexgmin:
Just want to add that this tool should make it easy to migrate to uv from the current setup.
https://github.com/mkniewallner/migrate-to-uv

ulgens:
@django/django-website How do we feel about this?

@django/ops-team Would this migration require any changes in the deployment proces?

tobiasmcnulty:
Hi @ulgens,

> [@django/ops-team](https://github.com/orgs/django/teams/ops-team) Would this migration require any changes in the deployment proces?

I think migrating to pyproject.toml + uv for managing dependencies is a great idea. Astral has some good [uv Docker examples](https://github.com/astral-sh/uv-docker-example) we could follow to update the Docker image.

Provided `python -m django`, `make`, and `gunicorn` behave the way the do now inside the container, it shouldn't require changes in the deployment process. It doesn't look like paths are hard-coded (with respect to the website image at least, different story for the Trac image), the playbook merely assumes `python -m django` and `gunicorn` are available in the `PATH`, and `make compile-scss` can be executed from the current working directory.

> Just want to add that this tool should make it easier to migrate to uv from the current setup. https://github.com/mkniewallner/migrate-to-uv

I tried this for another project, but found it harder to use than uv's own built-in tools. Disclaimer and a shameless plug: I wrote about [migrating to uv](https://www.caktusgroup.com/blog/2025/06/11/migrating-python-django-projects-uv/) recently.

Good luck & have fun!

Issue #1574, opened 2024-07-01
Upgrade Python version to 3.12


bmispelon:
Some of our dependencies have started dropping support for Python 3.8 (the one we're currently running in production).

I don't think there's much to be done in this project's codebase (this is mostly an ops issue), but I wanted to have an issue where I can track which dependencies are being affected:

- `docutils` (+ `Sphinx`): #1571
- `pre-commit`: #1568
- `watchdog`: #1617

PR #2135, opened 2025-07-22
Remove tox

ulgens:
Resolves:
- #1817 
- #1890 

Depends on:
- #2134


.github/workflows/tests.yml
```diff
@@ -10,12 +10,6 @@ on:
 jobs:
     tests:
         runs-on: ubuntu-24.04
-        strategy:
-            fail-fast: false
-            matrix:
-                # tox-gh-actions will only run the tox environments which match the currently
-                # running python-version. See [gh-actions] in tox.ini for the mapping.
-                python-version: ["3.12"]
```
tobiasmcnulty, line 18: My preference would be to leave the matrix strategy in place, remove the comment about `tox-gh-actions`, and add a comment to say that new Python versions can be added here when upgrading to support both side by side.

tobiasmcnulty (commented):
Thanks for cleaning this up! I have not tested, but I think removing tox will help simplify things. One nitpick.

.github/workflows/tests.yml
ulgens: Hey @tobiasmcnulty , this was discussed in https://github.com/django/djangoproject.com/issues/1890#issuecomment-2605572403 . Because we don't need to support multiple versions at the same deployment, Python version updates and their testing / QA processes can be handled in their own branches. I see no harm in keeping the matrix but I also don't see the benefit.

ulgens: Btw, because of Trac, I don't think we will be able to update the Python version any time soon. This was one of the major points for discussions about moving away from Trac.

tobiasmcnulty: Hi @ulgens, if I recall, upgrading the website separately from Trac is the benefit of the matrix strategy.

I'm not very familiar with the dependencies between the two repos, but perhaps we should add `python3.11` to the matrix now, since that's what Trac is using? Maybe you have some insight on this @bmispelon?

When I added 3.12 support for the website, a few requirements had to [remained pinned to the older versions](https://github.com/django/djangoproject.com/issues/1574#issuecomment-2389742140) I believe so as not to break the Trac integration.

ulgens: https://github.com/django/djangoproject.com/issues/1890#issuecomment-3122514625

.github/workflows/tests.yml
```diff
@@ -38,7 +32,8 @@ jobs:
                   python-version: "${{ matrix.python-version }}"
```
bmispelon: Isn't that going to be broken now that you've removed the version matrix?

ulgens: >  The `python-version` input is not set.  The version of Python currently in `PATH` will be used.

It seems Github was automatically picking up Python 3.12.3 and doesn't fail, and I missed that.

I pushed a fix to specify 3.12 explicitly.

ulgens:
>  I think those should be removed.

You are right, I pushed an update.

> what will be the new single source of truth when it comes to the Python version if we remove the matrix?

For the test pipeline, we can continue to use the same `python-version` argument with a specific Python version as the input. In case of a migration to uv / pyproject.toml (https://github.com/django/djangoproject.com/issues/1901) uv Python installer has the same argument: https://github.com/astral-sh/setup-uv?tab=readme-ov-file#python-version

bmispelon (changes requested):
There is still a reference to `matrix.python-version` which should be removed:
```
COVERALLS_FLAG_NAME: "${{ matrix.python-version }}"
```
(line 64)

Other than that I think this is ready to merge, nice work 👍🏻

adamzap:
Do we still use Coveralls? https://coveralls.io/github/django/djangoproject.com

ulgens:
> There is still a reference to `matrix.python-version` which should be removed:

Yep. Pushed another fix.

bmispelon (approved):

bmispelon:
> Do we still use Coveralls? https://coveralls.io/github/django/djangoproject.com

No idea 🤷🏻 I know I don't, but I don't know when that was added and/or why. It's possible it doesn't make sense anymore.

SaptakS (approved):
Looks good to me.

PR #2136, opened 2025-07-23
Support of including blog in the website search

tim-schilling:
The blog results should have a property of whether it is included in the search results. We should also limit the blogs that are searchable for a version of Django based on the support end. This will allow us to limit the inclusion of blog posts in the search based on the time the entry was created, keeping the search results relevant to that version of Django.

This is a WIP yet. I'm opening the draft PR to solicit early feedback on the approach. I'm thinking we'll re-use the `DocumentationCategory.WEBSITE` in place of "weblog".


tim-schilling:
I was given tacit approval (lack of pushback) from the website WG to use this approach. I plan on working on this at the DjangoCon Africa sprints (Aug 14 & 15), but if anyone wants to complete it before that, please do!

tim-schilling:
I'm going to try to clean up the commits shortly here.

tim-schilling:
Cleaned up the commits. Please let me know if any other changes are necessary.

blog/migrations/0008_alter_entry_is_searchable.py
adamzap, line 1: Could you explain why you have created three migrations for this field?

tim-schilling: Technically you're supposed to have these three migrations to support adding a non nullable column to a table that has data and supporting backwards migrations. It covers any possibility of an operational error during deployment.

tim-schilling: If you meant the technical details justifying it, please let me know. I may have misunderstood your question.

adamzap: Thanks for the explanation. I'm fine with it if you and others prefer it. It seems overkill to me, but I may be missing something.

Is this approach mentioned in the docs somewhere? I'm having trouble finding it.

tim-schilling: Yeah no problem. Second paragraph from https://docs.djangoproject.com/en/5.2/topics/migrations/#data-migrations

> Migrations that alter data are usually called “data migrations”; they’re best written as separate migrations, sitting alongside your schema migrations. 

And then from the [RunPython docs](https://docs.djangoproject.com/en/5.2/ref/migration-operations/#runpython)

> On databases that do support DDL transactions (SQLite and PostgreSQL), RunPython operations do not have any transactions automatically added besides the transactions created for each migration. Thus, on PostgreSQL, for example, you should avoid combining schema changes and RunPython operations in the same migration or you may hit errors like OperationalError: cannot ALTER TABLE "mytable" because it has pending trigger events.

adamzap: Oh, yeah. It makes sense to me that the data migration is separate from the other ones.

What I've been referring to is why this sequence of migrations:

1. `0006` => Create nullable field
2. `0007` => Set field to `False` in a data migration
3. `0008` => Set `default=False` and remove nullability

Can't be a single one like this:

1. `0006` => Create field with `default=False` (the current state of the field in `models.py`)

tim-schilling: Ohhh. Thanks for clarifying that. I guess I landed on wanting to avoid any downtime, but the number of blog posts is probably pretty small. Any queries blocked by the table lock to create that column shouldn't be blocked for too long.

docs/models.py
```diff
@@ -212,6 +221,83 @@ def sync_to_db(self, decoded_documents):
...
+            synced_views = [
+                # Page title, url name, url kwargs
+                ("Django's Ecosystem", "community-ecosystem", {}),
+            ]
+            for title, url_name, kwargs in synced_views:
+                absolute_url = reverse(url_name, kwargs=kwargs, host="www")
+                path = reverse_path(url_name, kwargs=kwargs)
+                request = RequestFactory().get(path, HTTP_HOST=www_host)
+                body = resolve(path).func(request).render().text
+                # Need to parse the body element.
```
tim-schilling: Forgot this TODO yet.

blog/models.py
```diff
@@ -126,6 +131,13 @@ class Entry(models.Model):
         ),
         default=False,
     )
+    is_searchable = models.BooleanField(
+        default=False,
+        help_text=_(
+            "Tick to make this entry allow this entry to appear in the "
+            "Django documentation search."
```
bmispelon: This doesn't seem quite right

docs/tests/test_models.py
```diff
@@ -477,8 +495,43 @@ def test_sync_to_db(self):
...
+        """
+        Releases must be English to include the blog and website results in search.
+        """
+        non_english = DocumentRelease.objects.create(
+            lang="es",
+            release=Release.objects.create(version="88.0"),
+            support_end=self.release.support_end,
+        )
+        non_english.sync_to_db([])
+        self.assertFalse(non_english.documents.exists())
```
bmispelon: Using `assertQuerySetEqual()` would yield a better failure message were the assertion to fail:

```suggestion
        self.assertQuerySetEqual(non_english.documents.all(), [])
```

docs/tests/test_models.py
```diff
@@ -465,7 +469,21 @@ def test_search_title(self):
...
+        cls.entry = Entry.objects.create(
+            pub_date=now,
+            is_active=True,
+            is_searchable=True,
+            headline="Searchable post",
+            slug="a",
+            body_html="<h1>Searchable Blog Post</h1>",
+        )
+        cls.docs_documents = cls.release.documents.exclude(
+            metadata__parents=DocumentationCategory.WEBSITE.value
```
bmispelon: Is `.value` needed here? My understanding of enums is that you can use them directly in place of their values, but I could be wrong 🤔 

```suggestion
            metadata__parents=DocumentationCategory.WEBSITE
```

docs/tests/test_models.py
```diff
@@ -477,8 +495,43 @@ def test_sync_to_db(self):
...
-        self.assertEqual(document.path, "foo/bar")
+        document_paths = set(self.release.documents.values_list("path", flat=True))
+        self.assertEqual(
+            document_paths,
+            {
+                "foo/bar",
+                reverse("community-ecosystem", host="www"),
+                self.entry.get_absolute_url(),
+            },
+        )
```
bmispelon: Can we use `assertQuerySetEqual` here instead of set comparison?

```suggestion
        self.assertQuerySetEqual(
            self.release.documents.all(),
            ["foor/bar", ...],
            ordered=False,
            transform=attrgetter("path"),
        )
```

docs/tests/test_models.py
```diff
@@ -562,3 +615,47 @@ def test_excluded_documents(self):
...
+                "release": cls.release,
+                "title": "Index",
+            },
+        ]
+        # Include the static views in the document search
+        cls.release._sync_views_to_db()
+        Document.objects.bulk_create(Document(**doc) for doc in documents)
+        cls.document_index, cls.document_view, cls.document_detail = (
+            cls.release.documents.order_by("path")
+        )
```
bmispelon: This will break if/when more views are added to `_sync_views_to_db()`, right?

I'd suggest breaking the single `test_document_url()` method into three separate tests then.

docs/templates/docs/search_results.html
```diff
@@ -43,11 +43,11 @@ <h2>{% translate "No search query given" %}</h2>
         {% for result in page.object_list %}
           <dt>
             <h2 class="result-title">
-              <a href="{% url 'document-detail' lang=result.release.lang version=result.release.version url=result.path host 'docs' %}{% if not start_sel in result.headline %}{{ result.highlight|fragment }}{% endif %}">{{ result.headline|safe }}</a>
+              <a href="{{ result.get_absolute_url }}{% if not start_sel in result.headline %}{{ result.highlight|fragment }}{% endif %}">{{ result.headline|safe }}</a>
```
bmispelon, line 46: This is a nice cleanup! 👍🏻

docs/search.py
```diff
@@ -65,6 +65,7 @@ class DocumentationCategory(TextChoices):
     TOPICS = "topics", _("Using Django")
     HOWTO = "howto", _("How-to guides")
     RELEASE_NOTES = "releases", _("Release notes")
+    WEBSITE = "weblog", _("Django Website")
```
bmispelon: I'd change the value to keep it consistent with the variable name (the goal is to eventually index all kinds of pages and not just the blog, right?):

```suggestion
    WEBSITE = "website", _("Django Website")
```

docs/models.py
```diff
@@ -224,7 +310,9 @@ def _clean_document_path(path):
 
 
 def document_url(doc):
-    if doc.path:
+    if doc.metadata.get("parents") == DocumentationCategory.WEBSITE.value:
```
bmispelon: Doesn't it work without the `.value` too?

```suggestion
    if doc.metadata.get("parents") == DocumentationCategory.WEBSITE:
```

docs/models.py
```diff
@@ -212,6 +221,83 @@ def sync_to_db(self, decoded_documents):
...
+                        self.lang[:2], DEFAULT_TEXT_SEARCH_CONFIG
+                    ),
+                )
+
+    def _sync_views_to_db(self):
+        """
+        Sync the blog entries into search based on the release documents
+        support end date.
+        """
+        if self.lang == "en":
```
bmispelon: I'd switch to an early return to reduce the indentation level here:
```
if self.lang != "en":
    return  # TODO: explain why here
```

docs/models.py
```diff
@@ -212,6 +221,83 @@ def sync_to_db(self, decoded_documents):
...
+        self._sync_views_to_db()
+
+    def _sync_blog_to_db(self):
+        """
+        Sync the blog entries into search based on the release documents
+        support end date.
+        """
+        if self.lang == "en" and self.support_end:
+            for entry in Entry.objects.published(self.support_end).searchable():
+                Document.objects.create(
```
bmispelon: Can we use `bulk_create()` instead of creating individual `Document` objects?

blog/models.py
```diff
@@ -31,12 +31,17 @@ def _md_slugify(value, separator):
 
 
 class EntryQuerySet(models.QuerySet):
-    def published(self):
-        return self.active().filter(pub_date__lte=timezone.now())
+    def published(self, pub_date=None):
```
bmispelon: I'd go for a more explicit naming of the argument, what about `published_as_of`?

```suggestion
    def published(self, published_as_of=None):
```

docs/models.py
```diff
@@ -212,6 +221,83 @@ def sync_to_db(self, decoded_documents):
...
+        """
+        if self.lang == "en":
+            # The request needs to come through as a valid one, it's best if it
+            # matches the exact host we're looking for.
+            www_hosts = [
+                host for host in settings.ALLOWED_HOSTS if host.startswith("www.")
+            ]
+            if not www_hosts or not (www_host := www_hosts[0]):
+                return
+            synced_views = [
```
bmispelon: I think this should live in a constant at the top level of a module (I'll let you decide which module makes more sense 😁 )

docs/models.py
```diff
@@ -212,6 +221,83 @@ def sync_to_db(self, decoded_documents):
...
+                # Page title, url name, url kwargs
+                ("Django's Ecosystem", "community-ecosystem", {}),
+            ]
+            for title, url_name, kwargs in synced_views:
+                absolute_url = reverse(url_name, kwargs=kwargs, host="www")
+                path = reverse_path(url_name, kwargs=kwargs)
+                request = RequestFactory().get(path, HTTP_HOST=www_host)
+                body = resolve(path).func(request).render().text
```
bmispelon: That's a lot of stuff going on in such a few lines: you make it look deceptively simple 😁 

I'm curious: did you try other approaches before settling on this? I'm a bit worried that the whole `www_host` logic could be a bit brittle. If we're calling the view function directly anyway, can we skip setting the host on the request object?

docs/models.py
```diff
@@ -212,6 +221,83 @@ def sync_to_db(self, decoded_documents):
...
+                            },
+                        ],
+                        "parents": DocumentationCategory.WEBSITE.value,
+                        "slug": entry.slug,
+                        "title": entry.headline,
+                        "toc": "",
+                    },
+                    config=TSEARCH_CONFIG_LANGUAGES.get(
+                        self.lang[:2], DEFAULT_TEXT_SEARCH_CONFIG
+                    ),
```
bmispelon: Can we factor this into a helper function that can be reused also in `sync_views_to_db()`?

```suggestion
                    config=get_search_config(self.lang),
```

docs/models.py
```diff
@@ -212,6 +221,83 @@ def sync_to_db(self, decoded_documents):
...
+                )
+
+    def _sync_views_to_db(self):
+        """
+        Sync the blog entries into search based on the release documents
+        support end date.
```
bmispelon: Bad copy/paste

bmispelon (commented):
This is a fantastic PR, thanks for submitting it! It addresses a real issue that's been plaguing the site for a long time, and as far as PR go, it's a really high quality one, kudos 👏🏻 

Of course because I'm me I'll always find something to complain about so I've left a few comments ranging from improvement suggestions to open questions.

I'll note that I have not tested the code locally (yet). I'm still working on getting the preview server up and running and I think this PR would be a great candidate for it.

blog/models.py
tim-schilling: You don't enjoy multiple self-references in your help text for your self?

docs/tests/test_models.py
tim-schilling: I think I got a bit overzealous. It's needed in another spot, but not for queryset filters. Thanks for catching this.

docs/search.py
tim-schilling: So this is a bigger question. We need the value to be `"weblog"` to make the breadcrumbs work properly in the search area without adding another hack on the search results side to adjust things. I suppose we could just have "www.djangoproject.com/website/" redirect to "www.djangoproject.com/weblog/". However, I have no idea what to call that category of results on the actual page. I think right now I have it as "Django Website" which doesn't seem right.

blog/models.py
tim-schilling: Sure. What do you think of `as_of` instead? `Entry.objects.published(as_of=yesterday)`

docs/models.py
tim-schilling: This is definitely for a reason, but let me try once more and better document the weirdness.

docs/models.py
tim-schilling: I can't tell if this is a suggestion or a question. If it's a question, yes it's possible. I didn't because I honestly don't expect this to be more than 50 records.

If this is more of a suggestion, should I take the optimization to batch fetch the searchable entries as well? That way if we somehow have 2000 records, we don't fetch them all from the DB at once. I've also used `.iterator()`, collected until the batch size, then called `bulk_create()`.

docs/models.py
tim-schilling: If we're moving this ~from the code~ away from the usage logic, then we should use a `dataclass` so it's not relying on positional arguments. That gets a bit more verbose and likely could go in an entirely separate module.

docs/tests/test_models.py
tim-schilling: Judging from the rest of the tests, it seems like there's an effort to keep them efficient. It may make more sense as two separate tests, one for the old `document_url` and then a new one for the `DocumentationCategory.WEBSITE` branch. Let me know if you disagree.

blog/migrations/0008_alter_entry_is_searchable.py
tim-schilling: I added a commit consolidating the migrations.

docs/models.py
tim-schilling: @bmispelon do you have any good ideas here?

docs/models.py
tim-schilling: Okay, so the simplest we can get it is to use `get_template("aggregator/ecosystem.html").render()`. However, that does sort of open us to the problem of the view (which is defined in the urls.py) of using a different template or implementation than expected. However, using this approach eliminates the need to get the host which eliminates having the search the ALLOWED_HOSTS setting. So I think it's worth it. Let me know what you think.

docs/models.py
tim-schilling: May be side-stepped due to https://github.com/django/djangoproject.com/pull/2136#discussion_r2320579514

blog/migrations/0008_alter_entry_is_searchable.py
adamzap: Thanks!

adamzap: @tim-schilling Forgive me for belaboring this, but I don't want to seem dismissive of your original intent when splitting the migrations. According to my research, the non-trivial table-locking behavior you are referring to is not a concern in modern Postgres (>= v11). This has been my personal observation as well.

1. https://stackoverflow.com/a/77987752
2. https://www.enterprisedb.com/blog/adding-new-table-columns-default-values-postgresql-11
3. https://www.postgresql.org/docs/release/11.0/
    - > Many other useful performance improvements, including the ability to avoid a table rewrite for `ALTER TABLE ... ADD COLUMN` with a non-null column default

Let me know what you think. I'd like to know if I'm misunderstanding this!

Thanks!

tim-schilling: Hmm. Weird. I swore I ran into this several times with postgres 13 or 14.

tim-schilling: I wonder if Django's migrations generated SQL that didn't utilize that feature of postgres in the past.

tim-schilling: Thank you for pointing that out! That's something new I learned.

blog/models.py
bmispelon: I'm not usually a big fan of self-references, [see here for why](https://github.com/django/djangoproject.com/pull/2136#discussion_r2322202831) 🙃

blog/models.py
bmispelon: I like it! 👍🏻

tim-schilling: Switched to it.

docs/models.py
```diff
@@ -95,6 +104,11 @@ class DocumentRelease(models.Model):
         on_delete=models.CASCADE,
     )
     is_default = models.BooleanField(default=False)
+    support_end = models.DateField(
+        null=True,
+        blank=True,
+        help_text="The end of support for this release of Django.",
+    )
```
bmispelon: Do we need a dedicated field for this, why couldn't we get the information from `release.eol_date` instead?

docs/models.py
```diff
@@ -202,16 +216,81 @@ def sync_to_db(self, decoded_documents):
...
+        support end date.
+        """
+        if self.lang != "en":
+            return  # The searchable views are only written in English currently
+
+        for searchable_view in SEARCHABLE_VIEWS:
+            absolute_url = reverse(searchable_view.url_name, host="www")
+            # This must match the template used for the url `community-ecosystem`
+            html = get_template("aggregator/ecosystem.html").render()
+            # Need to parse the body element.
```
bmispelon: Can we make `absolute_url` and `html` attributes (likely a `property`) of `SearchableView`? The name of the template should be an attribute of the `SearchableView` too and not be hardcoded here.

docs/search.py
bmispelon: Do the breadcrumbs work for you in the first place? I don't think I've ever used that feature and it doesn't seem to work correctly for me locally: even for documents with two or more parents, the links for all the parents are all the same and go to the document's own page.

Not sure I'd want to start adding redirects, that seems like solving the problem at the wrong level to me.

bmispelon (commented):
I like the move to a `dataclass` for `SearchableView`, nice one 👍🏻 

I think we can go further and delegate the nitty-gritty details of getting a "document"'s URL and content to that class.

I've got grand ideas of generalizing this and having a collection of searchable/indexable models (sphinx document, blog, template view, flat page, ...) but that'd make the scope of this PR explode and I'd rather see this shipped. We can always iterate.

Other than that I left a question about the new `support_end`. I think we should be conservative with adding new fields and would like to explore reusing existing ones, to minimize the combinatorial explosion of possible model states. If we have to keep `support_end`, would it be possible for the existing `DocumentRelease` to get a sensible default assigned to them with a data migration (we have 100 of them in production, it'd be a shame if we had to go and back-populate them by hand).

Thanks again for your work on this!

docs/models.py
tim-schilling: This is why we do code reviews. Let me get this switched over.

docs/search.py
tim-schilling: I have no idea what I ran into originally. Looks like we can rename this to be whatever we want.

The breadcrumbs links on the search just point directly to the search result. Not the actual parents. May be a bug to be honest :face_with_peeking_eye:

docs/models.py
tim-schilling: @bmispelon I think this is still an open conversation.

tim-schilling: Converted over.

pauloxnet:
Can you review this PR based on this latest change in the search docs https://github.com/django/djangoproject.com/pull/2192

tim-schilling:
@pauloxnet I don't see anything obvious that would cause a problem. As long as #2192 didn't change how the search data needs to be stored, then this PR won't be impacted. The hack that is this PR is to insert search-able data and point it to the blog or ecosystem page.

pauloxnet:
> I don't see anything obvious that would cause a problem. As long as #2192 didn't change how the search data needs to be stored, then this PR won't be impacted. The hack that is this PR is to insert search-able data and point it to the blog or ecosystem page. 

I was indirectly suggesting using the approach outlined in the documentation, which involves delegating the generation of search vectors to a generated field. I'm sorry I don't have the time to do a more thorough review of this PR (I'm recuperating from my Django sprints). 😅

tim-schilling:
@pauloxnet I think you can agree there's value in shipping functionality rather than perfecting the approach. If that suggestion is more of a, "hey we need to conform to this new standard", I can make the changes. If it's a "it'd be cool if it did this" type suggestion, I'm going to push for shipping something first.

pauloxnet:
> I think you can agree there's value in shipping functionality rather than perfecting the approach. If that suggestion is more of a, "hey we need to conform to this new standard", I can make the changes. If it's a "it'd be cool if it did this" type suggestion, I'm going to push for shipping something first.

My comment was, rather, I saw that you started this PR before the docs search improvement with GeneratedField, and, not having time to do a full review of the PR, I tried to point it out in this PR. No intent of perfectionism.

docs/models.py
```diff
@@ -202,16 +210,86 @@ def sync_to_db(self, decoded_documents):
...
 
+        self._sync_blog_to_db()
+        self._sync_views_to_db()
+
+    def _sync_blog_to_db(self):
+        """
+        Sync the blog entries into search based on the release documents
+        support end date.
+        """
+        if self.lang != "en" or not self.release.eol_date:
```
sarahboyce:
```suggestion
        if self.lang != "en" or (self.release and not self.release.eol_date):
```
You need something like that as there are english "dev" docs which don't have a release object and eol date
Might be worth testing this case

blog/models.py
```diff
@@ -126,6 +131,12 @@ class Entry(models.Model):
         ),
         default=False,
     )
+    is_searchable = models.BooleanField(
```
sarahboyce, line 132: Do you have examples of blog entries you're expecting to be included and excluded? To me, I feel like everything should be searchable

tim-schilling: No I haven't, but since this is searching the docs, not searching the site, I would imagine blog posts such as announcing board elections would not be searchable.

docs/models.py
tim-schilling: Ah, my assumption was that any release had an EOL date set when it was created.

sarahboyce: That might be correct, but we have "dev" which doesn't have a release object I think

blog/models.py
sarahboyce: I would be tempted to default to True so that once this is launched, everything is included and we exclude if necessary following feedback

My motivation is I'm very tempted to make the searchbar, that's only available on docs, available everywhere and have us slowly add more website pages as searchable/indexed

tim-schilling: Since I don't have admin access and can't manage this myself, I've written the PR to be very conservative. If you're volunteering to handle that, I'm happy to make those changes 😁🚀

bmispelon: > Since I don't have admin access and can't manage this myself [...]

Well, you have access now 😁  (I've put you in the "blog authors" group so you should be able to see all the relevant models)

tim-schilling: :melting_face:

sarahboyce: I suppose the most conservative is not to have the field and it's all searchable. I want to test but I feel like if you searched for someone's name or about Fellows or DjangoCons, you'd want to see blog results but not that many blog results would show up about specific features of Django unless it's a security release description 🤔

tim-schilling: @sarahboyce I think we may have differing opinions on what we're trying to make searchable. To me the following blog posts should not be included in the search results:

- 2026 DSF Board Nominations
- 2025 Malcolm Tredinnick Memorial Prize awarded to Tim Schilling
- Django security releases issued: 5.2.7, 5.1.13, and 4.2.25
  - This exists in the docs already
- Sarah Boyce - Maternity leave announcement
- Django 6.0 alpha 1 released
  - This exists in the docs already
- Nominate a Djangonaut for the 2025 Malcolm Tredinnick Memorial Prize
- Last call for DjangoCon US 2025 tickets!
  - Maybe if we could limit when it gets included? I don't want results about tickets being sold for a conference that's already happened.

Posts that should be searchable:
- Getting Started With Open Source Through Community Events
- Keyboard shortcuts in Django via GSoC 2025
- Watch the DjangoCon Europe 2025 talks

Posts that I'm unsure about:
- DSF at EuroPython 2025: Celebrating 20 years of Django
  - This feels really similar to the "tickets being sold" type post from above
- DSF member of the month - [Name]
  - I can be easily convinced to include these. As of right now, my personal goal is to make the framework material more searchable, not necessarily the community.
- Djangonaut Space is looking for contributors to be mentors
  - It's timeboxed and not always relevant

docs/models.py
tim-schilling: Good catch @sarahboyce. Fixed in the second commit.

tim-schilling:
@sarahboyce identified a problem outside of the PR in that the `update_docs` command is only run periodically which means we wouldn't be introducing any searchable website content until that next run.

sarahboyce:
> the `update_docs` command is only run periodically which means we wouldn't be introducing any searchable website content until that next run.

Want to add that there is a little bit more here. We only update the docs for the search if there are changes to the docs. This means only when the `django/django` repo has a new commit on the branch for that release (e.g. `main` = dev, `stable/5.2.x` = 5.2). This happens quite frequently for `main`, versions in pre-release and the version in active support, but would take a security release that might happen every other month for versions like 4.2

So we _might_ want to have that the blog/website entries can be created without changes being required on the `django/django` repo

tim-schilling:
@sarahboyce and I discussed this PR. We arrived at the following conclusions:

- We should remove the `.published(as_of=)` parameter because it's backwards and going to cause more problems than it's worth down the line. We should let users decide what content is relevant for them and eventually we can apply search rankings to the date of the blog post.
- We're going to stick with the `is_searchable: bool` property, though in the future this may change as the site wide search feature is implemented. (the idea of categories was floated for the `Entry` / blog post model)
- This approach has the downside of a blog post being created and not being immediately searchable. However, given that most documentation is rebuilt within a week, this seems acceptable to release this. A follow-up issue and PR should be created to fix this though.

tim-schilling:
This is ready for a hopefully final review.

Can I provide a script to run on production to batch update blog posts?

bmispelon:
> Can I provide a script to run on production to batch update blog posts?

You mean a data migration? Sure you can add one to the PR 😁 
A custom management command could work too (and is easier to test).
A script would be a last resort option I would say.

I'm hoping to be able to review this tomorrow during Mr Triplett's office hours, so you have time to push a few more commits.

blog/models.py
sarahboyce: I discussed with Tim and I understand that this can be used to distinguish "technical" content which we want to be included in a "docs/technical" search

docs/tests/test_models.py
```diff
@@ -547,3 +584,46 @@ def test_excluded_documents(self):
...
+        document_index, document_detail = self.release.documents.order_by("path")
+        self.assertEqual(
+            document_index.get_absolute_url(),
+            "http://docs.djangoproject.localhost:8000/en/1.2.3/",
+        )
+        self.assertEqual(
+            document_detail.get_absolute_url(),
+            "http://docs.djangoproject.localhost:8000"
+            "/en/1.2.3/topics/http/generic-views/",
+        )
```
bmispelon: This PR is too good and I can't find anything to criticize, so I'll joint point out that you can save 2 lines (and get a better failure exception) by using `assertQuerySetQual`:
```suggestion
        self.assertQuerySetEqual(
            self.release.documents.order_by("path"),
            [
                "http://docs.djangoproject.localhost:8000/en/1.2.3/",
                "http://docs.djangoproject.localhost:8000/en/1.2.3/topics/http/generic-views/",
            ],
            transform=lambda doc: doc.get_absolute_url(),
        )
```

tim-schilling: Ack, I thought I cleaned all these up before.

tim-schilling:
@bmispelon I don't agree with using a data migration to bulk update the entries. I have primary keys and if we put it in a data migration, we run the risk of updating data in non-production environments unexpectedly causing weird issues that are difficult to debug.

Edit: Considering there are only 22, I can do them manually after it's been deployed.

```
805
804
801
796
792
789
788
783
781
774
773
744
739
738
721
707
702
655
650
559
451
442
418
391
```

bmispelon:
Oh, so the script is just:
```
Entry.objects.filter(pk__in=[805, 804, 801, 796, 792, 789, 788, 783, 781, 774, 773, 744, 739, 738, 721, 707, 702, 655, 650, 559, 451, 442, 418, 391]).update(is_searchable=True)
```
?
Yeah, I can run that after deploy. I agree that a migration is definitely not the way to go. I thought maybe there was some more complicated logic based on dates and/or content.

tim-schilling:
Oh, no. It was just "Tim's picks". If others want to add more, that's fine, but figured we could start with those.

bmispelon (approved):
I've tested this locally and it works great!

I'm happy to merge, would you prefer to squash all the commits, or do you think it makes sense to keep some of them separate?

Also, do you want to add a couple dozen "Co-authored: Tim Schilling" to the commit message for good measure? 😁

tim-schilling:
I was assuming you'd use the squash and merge option. I can squash them in a bit though.

`Co-Authored-By Tim`

sarahboyce:
Unsure how much of an issue this would be but be aware that blog items that currently exist in https://www.djangoproject.com/sitemap.xml would now also exist in https://docs.djangoproject.com/sitemap-en.xml

Fixing this should be relatively simple :thinking: 
I think we should exclude from the docs sitemap as this would also duplicate the same urls per version of Django within the sitemap

sarahboyce:
Can cherry-pick 932d8ceb3261337d5860047ace9fe72e47976034 onto this PR :+1:

tim-schilling:
Squashed and cherry-picked that commit @sarahboyce. I had to add the setting as well which wasn't in that commit from what I could find. I'm not 100% sure why I'm cherry-picking it into this PR though.

bmispelon:
Sorry @tim-schilling, we never made it to 💯 comments

tim-schilling:
It's okay :grin: 93 comments is a nice number to end on. Thank you @bmispelon, @sarahboyce and @adamzap for your feedback!

PR #2137, opened 2025-07-26
Remove unnecessary limitation for allowed hosts in development environments

ulgens:
My original goal was to resolve the issue with the Docker setup and `0.0.0.0`.

CleanShot 2025-07-27 at 00 26 08@2x

But after a quick investigation, I couldn't find a case where we need to keep a whitelist of hosts for development purposes. Because the setting wasn't providing any benefit but adding complexity, I used this as as a small refactoring opportunity. Please let me know if there is any case that we need to limit the allowed hosts in non-prod environments. 

If we actually need this limitation, I'll ask for lists of allowed hosts for dev and docker envs. It seems the dev is missing `0.0.0.0`, and docker doesn't have any of the entries the dev has - which makes me think both settings need to be revamped.

### DIFF
--- a/djangoproject/settings/dev.py
+++ b/djangoproject/settings/dev.py
@@ -1,11 +1,6 @@
 from .common import *  # noqa
 
-ALLOWED_HOSTS = [
-    "www.djangoproject.localhost",
-    "djangoproject.localhost",
-    "docs.djangoproject.localhost",
-    "dashboard.djangoproject.localhost",
-] + SECRETS.get("allowed_hosts", [])
+ALLOWED_HOSTS = ["*"]
 
 LOCALE_MIDDLEWARE_EXCLUDED_HOSTS = ["docs.djangoproject.localhost"]
 
--- a/djangoproject/settings/docker.py
+++ b/djangoproject/settings/docker.py
@@ -13,8 +13,6 @@ DATABASES = {
 
 SECRET_KEY = os.environ.get("SECRET_KEY")
 
-ALLOWED_HOSTS = [".localhost", "127.0.0.1", "www.127.0.0.1"]
-
 LOCALE_MIDDLEWARE_EXCLUDED_HOSTS = ["docs.djangoproject.localhost"]
 
 # django-hosts settings


djangoproject/settings/dev.py
```diff
@@ -1,11 +1,6 @@
 from .common import *  # noqa
 
-ALLOWED_HOSTS = [
-    "www.djangoproject.localhost",
-    "djangoproject.localhost",
-    "docs.djangoproject.localhost",
-    "dashboard.djangoproject.localhost",
-] + SECRETS.get("allowed_hosts", [])
+ALLOWED_HOSTS = ["*"]
```
marksweb, line 3: Technically this could just be done in the docker settings because I think it's only related to running with docker due to the docker network.

marksweb (approved):

djangoproject/settings/dev.py
sabderemane: I agree with @marksweb , what is the purpose of changing this part?

sabderemane (changes requested):
I'm unsure to understand the change in dev file, could we clarify this before merging?

djangoproject/settings/dev.py
ulgens: >  I couldn't find a case where we need to keep a whitelist of hosts for development purposes

The purpose is to remove a broken setting that creating extra complexity with no benefit; refactoring.

ulgens:
I updated the description and the title to further clarify this is a refactoring.

djangoproject/settings/dev.py
ulgens: >  I think it's only related to running with docker due to the docker network.

0.0.0.0 is not related to the Docker network.

tobiasmcnulty: Have you run the site with this configuration, accessed all of the different domains listed in the file, and confirmed that they display the appropriate site?

If the proper host name is not getting through to Django, it seems like the problem needs to be fixed somewhere else.

ulgens: > Have you run the site with this configuration, accessed all of the different domains listed in the file, and confirmed that they display the appropriate site?

Yep. This setting doesn't affect how existing host configuration works.

> If the proper host name is not getting through to Django

Not sure what does that mean. `0.0.0.0` is currently not usable because of this setting, not because another config related to hosts.

tobiasmcnulty: I tried the Docker instructions, and I think the issue is the README. "View the site at http://localhost:8000/" should be updated to show the correct domains, which can be found at the end of the non-Docker instructions ("http://www.djangoproject.localhost:8000/, http://docs.djangoproject.localhost:8000/, or http://dashboard.djangoproject.localhost:8000/.").

The reference to 0.0.0.0:8000 in the README is for Foreman (I am not familiar with this), but I suspect it's wrong and should be removed or updated as well.

Working for me with Docker and the correct domain, and no local changes:

Screenshot 2025-07-27 at 3 35 00 PM

Screenshot 2025-07-27 at 3 34 56 PM

ulgens: @tobiasmcnulty Yes, there is a workaround, but once again, the point is to remove a setting that doesn't provide any benefit. I'm missing what is recommended in this thread. Are you saying "we don't need to allow 0.0.0.0"? If that's the case, I see no reason to block it, and the instructions can be updated (and will be updated, because Docker setup is not ready to use yet). If your point is "we don't need refactoring", I don't think not improving a config is a good thing. In any case, I need to ask for actionable insight.

ulgens: About foreman, not sure why you think it's wrong but accessing a local project over 0.0.0.0, 127.0.0.1 or localhost is common practice, which I see no issue in having mentioned in a doc. Having the mention of foreman, and the procfile in the repo, is something I'd like to discuss but that would be a separate refactoring discussion.

tobiasmcnulty: As I said before, my recommendation is to update the README. This is not a workaround; these are the actual domains you need to use to access the development site, as listed in the README, the `ALLOWED_HOSTS` setting, the `dev_sites` fixture, and the django-hosts configuration (`djangoproject/hosts.py`). They're also the domains used by the test suite. It's the same for the non-Docker setup, and the `ALLOWED_HOSTS` settings helps you see that you're not using the development server as it was intended.

If you would like to propose a change to allow the main site to be access on `localhost:8000` instead of or in addition to `www.djangoproject.localhost:8000`, the change deserves its own issue where the @django/django-website team and others who work on the site can discuss and decide how to move forward.

ulgens: Thanks for the clarification. So, we need developers to access the project locally via specific hostnames 👍🏻 Blocking undesired hosts via `ALLOWED_HOSTS` provides only a partial solution for that need and creates further confusion. It blocks access in particular scenarios, but it doesn't help to find the right entry points or provide any information that the developer is doing something undesired. I think a better solution could be to apply the changes in this PR, and then follow up with forwarding requests from common entry points, such as "0.0.0.0" and "127.0.0.1", to the desired hostnames. I'll create an issue to have a healthier discussion 🌻

adamzap:
@ulgens @tobiasmcnulty Is this draft PR still relevant given #2207 and other recent Docker-related changes?

tobiasmcnulty:
> @ulgens @tobiasmcnulty Is this draft PR still relevant given #2207 and other recent Docker-related changes?

No, I think it can be closed. #2198 (merged) and the follow-on work in #2207 (ready for review) get the Docker setup working again for tests and local dev, respectively. Please try it out though. :)

ulgens:
@adamzap only partially. I think I found a misunferstanding about what does this config do but I couldnt find the time to explain it yet. I'd be happier if we keep the PR open for now.

djangoproject/settings/dev.py
marksweb: > >  I think it's only related to running with docker due to the docker network.
> 
> 0.0.0.0 is not related to the Docker network.

Yeah I don't think that's what I was getting at.

The zero IP allows you to bind all interfaces. It's an easy "allow all" type approach.

Looking back at the screenshots, it's not something you'd then use to connect from the browser.

ulgens:
> I'd be happier if we keep the PR open for now.

If we can't finalize the discussion until the next meeting, let's give it a one last shot in the meeting and then I'm okay with closing it if that would be the consensus.

djangoproject/settings/dev.py
ulgens: > This is not a workaround; these are the actual domains you need to use to access the development site, as listed in the README, the ALLOWED_HOSTS setting, the dev_sites fixture, and the django-hosts configuration (djangoproject/hosts.py)

> the dev_sites fixture

This can be updated/improved, but I don't have the energy left to start a discussion about another arbitrary choice.

> django-hosts configuration (djangoproject/hosts.py)

This is not correct. Hosts config doesn't check for the domain:

```
host_patterns = [
    host(r"www", settings.ROOT_URLCONF, name="www"),
    host(r"docs", "djangoproject.urls.docs", name="docs"),
    host(r"dashboard", "dashboard.urls", name="dashboard"),
]
```

So django-hosts doesn't care if you access it via `docs.0.0.0.0`, `docs.localhost` or `docs.djangoproject.localhost`

ulgens:
There was strong opposition to this PR, arguing that there is a technical reason to have this limitation in place, but it appears to be more a matter of preference than a technical requirement. The fixture that has the hardcoded domain name can be updated, but I already spent more than 10x of the effort I initially allocated for this PR, and I'm not interested in keeping it up.

PR #1198, opened 2022-09-03
Fixed #1199 -- Added pre-commit hooks

pauloxnet:
As a consequence of the comments relating to the old pull request, I created a new pull request to add pre-commit.
The configuration file now has only basic hooks with the exclusion of migrations files.

PR #2138, opened 2025-07-27
Don't exclude migrations from black

ulgens:
The exclusion of migration files from black was presented in https://github.com/django/djangoproject.com/pull/1198

> ... The configuration file now has only basic hooks with the exclusion of migrations files.

I suppose this was considered a sane default, and I couldn't find any discussion about that exclusion in the PR. Considering that migration files are code, they are already revisioned and formatting them is [Django's default behaviour](https://docs.djangoproject.com/en/5.2/ref/django-admin/#black-formatting); not excluding them makes feels more natural.

Relates to:
- https://github.com/django/djangoproject.com/issues/1184
- https://github.com/django/djangoproject.com/pull/1198

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -1,4 +1,5 @@
 -r common.txt
+black==25.9.0  # Used by Django's find_formatters and run_formatters
 django-debug-toolbar==6.0.0
 pre-commit~=4.3.0
 watchdog==6.0.0


SaptakS:
I don't feel very strongly against or in support of this. I personally love black, but the only rationale I can think of excluding migrations is that it increases the migration file sizes a lot in attempt to make it more readable, and I am not sure that's necessary. I feel the number of migrations can become a lot, and with bigger file sizes, that don't need to be very readable, this may take more time in things like git cloning, etc.? Though maybe realistically not so much, hence I don't have a strong personal opinion against this PR.

marksweb:
I wouldn't format migration files. They rarely need reading and when they do, the long lines tend to be beneficial because you can [more] easily see all the field names in a model without the distraction of the attributes, for example.

ulgens:
@SaptakS I checked the before-and-after total file sizes. The difference is 12 KB. For comparison, `foundation/migrations/0005_alter_approvedgrant_amount_currency_and_more.py` file is 20 KB. The difference can easily be compensated for by squashing a single migration. Also, git is extremely scalable, and it can handle repositories with multiple hundreds of GBs in size (We are at 33 MB)

As an additional data point, this is [GrandPerspective](https://grandperspectiv.sourceforge.net/) output for the project folder:

CleanShot 2025-07-28 at 12 06 25@2x

A single file in the .git folder is ~2000x bigger than the size difference created by the formatting.

@marksweb Thanks for the input. I don't personally disagree with your points, but I also don't strongly agree with them. Both points you mentioned are subjective, and the second one depends on personal taste, your IDE config, etc. As a tool, black's philosophy is 

> Black is the uncompromising Python code formatter. By using it, you agree to cede control over minutiae of hand-formatting. In return, black gives you speed, determinism, and freedom from pycodestyle nagging about formatting. You will save time and mental energy for more important matters.

That approach aims to remove the need for this type of "how do we format things" questions and apply a consistent styling without requiring one to think about which rules to use or where to apply them. And again, Django's default is to format the generated files. For the sake of compliance with both tools' approaches and to avoid falling into the trap of bikeshedding, I think it's better to proceed with a minimal amount of customization.

SaptakS:
> Django's default is to format the generated files.

This point does make me more inclined to accepting this PR.

requirements/dev.txt
```diff
@@ -1,4 +1,5 @@
 -r common.txt
+black==25.1.0  # Used by Used by Django's find_formatters and run_formatters
```
CuriousLearner:
```suggestion
black==25.1.0  # Used by Django's find_formatters and run_formatters
```

SaptakS:
@ulgens can you resolve the merge conflict? And let's have a consensus vote with discussion in our next monthly call to conclusively make a decision. Thoughts?

Stormheg:
I too am in favor of formatting migration files 👍

requirements/dev.txt
ulgens: Resolved.

ulgens:
> can you resolve the merge conflict?

@SaptakS Done 🌻

pauloxnet (approved):
LGTM

ulgens:
Updated black version to 25.9.0

ulgens:
@django/mergers We have 3 in favor, 1 againist and 1 neutral. Do we need to discuss further, what is our next step?

jacobtylerwalls:
I see a comment above about discussing on the monthly call, did that happen? (The mergers team only applies to django/django, everything else is self-managed by working groups.)

SaptakS:
@ulgens the group to be tagged here is: @django/django-website-mergers . And I think I am going to merge this. Though we didn't get to discussing it on the monthly call, I don't think there is a strong opinion or logic against this.

ulgens:
> The mergers team only applies to django/django, everything else is self-managed by working groups.

> the group to be tagged here is: @django/django-website-mergers

Sorry, my bad.

jacobtylerwalls:
No worries! Thanks for the PR!

PR #2144, opened 2025-08-01
Added flatpages-extra

bmispelon:
This PR adds `django-flatpages-extra`, which is a package I created after a discussion with @StephanieAG from the fundraising WG on how to make it easier to update flat pages.

As you can tell from the very short PR, the package is a drop-in replacement and is reversible. It adds the following features to the admin:

- The ability to create a revision for a flat page
- preview revisions (including a shareable link for people who don't have an admin account)
- reverting revisions
- better history of changes for flat pages

Initially I was going to develop this as an app in this project, but then realized it might be useful as a 3rd party too. If accepted I would release a 1.0 of the package.

## Screenshots:

<details><summary>Editing content directly is disabled</summary>
Screenshot 2025-08-02 at 12-02-50 _accessibility_ -- Django Accessibility Statement! Change flat page Django site admin

</details> 

<details><summary>Preview page</summary>
Screenshot 2025-08-02 at 12-07-34 Django Accessibility Statement! Django

</details> 

<details><summary>Compare view (diff)</summary>
Screenshot 2025-08-02 at 12-08-18 Comparing revision with original Django site admin

</details>

### DIFF
--- a/djangoproject/settings/common.py
+++ b/djangoproject/settings/common.py
@@ -75,6 +75,7 @@ INSTALLED_APPS = [
     "django_hosts",
     "sorl.thumbnail",
     "djmoney",
+    "flatpages_extra",
     "django.contrib.sites",
     "django.contrib.auth",
     "django.contrib.admin",
--- a/djangoproject/urls/www.py
+++ b/djangoproject/urls/www.py
@@ -139,6 +139,7 @@ urlpatterns = [
     path("weblog/", include("blog.urls")),
     path("download/", include("releases.urls")),
     path("svntogit/", include("svntogit.urls")),
+    path("flatpagesextra/", include("flatpages_extra.urls")),
     path("", include("legacy.urls")),
 ]
 
--- a/requirements/common.txt
+++ b/requirements/common.txt
@@ -1,6 +1,7 @@
 Babel==2.17.0
 django-contact-form==5.2.0
 django-countries==7.6.1
+django-flatpages-extra==1.0
 django-hosts==5.1
 django-money==3.5.4
 django-push @ git+https://github.com/brutasse/django-push.git@22fda99641cfbd2f3075a723d92652a8e38220a5


bmispelon:
The test failure seems to be caused by a mismatch in the type of the `id` field (`Autofield` vs. `BigAutoField`), I'll work on a fix shortly.

bmispelon:
(Just noticed there's an issue with using HTML inside `contrib.messages`, working on a fix)

ulgens:
Adding the link for easier review: https://pypi.org/project/django-flatpages-extra/

ulgens:
+1 for having the "is published" feature and revisions, but not sure about offloading both of them to a new package.
* If we are going to keep history, I'd prefer a package like https://django-simple-history.readthedocs.io/en/stable/ that works with every model. 
* For "is published" feature, I think a small custom model could solve it.

tim-schilling:
@bmispelon I think we may want to have CI on that package as well, otherwise upgrades could be a bit tricky in the future [if things fall out of date with new versions of Django].

bmispelon:
> @bmispelon I think we may want to have CI on that package as well, otherwise upgrades could be a bit tricky in the future [if things fall out of date with new versions of Django].

The code is hosted on Codeberg where CI is not so easy to set up. I've added a `just ci` command that runs the test suite (100% coverage) on all supported versions of Django: 5.0, 5.1, 5.2 and event 6.0 (alpha release). They all pass locally, so I've released the current code as version 1.0 on pypi.

I also took the liberty of rebasing this PR onto the latest `main` branch.

Would folks be ok if I deployed this on the preview server?

bmispelon:
> +1 for having the "is published" feature and revisions, but not sure about offloading both of them to a new package.
> 
>     * If we are going to keep history, I'd prefer a package like https://django-simple-history.readthedocs.io/en/stable/ that works with every model.
> 
>     * For "is published" feature, I think a small custom model could solve it.

The "small custom model" is how django-flatpages-extra started. I originally created it inside the source code of this site, but realized it could be useful as a third party.

I've designed the `django-flatpages-extra` so it can be added and removed with no ill effect, so we can install it temporarily while an alternate solution is worked on (maybe one involving django-simple-history and/or a simpler custom model).

ulgens:
> The "small custom model" is how django-flatpages-extra started. I originally created it inside the source code of this site, but realized it could be useful as a third party.

Thanks for the details.

> I've designed the django-flatpages-extra so it can be added and removed with no ill effect, so we can install it temporarily while an alternate solution is worked on (maybe one involving django-simple-history and/or a simpler custom model).

That sounds okay to me.

bmispelon:
I've deployed this to the preview server. I'll reach out to Stephanie to see if she has time to test it to see if it this solution would work for her.

Issue #2139, opened 2025-07-28
https://dashboard.djangoproject.com/foundation/corporate-membership/ > RETURNS 404


MarioVlado:
Link origin:            https://dashboard.djangoproject.com/
Link placement:     footer right, under 'Support us'
Link title:                Corporate membership
Link:                       https://dashboard.djangoproject.com/foundation/corporate-membership/

Best,
M

ulgens:
Thanks for reporting this. Looks like a valid bug. The correct link should be https://www.djangoproject.com/foundation/corporate-membership/

Not sure about the root cause but it seems we have an issue with the rendering of the link URL. The issue applies to "Benevity Workplace Giving Program" link too.

kleeena:
@ulgens Hi! I’d like to work on this. Let me know if it’s okay to proceed.

ulgens:
@kleeena Please feel free to proceed with it. If this will be your first code contribution, don't forget to check the readme in this repo, and https://docs.djangoproject.com/en/dev/internals/contributing/ before you start 🌻

kleeena:
@ulgens I looked into this and noticed a few things: the /foundation/corporate-membership and Benevity Workplace Giving Program paths don’t seem to be defined in urls.py, which leads to a "Page Not Found" error when testing locally. I also couldn’t find any templates or views related to these routes in the repo.

Would it make sense to handle these with a redirect or via a context processor, or is there another part of the setup that’s serving these pages in production?

Just want to make sure I'm not missing something, I'd appreciate any direction since it's indeed my first code contribution.

ulgens:
@django/django-website Do these links point to flat pages? Do we have an easy debug route for them?

harriebird:
Hi everyone, check my PR #2146 for this issue.

harriebird:
Hi everyone, let me know if this issue will be open once again for contributors to work on.

ulgens:
@kleeena Hey, it looks like @harriebird has a solution for the issue. I believe https://github.com/django/djangoproject.com/pull/2146 could also answer your latest question. If you would like to continue and push a fix, please feel free to do so; but if not, please let us know, and we can try to help you find a new issue.

kleeena:
@ulgens Hi! Yes definitely, @harriebird can push the fix and I'd love to help with a new issue.

JohananOppongAmoateng:
@ulgens i think this issue should be closed. Since a fix was merged

ulgens:
[...]

PR #2146, opened 2025-08-04
Fix error 404 on foundations urls when coming from other subdomain

harriebird:
The website displays a 404 page when clicking the `Corporate membership` and `Benevity Workplace Giving Program` in the footer if coming from the `docs` or `dashboard` subdomains.

To fix this issue, the host `www` was specified by utilizing the `host_url` template tag from the django-hosts.

This fixes Issue #2139.


SaptakS (approved):
Makes sense to me.

Issue #1922, opened 2025-02-01
Create a stage environment


pauloxnet:
At the moment, the only way to test the new features added to the website is to deploy them directly in production.

I propose to create a stage environment where to test the updates and new versions of the portal to avoid creating disruptions on the production site.

sarahboyce:
I _think_ we had one set up at some point but closed it down to reduce costs. I think we can set one up, but should have a way to basically spin it up on demand and have it close down again to reduce costs.
There might be a time where we are working on the website consistently enough that it's worth the extra cost but I'm not sure that's the case today 🤔

knyghty:
Branch deploys that can be spun up for bigger changes (by a small number of people) and spun back down after merging would be ideal. Perhaps a bit of work though.

adamzap:
I think this is a great idea if we can integrate it into various processes (deployment, testing, maintenance, etc).

I wonder how hard it would be to use a PaaS for this. Ultimately it's up to the ops team, but [fly.io seems pretty friendly to Django](https://fly.io/django-beats/). I wonder if they would provide hosting for a small staging server (or more?) for free. Heroku could be another option.

If there is no interest in using a PaaS, maybe we could decide on the requirements of the server we'd need. That could make it easier for someone to donate the resources if cost is a problem.

adamzap:
We have a staging environment now, so I'm going to close this.

PR #2153, opened 2025-08-07
Changed GH action to publish images from the preview branch too

bmispelon:
I believe this is all that's needed in order to start publishing a `preview` tag to the docker registry used for deployment.

Here's some documentation about the strange `{{is_default_branch}}` condition: https://github.com/docker/metadata-action?tab=readme-ov-file#latest-tag.

I removed the `flavor: latest=true` to avoid tagging a preview image with `latest`.

Refs #1922

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -2,7 +2,7 @@ name: Create and publish a Docker image
 
 on:
     push:
-        branches: ["main"]
+        branches: ["main", "preview"]
 
 env:
     REGISTRY: ghcr.io
@@ -31,9 +31,9 @@ jobs:
               uses: docker/metadata-action@v5
               with:
                   images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
-                  flavor: |
-                      latest=true
                   tags: |
+                      # set latest tag for main branch only
+                      type=raw,value=latest,enable={{is_default_branch}}
                       type=ref,event=branch
                       type=sha
 
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -33,7 +33,7 @@ jobs:
                   images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
                   tags: |
                       # set latest tag for main branch only
-                      type=raw,value=latest,enable={{is_default_branch}}
+                      type=raw,value=latest,enable=main
                       type=ref,event=branch
                       type=sha
 


ulgens:
@bmispelon This is not a direct review for this PR, but I have a question:

How do we feel about making this configurable? We can add a workflow_dispatch trigger, and the person doing the deployment can select which branch they need to be dockerized. This could only need a couple of lines of change in the action (I need to check how it affects the deployment script but I don't expect a huge change), and I'm willing to help with that. Having a "branch" input in the workflow will remove the need to have a specific branch for preview/test deployments.

.github/workflows/docker-publish.yml
```diff
@@ -31,9 +31,9 @@ jobs:
               uses: docker/metadata-action@v5
               with:
                   images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
-                  flavor: |
-                      latest=true
                   tags: |
+                      # set latest tag for main branch only
+                      type=raw,value=latest,enable={{is_default_branch}}
```
ulgens: Do we need to make this dynamic? I believe having a hardcoded "main" is safe enough.

bmispelon: I got the `{{is_default_branch}}` from the action's documentation: https://github.com/docker/metadata-action?tab=readme-ov-file#latest-tag. I do agree that having a hardcoded `main` is unlikely to cause issues so I'm happy to change that.

bmispelon:
> How do we feel about making this configurable? We can add a workflow_dispatch trigger, [...]

That would work too I guess. 

* How would the configuration look like? A drop-down where I can pick a branch, or would I have to copy/paste the branch name manually?
* Would this action publish a `preview` tag any time it's run, or would it tag with a commit id instead (which I'd then need to copy in order to tell the deploy script which tag to use)?
* Having a dedicated `preview` branch has the advantage that it's easier to know what is actually deployed on the preview server
* Similarly, a `preview` branch could serve as an integration branch for separate PRs, rather than having to create ad-hoc integration branch when we want to test multiple things at once.

My strategy with this proposal was to do the simplest thing that works. A `preview` branch is really close to the existing infrastructure and so will require minimal changes (both to this project's github actions and to the deploy playbooks). We can always iterate on it in the future (if we're to have an auto-deploy of some kind for example, then a `workflow_trigger` style of deployment could be great).

bmispelon:
I've also created to do the same for Trac: https://github.com/django/code.djangoproject.com/pull/270

tobiasmcnulty (approved):
Looks good to me! 🚢

ulgens:
> How would the configuration look like? A drop-down where I can pick a branch, or would I have to copy/paste the branch name manually?

CleanShot 2025-08-09 at 15 17 45@2x

This screenshot is from a project that we use a `workflow_dispatch` trigger for the preview/test deployments. Branch dropdown is provided by Actions and you can add other dropdown or text input fields.

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onworkflow_dispatchinputs

> Would this action publish a preview tag any time it's run, or would it tag with a commit id instead (which I'd then need to copy in order to tell the deploy script which tag to use)?

It depends. We can tag it as "preview" and instruct the deployment script to use the latest preview image, or tag it with the commit hash and copy the commit hash to the deployment script. However, in my ideal scenario, this trigger would also handle the triggering of deployment, eliminating the need to call the deployment script separately.

> Having a dedicated preview branch has the advantage that it's easier to know what is actually deployed on the preview server

Logs of the workflow runs could provide this information if the deployment was integrated. I see the current implementation and the need, and this is the easier solution, but there is also room for improvement. Another common thing I do in the projects I work is creating an admin-only "version" endpoint that returns the current git ref (commit hash + branch/tag name).

> Similarly, a preview branch could serve as an integration branch for separate PRs, rather than having to create ad-hoc integration branch when we want to test multiple things at once.

I don't want to trigger another "this was the case in past but we can/should do better" discussion, so I'll keep it short: It makes sense, but it feels like a workaround. I guess I have to agree with it as the short-term solution, but there is room for improvement.

> My strategy with this proposal was to do the simplest thing that works ... We can always iterate on it in the future

💯

PR #2155, opened 2025-08-09
Isolated `dashboard` app JavaScript

adamzap:
This patch moves the `dashboard` app's JavaScript inclusion from the project's `require.js` structure to the app itself.

This was achieved with a new block in the base template called `javascript`. The rest of the site gets the default JavaScript includes, and the `dashboard` app overrides the block to include the JavaScript specific to it. The `dashboard` app benefits by not knowing about JavaScript related to the rest of the site and vice versa.

None of the actual `dashboard` JavaScript code was changed. It was only adapted to the new structure.

- Added `javascript` block to base template
- Added `dashboard_javascript` block to `dashboard` base template
- Included `dashboard` JavaScript directly in `dashboard` templates
- Removed `dashboard`-related entries in `main.js`
- Remove references to `jquery.flot.min.js` from base template
  - This library is only used by the `dashboard` app
- Removed `dashboard-*` CSS classes only used by `require.js`
- Removed `require.js` structure in `dashboard` JavaScript files
- Simplified structure of `utils.js`

### DIFF
--- a/dashboard/templates/base_dashboard.html
+++ b/dashboard/templates/base_dashboard.html
@@ -14,6 +14,4 @@
   <script src="{% static "js/lib/jquery.min.js" %}"></script>
   <script src="{% static "js/lib/jquery.flot.min.js" %}"></script>
   <script src="{% static "js/dashboard/utils.js" %}"></script>
-
-  {% block dashboard_javascript %}{% endblock %}
 {% endblock %}
--- a/dashboard/templates/dashboard/detail.html
+++ b/dashboard/templates/dashboard/detail.html
@@ -13,6 +13,8 @@
   </div>
 {% endblock %}
 
-{% block dashboard_javascript %}
+{% block javascript %}
+  {{ block.super }}
+
   <script src="{% static "js/dashboard/detail.js" %}"></script>
 {% endblock %}
--- a/dashboard/templates/dashboard/index.html
+++ b/dashboard/templates/dashboard/index.html
@@ -25,6 +25,8 @@
   </div>
 {% endblock %}
 
-{% block dashboard_javascript %}
+{% block javascript %}
+  {{ block.super }}
+
   <script src="{% static "js/dashboard/index.js" %}"></script>
 {% endblock %}


djangoproject/templates/base.html
```diff
@@ -117,32 +117,33 @@
...
-          "stripe-checkout": "https://checkout.stripe.com/checkout",
-          "stripe": "https://js.stripe.com/v3/?" // ? needed due to require.js
-        }
-      };
-    </script>
-    <script data-main="{% static "js/main.js" %}" src="{% static "js/lib/require.js" %}"></script>
     {% block body_extra %}{% endblock body_extra %}
-    <script src="{% static "js/djangoproject.js" %}"></script>
+
+    {% block javascript %}
```
adamzap, line 122: The addition of this block is probably the most controversial change in this PR.

dashboard/templates/dashboard/detail.html
```diff
@@ -1,14 +1,19 @@
 {% extends "base_dashboard.html" %}
 {% load i18n %}
+{% load static %}
```
marksweb: Why not load it alongside `i18n`?

dashboard/templates/base_dashboard.html
```diff
@@ -8,3 +9,11 @@
 {% block header %}
   <h1>Development <em>dashboard</em></h1>
 {% endblock %}
+
+{% block javascript %}
+  <script src="{% static "js/lib/jquery.min.js" %}"></script>
+  <script src="{% static "js/lib/jquery.flot.min.js" %}"></script>
+  <script src="{% static "js/dashboard/utils.js" %}"></script>
+
+  {% block dashboard_javascript %}{% endblock %}
```
marksweb: It might be simpler to not have `dashboard_javascript` (and arguably it's already dashboard JS as we're in dashboard templates). Then when more JS is needed, use the super...

```
{% block javascript %}
  {{ block.super }}
  <script src="{% static "js/dashboard/index.js" %}"></script>
{% endblock %}

dashboard/templates/dashboard/detail.html
adamzap: Sure, I can do that. Just a personal style thing.

Updated in 9e53eab3.

dashboard/templates/base_dashboard.html
adamzap: Good thinking!

Updated in 7ac24bcc.

PR #2176, opened 2025-08-22
Update download.html: Remove Python2 from top

guettli:
Those who need Python2 will find the info. No need to show this to all users at the top (my point of view).

### DIFF
--- a/djangoproject/templates/releases/download.html
+++ b/djangoproject/templates/releases/download.html
@@ -20,7 +20,7 @@
   <h1>How to get Django</h1>
   <p>Django is available open-source under the
     <a href="https://github.com/django/django/blob/main/LICENSE">BSD license</a>.
-    We recommend using the latest version of Python 3.
+    We recommend using the latest version of Python.
     See <a href="{% url 'document-detail' lang='en' version='dev' url='faq/install' host 'docs' %}#what-python-version-can-i-use-with-django">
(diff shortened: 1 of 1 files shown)


carltongibson (commented):
Yes, +1, it's more than time.

djangoproject/templates/releases/download.html
```diff
@@ -20,9 +20,8 @@
   <h1>How to get Django</h1>
   <p>Django is available open-source under the
     <a href="https://github.com/django/django/blob/main/LICENSE">BSD license</a>.
-    We recommend using the latest version of Python 3. The last version to
-    support Python 2.7 is Django 1.11 LTS. See <a href="
-                                                        {% url 'document-detail' lang='en' version='dev' url='faq/install' host 'docs' %}#what-python-version-can-i-use-with-django">
+    We recommend using the latest version of Python 3.
```
ulgens: Do we still need to say "Python 3"?  I think "Latest version of Python" is clear that it doesn't mean Python 2.

bmispelon: I agree with Ülgen, we don't need to mention "Python **3**" explicitly anymore.

```suggestion
    We recommend using the latest version of Python.
```

bmispelon (approved):
I'd say this is good to go 👍🏻 

Thanks for working on this! 🎸

Issue #1766, opened 2024-11-23
The fundraising text for 1000$ donors is inconsistent


alexgmin:
In the [fundraising page ](https://www.djangoproject.com/fundraising/) it says "Leadership-level donors contribute $1,000 or more in a calendar year. "
https://github.com/django/djangoproject.com/blob/5537f965ece29b766efe17c1d2137c3d2037811c/djangoproject/templates/fundraising/includes/display_django_heroes.html#L66-L69

However in the page individual donators see it says this "If you've donated at least US $1000, you can submit your logo and we will display it, too."
https://github.com/django/djangoproject.com/blob/5537f965ece29b766efe17c1d2137c3d2037811c/fundraising/forms.py#L46-L53

One implies you have to donate 1000$ in a year, the other 1000$ in total over any period of time.

The correct one is the first as far as I've been told, but someone from the board should confirm.

arpan8925:
Please confirm, Will
Work in this issue too, after confirmation

bmispelon:
Marking this as "on hold" pending confirmation from the DSF board as to which version is correct.

Mamatha1718:
"If you've donated at least US $%d in a calendar year, you can submit your logo and " 
         "we will display it, too."
These changes need.
please assign to me , I will work on this

thibaudcolas:
thank you, I’ll follow up internally to make sure I understand the intention there and get back to this thread

thibaudcolas:
I can confirm the intention is to showcase any donor whose total donations are $1000 and above, regardless of how many payments that is. Would recommend updating the text accordingly, will leave the wording at your discretion.

alexgmin:
@thibaudcolas Just to confirm, it's not about the number of payments, it's whether the amount is in a calendar year or not.

thibaudcolas:
@alexgmin ah, I’m not sure what the cut-off is. I’d recommend someone looks at the code when they want to work on this.

sarahboyce:
Fixed by github.com/django/djangoproject.com/pull/2191

PR #2191, opened 2025-09-11
Update Leaders donation limit text

house-ben:
Fixed the contribution description for #1766 

Since the heroes are based on the total donated_amount contributed, not the donations over the past year, the text has to be updated. 
https://github.com/django/djangoproject.com/blob/100d673a8ff0a9072ee0eeebcee4252f396049bf/fundraising/templatetags/fundraising_extras.py#L95-L113

The fundraisers also have to have donated in the past year to show up
https://github.com/django/djangoproject.com/blob/5537f965ece29b766efe17c1d2137c3d2037811c/fundraising/models.py#L25-L35

### DIFF
--- a/djangoproject/templates/fundraising/includes/display_django_heroes.html
+++ b/djangoproject/templates/fundraising/includes/display_django_heroes.html
@@ -65,7 +65,7 @@
   <p>
     {% blocktranslate trimmed with amount=display_logo_amount|intcomma %}
       Leadership-level donors contribute ${{ amount }} or
-      more in a calendar year.
+      more in total donations.
     {% endblocktranslate %}
   </p>
   <div>
--- a/djangoproject/templates/fundraising/includes/display_django_heroes.html
+++ b/djangoproject/templates/fundraising/includes/display_django_heroes.html
@@ -64,8 +64,8 @@
     Leaders (${{ amount }}+){% endblocktranslate %}</h3>
   <p>
     {% blocktranslate trimmed with amount=display_logo_amount|intcomma %}
-      Leadership-level donors contribute ${{ amount }} or
-      more in total donations.
+      Leadership-level donors contribute at least ${{ amount }} 
+      and have donated in the last {{ display_donor_days }} days.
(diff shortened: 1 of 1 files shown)


house-ben:
I tried squashing my commits. I'm not familiar enough with git to understand how to get my two commits down to one. Anyone that can help?

djangoproject/templates/fundraising/includes/display_django_heroes.html
```diff
@@ -64,8 +64,8 @@ <h3>{% blocktranslate trimmed with amount=display_logo_amount|intcomma %}
     Leaders (${{ amount }}+){% endblocktranslate %}</h3>
   <p>
     {% blocktranslate trimmed with amount=display_logo_amount|intcomma %}
-      Leadership-level donors contribute ${{ amount }} or
-      more in a calendar year.
+      Leadership-level donors contribute at least ${{ amount }}
+      and have donated in the last {{ display_donor_days }} days.
```
CuriousLearner, line 68: I am not sure if we want to have this displaying `in the last 365 days`. I think I'll prefer to say `in the last calendar year` instead.

```suggestion
      and have donated in the last calendar year.
```

CuriousLearner: @thibaudcolas What do you think about this? Is the intention here to say 365 days explicitly?

CuriousLearner: I rechecked this, and the way we calculate this is indeed for the last 365 days. So, this change looks good to me.

CuriousLearner (approved):
LGTM 🌮 

Great work @house-ben 🎉 Congrats on your first contribution!

SaptakS (approved):
@house-ben thanks for the PR! The code change makes sense to me. I will let the stakeholders give the final approval for the language change and to check if this PR fulfills the requirements.

cc @sabderemane @thibaudcolas

sabderemane (approved):
LGTM, thank you!

PR #2182, opened 2025-09-04
Removed Coveralls references and configuration

adamzap:
After some discussion, the Website WG decided that we no longer want to use Coveralls. A coverage report is generated when running `make ci` locally.

Closes #2172.

PR #2192, opened 2025-09-11
Fix #1650 Replace search field with generated vector field

### LINKED ISSUE #1650: Start using GeneratedField for Documents

pauloxnet:
## Overview

This pull request addresses [issue #1650](https://github.com/django/djangoproject.com/issues/1650), which proposed migrating the search vector field for Document models to a **Generated Field**. The goal is to leverage the database's capabilities to automatically update search-related data, eliminating the need for manual index updates.

## Key Changes

- **Replaces the manually updated `SearchVectorField`** with a **database-generated vector field** (using Django's `GeneratedField`).
- The search vector is now always in sync with the Document's data, as the database computes the value automatically whenever the document changes.
- Removes the need for custom management commands or signals to update the search index after document changes.
- Refactors any logic that previously depended on manual updates to now use the generated field directly.

## Why This Matters

Previously, the search index was updated via a management command after document changes. This process introduced a **delay** between the update of a document and the update of the search index. During this window, users searching for recently updated documentation often encountered **blank results**, leading to confusion and frequent new issues being opened by contributors and users who could not find expected documentation.

With the **Generated Field**:

- **Index updates are instant:** The search vector is recalculated as soon as the Document changes in the database.
- **No more blank results:** Users will get up-to-date search results without waiting for a scheduled command or manual action.
- **Reduces maintenance overhead:** Volunteers no longer need to monitor or troubleshoot delayed index updates.
- **Simplifies codebase:** Removes the complexity associated with keeping the search index in sync.

## Review Notes

- Please verify that all code paths previously relying on manual index update now use the generated field.
- Check that tests for document creation and update reflect the new instant searchability.
- Confirm that documentation has been updated to reflect the new workflow for search indexing.

## Impact

This change will significantly improve the reliability and responsiveness of documentation search on djangoproject.com. It removes a longstanding source of user confusion and contributor frustration, and leverages modern Django/database features for maintainability and performance.

---

**Thank you for reviewing! Please raise any questions or concerns about edge cases, performance, or deployment.**


docs/models.py
```diff
@@ -344,8 +332,44 @@ class Document(models.Model):
     path = models.CharField(max_length=500)
     title = models.CharField(max_length=500)
     metadata = models.JSONField(default=dict)
-    search = SearchVectorField(null=True, editable=False)
-    config = models.SlugField(default=DEFAULT_TEXT_SEARCH_CONFIG)
+    vector = models.GeneratedField(
+        expression=Case(
+            When(config="arabic", then=get_document_search_vector("arabic")),
```
tobiasmcnulty: Can we use `TSEARCH_CONFIG_LANGUAGES` to avoid hard-coding the language list in another place?

Untested, but perhaps something like this would work:
```suggestion
            *[
                When(config=lang, then=get_document_search_vector(lang)),
                for lang in TSEARCH_CONFIG_LANGUAGES.values()
             ]
```

docs/models.py
```diff
@@ -344,8 +332,44 @@ class Document(models.Model):
     path = models.CharField(max_length=500)
     title = models.CharField(max_length=500)
     metadata = models.JSONField(default=dict)
-    search = SearchVectorField(null=True, editable=False)
-    config = models.SlugField(default=DEFAULT_TEXT_SEARCH_CONFIG)
+    vector = models.GeneratedField(
```
tobiasmcnulty: If we are renaming the field maybe rename it to something more easy to search for in the code, `search_vector` for example?

If it would be safer to leave the field with the original name, now that the migration exists, you can manually rename it back to `search` here and in the migration.

docs/models.py
pauloxnet: Thanks for the suggestion, the code is more readable now. :)

docs/models.py
pauloxnet: Yes, `search_vector` is for sure more explicit, I renamed it.

docs/models.py
```diff
@@ -344,8 +332,20 @@ class Document(models.Model):
...
+                for lang in TSEARCH_CONFIG_LANGUAGES.values()
+            ],
+            default=get_document_search_vector(),
+        ),
+        output_field=SearchVectorField(),
+        db_persist=True,
+    )
+    config = models.SlugField(
+        db_default=DEFAULT_TEXT_SEARCH_CONFIG, default=DEFAULT_TEXT_SEARCH_CONFIG
+    )
```
adamchainz, line 350: You could also add a `CheckConstraint` that ensures the value is an allowed name.

docs/models.py
```diff
@@ -344,8 +332,20 @@ class Document(models.Model):
...
-    search = SearchVectorField(null=True, editable=False)
-    config = models.SlugField(default=DEFAULT_TEXT_SEARCH_CONFIG)
+    search_vector = models.GeneratedField(
+        expression=Case(
+            *[
+                When(config=lang, then=get_document_search_vector(lang))
+                for lang in TSEARCH_CONFIG_LANGUAGES.values()
+            ],
+            default=get_document_search_vector(),
+        ),
```
adamchainz, line 344: I don't understand why you need the giant `Case` / `When`. I would think the original version using `F` works:

```python
expression=(
    SearchVector("title", weight="A", config=F("config"))
    + SearchVector(KeyTextTransform("slug", "metadata"), weight="A", config=F("config"))
    + SearchVector(KeyTextTransform("toc", "metadata"), weight="B", config=F("config"))
    + SearchVector(KeyTextTransform("body", "metadata"), weight="C", config=F("config"))
    + SearchVector(KeyTextTransform("parents", "metadata"), weight="D", config=F("config"))
)
```

It seems `SearchVector` supports `config` being a resolvable, so `F` should work…

https://github.com/django/django/blob/94680437a45a71c70ca8bd2e68b72aa1e2eff337/django/contrib/postgres/search.py#L52-L53

If it's broken, leave a comment explaining the issue, and maybe there's something we can fix in Django?

docs/tests/test_models.py
```diff
@@ -350,9 +350,6 @@ def setUpTestData(cls):
         ]
         Document.objects.bulk_create(Document(**doc) for doc in documents)
 
-    def setUp(self):
-        Document.objects.search_update()
```
adamchainz, line 354: I love to see a `setUp` method going! Faster tests!

docs/models.py
pauloxnet: You can have a config that refers to another text field with the language, but it won't work in a `GeneratedField` because the expression is not `IMMUTABLE`.

I explained this case in my talk about Generated Field at DjangoCon US 2025 last Monday.
https://speakerdeck.com/pauloxnet/djangos-generatedfield-by-example-djangocon-us-2025?slide=66

This is actually why I waited so long to open this PR. Infact I found the workaround using Case/When only recently working on my talk.

My proposed implementation works correctly and the expression is IMMUTABLE as PostgreSQL requests.

PS: there's nothing we can fix in Django ORM that permits us to use SearchVector with a dynamic config from another field since this is a restriction of PostgreSQL.

docs/models.py
pauloxnet: Thanks for the suggestions. I added the constraint.

docs/models.py
adamchainz: Interesting, thanks for the slides. I have yet to watch your talk on the livestream replay, will get round to it shortly!

Can you drop the link in a comment above the `expression` like:

```py
# Use Case/When to force the expression to be immutable, per:
# https://www.paulox.net/2025/09/08/djangocon-us-2025/
```

?

I think this would be useful for future readers!

adamchainz:
Looks like you got unlucky with a Coveralls failure in the test run there:

```
coveralls.exception.CoverallsException: Could not submit coverage: 504 Server Error: Gateway Timeout for url: https://coveralls.io/api/v1/jobs
```

docs/models.py
pauloxnet: > Interesting, thanks for the slides. I have yet to watch your talk on the livestream replay, will get round to it shortly!
> 
> Can you drop the link in a comment above the `expression` like:
> 
> ```python
> # Use Case/When to force the expression to be immutable, per:
> # https://www.paulox.net/2025/09/08/djangocon-us-2025/
> ```
> 
> ?
> 
> I think this would be useful for future readers!

@adamchainz thanks for the suggestions. I've added a comment. Do you mind approving the PR if now it's all clear for you?

pauloxnet:
> Looks like you got unlucky with a Coveralls failure in the test run there:

Coveralls has been just removed :)
https://github.com/django/djangoproject.com/pull/2182

adamchainz (approved):
Let's gooooo

pauloxnet:
> Note to self (or @bmispelon, or whoever ends up deploying this): We'll need to remove `--update-index` from the Ansible repo at the same time. I'll make a companion PR.

@tobiasmcnulty Do you think there's something that needs to be done on the Ops side before merging this PR? If so, please proceed when you can. (I've no intention of pushing from my side) :)

In fact, let me know if I can make this easier for you in any way. Thanks again for your help.

Issue #1705, opened 2024-10-29
Add next board meeting date to DSF minutes


nanorepublica:
Allow the DSF minutes to include the next board date (when known) to the minutes.

This improves transparency to the wider community.

bmispelon:
Same question as #1704 : does this change need to be approved by the DSF board first, or can someone start writing a PR already?

nanorepublica:
I asked Thibaud in a DM as I wasn't aware the meeting minutes were a model. He gave a thumbs up to both of these changes.

thibaudcolas:
Yes please! This isn’t significant enough to require a board vote.

If anyone wants to proceed with this, please make sure to:

- Make this field optional
- Choose wording to make it clear the date is provisional / can change (we often have special meetings being called, or reschedule just to have more board members available)

thibaudcolas:
So feedback from the board is good, with a suggestion that the exact date could also be replaced or complemented with a generic note that "meetings are generally the second Thursday of the month but are occasionally rescheduled for attendance".

---

So, I think we could add this at the end of the template that displays the minutes:

```
<p>Board meetings are generally the second Thursday of the month but are occasionally rescheduled for attendance.{% if next_meeting_date %} Next meeting scheduled for: {{ next_meeting_date }}{% endif %}</p>
```

With `next_meeting_date`set up as either a date or plain-text field.

thibaudcolas:
cc our 2025 Secretary @knyghty

FarhanAliRaza:
The next_meeting_date should be part of the `Meeting` model ?

sarahboyce:
Given the DSF meeting minutes now live in https://github.com/django/dsf-minutes (see https://github.com/django/djangoproject.com/issues/2079). I suppose we can close this issue and this feedback can be incorporated in the DSF secretary monute writing process (cc @knyghty)

knyghty:
Yes, and I would say it's a little tricky to update since each year the incoming board chooses a new meeting time to make sure it's as convenient as possible for everyone.

adamzap:
Thanks for the insight. I'll close this!

PR #2197, opened 2025-09-12
Add next meeting date field to Meeting Model

house-ben:
Adds next meeting field discussed in #1705 

Updates the foundation sidebar to include the next meeting date to add transparency to the community. At the direction of thibaudcolas discussed in the issue.

image

Example with the last meeting having a `next_meeting_date`

image

Unsure if it would be better to have the next_meeting_date be calculated inside the template.

### DIFF
--- a/foundation/tests.py
+++ b/foundation/tests.py
@@ -106,7 +106,9 @@ class MeetingTestCase(TestCase):
         self.assertContains(response, "Latest DSF meeting minutes")
 
         self.assertContains(response, "DSF Board monthly meeting, May 12, 2023")
-        self.assertContains(response, "The next meeting is scheduled for: June 13, 2023")
+        self.assertContains(
+            response, "The next meeting is scheduled for: June 13, 2023"
+        )
         self.assertContains(response, latest_meeting.get_absolute_url())
         self.assertContains(response, "DSF Board monthly meeting, April 12, 2023")
         self.assertContains(response, previous_meeting.get_absolute_url())


foundation/models.py
```diff
@@ -95,6 +95,7 @@ class Meeting(models.Model):
     )
     treasurer_report = models.TextField(blank=True)
     treasurer_report_html = models.TextField(editable=False)
+    next_meeting_date = models.DateField(null=True, blank=True)
```
tim-schilling:
```suggestion
    next_meeting_date = models.DateField(
        null=True,
        blank=True,
        help_text=_("This will be shown on the website as the board's next meeting date. The latest meeting's Next Meeting Date will be used.")
    )
```
I wonder if a little bit of context will be helpful to future DSF Secretaries.

djangoproject/templates/foundation/meeting_snippet.html
```diff
@@ -23,3 +23,14 @@ <h4>
     </li>
   {% endfor %}
 </ul>
+<p>
+  {% blocktranslate trimmed %}
+    Board meetings are generally the second Thursday of the month but are
+    occasionally rescheduled for attendance.
+  {% endblocktranslate %}
+  {% if next_meeting_date %}
+    {% translate "Next meeting scheduled for: " context "Following meetings schedule" %}
```
tim-schilling:
```suggestion
    {% translate "The next meeting is scheduled for: " context "Following meetings schedule" %}
```

I think if this were a bullet point, I would agree with your more concise version, but since it's part of a paragraph, I think a complete sentence reads a bit better.

foundation/tests.py
```diff
@@ -104,6 +106,7 @@ def test_latest_meeting_minutes(self):
         self.assertContains(response, "Latest DSF meeting minutes")
 
         self.assertContains(response, "DSF Board monthly meeting, May 12, 2023")
+        self.assertContains(response, "Next meeting scheduled for: June 13, 2023")
```
tim-schilling: We may want to include a test for when there is no `next_meeting_date` in the context to confirm "Next meeting schedule for" doesn't appear.

foundation/models.py
house-ben: Updated! I like this idea too!

djangoproject/templates/foundation/meeting_snippet.html
house-ben: Such a better version thanks!

foundation/tests.py
house-ben: Good Idea! I'll add that!

djangoproject/templates/foundation/meeting_snippet.html
```diff
@@ -23,3 +23,13 @@ <h4>
...
   {% endfor %}
 </ul>
+<p>
+  {% blocktranslate trimmed %}
+    Board meetings are generally the second Thursday of the month but are
+    occasionally rescheduled for attendance.
+  {% endblocktranslate %}
+  {% if next_meeting_date %}
+    {% translate "The next meeting is scheduled for: " context "Following meetings schedule" %}{{next_meeting_date|date:"DATE_FORMAT"}}
+  {% endif %}
```
house-ben, line 33: Unsure if there is a better to do this, so that line 32 isn't as long. While not having a `\n` between `The next meeting is scheduled for: ` and the date. I know django `{% spaceless %}` tag would work here, but seems a bit overkill.

house-ben:
Thanks for all the feedback! I implemented all the changes.

tim-schilling (approved):
I'm punting on the line length change. I personally would let it be as is, but I'm not a maintainer of this project. Thanks for making the changes I suggested @house-ben!

adamzap:
Is this PR still relevant now that meeting notes are stored in [a GitHub repository](https://github.com/django/dsf-minutes)?

sarahboyce:
> Is this PR still relevant now that meeting notes are stored in [a GitHub repository](https://github.com/django/dsf-minutes)?

We also are planning to drop these tables soon (https://github.com/django/djangoproject.com/pull/2384), so i agree I don't think it's needed

adamzap:
@house-ben Thank you for this contribution, but it seems that the direction on this has changed. I'm going to close this PR for now.

PR #2199, opened 2025-09-12
Compiled translation when building the docker image

### LINKED ISSUE #2174: Add Swedish translation for djangoproject.com
The website is now translated in Swedish. Please enable this. Only one string is in Swedish so something must be wrong somewhere

bmispelon:
This should fix #2174.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,8 +45,7 @@ RUN apt-get update \
 # copy project
 COPY . .
 
-RUN python -m django compilemessages \
-    && apt-get purge --assume-yes --auto-remove gettext
+RUN python -m django compilemessages
 
 # ENTRYPOINT is specified only in the local docker-compose.yml to avoid
 # accidentally running it in deployed environments.
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,5 +45,8 @@ RUN apt-get update \
 # copy project
 COPY . .
 
+RUN python -m django compilemessages \
+    && apt-get purge --assume-yes --auto-remove gettext
+
 # ENTRYPOINT is specified only in the local docker-compose.yml to avoid
 # accidentally running it in deployed environments.


Dockerfile
```diff
@@ -45,5 +45,8 @@ RUN apt-get update \
 # copy project
 COPY . .
 
+RUN python -m django compilemessages \
+    && apt-get purge --assume-yes --auto-remove gettext
```
bmispelon: `gettext` was only needed because we were doing the `compilemessages` on the server, but with this change it's not needed anymore and we can shave off a few megabytes off the image

tobiasmcnulty: Nitpick: `gettext` is already present in an earlier layer, so I don't think this will reduce the image size. It might be helpful for debugging or if someone wants to run this command during local development in Docker, so I vote for leaving it in the image.

```suggestion
RUN python -m django compilemessages
```

alexgmin: Well, I learned something today, I thought it would but I was wrong.
I tested out of curiosity and these are my results:

```
djangoprojectcom-web 677MB - With gettext purge
djangoprojectcom-web 676MB - Without gettext purge
```

I have no idea how without the last purge is less considering the command is not in a new layer, any ideas?

tobiasmcnulty: @alexgmin Hmm, I believe the earlier `RUN` command that installs `gettext` is its own layer, and the later command that removes `gettext` is another layer, so the final image includes two (plus) layers, one with and one without `gettext`.

It's also worth noting that there are some references in the README that say `gettext` is required by the `update_docs` management command, although I couldn't verify that it actually breaks without `gettext`.

All of that to say, I don't think it's worth holding up the `compilemessages` fix while we try to reduce the image size (let alone the risk of removing it without more testing)... Let's perhaps make a separate issue for reducing the image size and we can proceed with that at its own pace.

Issue #2172, opened 2025-08-19
Do we still use Coveralls? https://coveralls.io/github/django/djangoproject.com


ulgens:
Do we still use Coveralls? https://coveralls.io/github/django/djangoproject.com

_Originally posted by @adamzap in https://github.com/django/djangoproject.com/issues/2135#issuecomment-3192985697_

SaptakS:
I don't know honestly. Maybe someone with more historical context can answer when it was added. I do like the idea of having a coverage tool to ensure that we have decent amount of tests for the features, but I don't think we need to depend on Coveralls for that. I think we can even just use coverage in the CI itself.

adamzap:
@bmispelon Would you be in favor of us removing references to Coveralls and deleting our account there?

ulgens:
I think it is useful to track coverage details, stats and history. Github doesn't keep or present that data.

bmispelon:
I've got no particular attachment to coveralls myself, but if others are finding it useful then I'm not opposed to keeping it either.

I personally don't find test coverage to be a useful metric, especially when it's low (like I assume ours is), and double especially when it's not enforced in CI.
I'd be more in favor of keeping coveralls (or a tool like it) if we had a target to reach 100%, and a CI check to make sure coverage never decreases.

adamzap:
@ulgens In the recent Website WG meeting, we decided that we don't need Coveralls. We can always add better tooling around coverage if there is a desire in the future.

ulgens:
Coming from the meeting notes:

> It is possible to run coverage locally, we could even report the coverage % on PRs and drop Coveralls

Not really. Coverall's value is to keep detailed and historical stats. GitHub interface doesn't provide it.

> Coverage currently is pretty low, not something we are very proud to showcase

All the more reason to fix the integration. We should fix it instead of burying the metrics.

> There is no opposition from members present to removing Coveralls

I wasn't present in the meeting, but I'm strongly against removing it. We can fix the integration / CI or we can replace with a different tool but removing it would be a step in to wrong direction.

P.S.: If the majority's decision is to remove it, I have to agree, but it looks like there is a slight misunderstanding about the value provided by the tool - hence my push for keeping the discussion open.

alexgmin:
If @ulgens is a strong vote against removing I'm okay with keeping the PR and discussion open until the next meeting.
There wasn't a strict vote on it during the meeting so I don't think we should force a decision on this.

tobiasmcnulty:
[...]

PR #2201, opened 2025-09-14
Enforced test coverage via CI

tobiasmcnulty:
Ensures test coverage does not fall below 80% (currently 81%) via `make ci`. This limit can be increased as test coverage is added.


ulgens:
My understanding is this was an easy & quick & dirty alternative for failing Coveralls integration. Now that Coveralls is working again, I don't think this is needed. We can, of course, merge it but I don't think the result will be meaningful. 
Coverage percentage, as a single numeric value, is as misleading as the number of lines of code. It doesn't give information about the history or context. I'm okay with us having 75% test coverage, or even 50%, but it's more important to understand an individual PR's effect on that value - which is one of the major benefits of tools like Coveralls.

tobiasmcnulty:
@ulgens To close the loop on this one, do you object to enforcing no decreases in test coverage via CI? Or do your comments and downvote stem from preferring to keep Coveralls or a different tool? As mentioned [here](https://github.com/django/djangoproject.com/issues/2172#issuecomment-3299464361), this was the solution preferred by the Website WG during its last meeting, not an "easy & quick & dirty alternative" when Coveralls went down. This may go without saying, but I also invite you to propose some positive changes with the coverage solution you would like to see. I don't think these two ideas are mutually exclusive. 🙂

adamzap (approved):
👍 This looks good to me since I hope any new code would include tests.

ulgens:
@tobiasmcnulty 

> enforcing no decreases in test coverage via CI

I don't agree that this change does what you describe here.

Regardless of the coveralls situation, I do think that this is not a good metric to follow up and enforce. As I mentioned earlier, the thing that provides value to the development process is the understanding of what is covered and how it changes, rather than an arbitrary numeric value about how much is covered.

As a general take on the topic, I don't think this discussion is going healthy. There was a weird push and rush to remove the integration without proper reasoning, and things were being framed as "everyone wants to remove it!" when it's not the case. Regardless of the final result, I believe we can and should have more respectful discussions instead of acting like one party has the complete authority and the other one is wrong by default.

tobiasmcnulty:
Hi @ulgens @adamzap, I get the sense we should wait to decide on this PR until we can discuss it at the next website working group meeting. How does that sound as a next step?

adamzap:
@tobiasmcnulty Sounds good, thanks!

CuriousLearner (approved):
This looks like a sensible change to me. It promotes adding tests for any new feature and ensures a healthy test suite.

marksweb (approved):
This is my approach to coverage so I'm on board 👍

ulgens (changes requested):
Just realized I had my comments but didn't actually review the PR.

Context-wise, this feels like a quick workaround for the rushed removal of coveralls, and I don't agree that checking for an arbitrary coverage percentage is a good practice. Considering that we already agreed to consider a coveralls replacement, I don't think this PR contributes positive value.

pauloxnet (approved):

SaptakS (approved):
I think I actually like this approach, promoting people to write more tests. I agree with having CI enforce that any further code we add has some tests for it. @tobiasmcnulty have you tested that this failing only applies to the newly added code in a PR, and not the entire codebase maybe?

tobiasmcnulty:
> I think I actually like this approach, promoting people to write more tests. I agree with having CI enforce that any further code we add has some tests for it. @tobiasmcnulty have you tested that this failing only applies to the newly added code in a PR, and not the entire codebase maybe?

Hi @SaptakS, thanks for the review. This check applies to the whole codebase.

sarahboyce:
I would like to propose that we close this PR in favor of https://github.com/django/djangoproject.com/pull/2325
You can remove code and end up reducing test coverage, having to improve test coverage before you can do this seems harsh. An example of a PR that could have come across a scenario like this would be https://github.com/django/djangoproject.com/pull/2384

Issue #2193, opened 2025-09-11
Docker only install instructions run into database connection error `trac` does not exist


easherma:
When trying to follow _just_ the docker instructions, we ran into an issue. It seems to be related to some combo of the installation instructions missing some steps that would still be relevant for Docker or some outdated or commented out lines in entrypoint. 

Traceback for future reference: 
```
docker compose run --rm web python -m manage test
[+] Creating 1/1
 ✔ Container djangoprojectcom-db-1  Running                                                                        0.0s 
/usr/local/lib/python3.12/site-packages/django_hosts/__init__.py:2: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  import pkg_resources
/usr/local/lib/python3.12/site-packages/django_hosts/__init__.py:2: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  import pkg_resources
Operations to perform:
  Apply all migrations: accounts, admin, aggregator, auth, blog, contenttypes, dashboard, debug_toolbar, docs, flatpages, foundation, fundraising, members, redirects, registration, releases, sessions, sites, subscriber, thumbnail, tracdb
Running migrations:
  No migrations to apply.
python -m pysassc djangoproject/scss/output.scss djangoproject/static/css/output.css --style=compressed
/usr/local/lib/python3.12/site-packages/django_hosts/__init__.py:2: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  import pkg_resources
Deleting 'robots.docs.txt'
Deleting 'robots.txt'
Deleting 'css/output.css'
Deleting 'admin/css/responsive_rtl.css'
Deleting 'admin/css/base.css'
Deleting 'admin/css/forms.css'
Deleting 'admin/css/unusable_password_field.css'
Deleting 'admin/css/rtl.css'
Deleting 'admin/css/dashboard.css'
Deleting 'admin/css/widgets.css'
Deleting 'admin/css/dark_mode.css'
[...]

PR #2207, opened 2025-09-16
Streamline Docker setup and run tests via Docker

tobiasmcnulty:


.github/workflows/docker-test-build.yml
```diff
@@ -25,10 +25,21 @@ jobs:
...
                   push: false
                   build-args: |
                       REQ_FILE=requirements/${{ matrix.req_file }}
+
+            - if: matrix.req_file == 'tests.txt'
+              name: Run the tests via Docker
+              run: |
+                  docker compose up -d db
+                  # Wait for Postgres to be ready
+                  docker compose exec db sh -c 'until pg_isready ; do sleep 1; done'
```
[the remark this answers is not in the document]
tobiasmcnulty: `pg_isready` needs to run in the same container as Postgres itself.

tobiasmcnulty:
Hi, @easherma! Thanks again for your helping improving the Docker setup at the sprints. Would you be up for giving this a review when you have a moment? No rush.

Makefile
```diff
@@ -30,3 +30,10 @@ test:
 
 watch-scss:
 	watchmedo shell-command --patterns=*.scss --recursive --command="make compile-scss-debug" $(SCSS)
+
+reset-local-db:
```
ulgens, line 37: I think we can move the `migrate` command too.

ulgens (commented):
I added a small comment about the migrate command call, but other than it looks nice 🌻

Makefile
tobiasmcnulty: Personally I don't mind leaving `migrate` where it is: Running migrations on container startup seems far less detrimental than wiping all the data from the database, and it might even be expected by some developers.

What do others think? @django/django-website

pauloxnet (approved):

.github/workflows/docker-test-build.yml
```diff
@@ -25,10 +25,24 @@ jobs:
...
+            - if: matrix.req_file != 'tests.txt'
+              name: Test docker image build (${{ matrix.req_file }})
               uses: docker/build-push-action@v6
               with:
                   context: .
                   push: false
                   build-args: |
                       REQ_FILE=requirements/${{ matrix.req_file }}
+
+            - if: matrix.req_file == 'tests.txt'
```
ulgens, line 37: Does this affect the existing test workflow? It looks like we will be running tests in CI twice.

.github/workflows/docker-test-build.yml
```diff
@@ -25,10 +25,24 @@ jobs:
...
+              uses: docker/setup-compose-action@v1
+
+            - if: matrix.req_file == 'tests.txt'
+              name: Run the tests via Docker
+              run: |
+                  docker compose build
+                  docker compose up -d db
+                  # Wait for Postgres to be ready
+                  docker compose exec db sh -c 'until pg_isready ; do sleep 1; done'
+                  docker compose run --rm web make ci
```
ulgens, line 45: Does this affect the existing test workflow? It looks like we will be running tests in CI twice.

Makefile
ulgens: > Running migrations on container startup 

This creates problems when the contributor doesn't want to immedietaly apply the migrations. I think it's better that non of the data related operations are applied automatically, but I agree that contexts are different 👍🏻

.github/workflows/docker-test-build.yml
tobiasmcnulty: > It looks like we will be running tests in CI twice.

That's correct, as a regression test for #2193 / assurance that the `docker compose` instructions aren't broken again.

Longer term it might be nice to consolidate the settings so they're the same for local dev docker and non-docker, and run tests only in docker, but I don't mind running them twice in the interim.

ulgens: > assurance that the docker compose instructions aren't broken again.

Wouldn't be enough to run `check` to validate the setup works? I feel like `make ci`  is a bit of overkill.

.github/workflows/docker-test-build.yml
```diff
@@ -25,10 +25,24 @@ jobs:
...
+              name: Set up Docker Compose
+              uses: docker/setup-compose-action@v1
+
+            - if: matrix.req_file == 'tests.txt'
+              name: Run the tests via Docker
+              run: |
+                  docker compose build
+                  docker compose up -d db
+                  # Wait for Postgres to be ready
+                  docker compose exec db sh -c 'until pg_isready ; do sleep 1; done'
```
ulgens: Does Github somehow bypasses the healthcheck definition in the compose file? Why do we need to define this again?

.github/workflows/docker-test-build.yml
tobiasmcnulty: `check` wouldn't have caught the Docker-only test failures fixed by [the change to `PARENT_HOST`](https://github.com/django/djangoproject.com/pull/2198/files#diff-b52b3077b98ef576a83107a5a99dea1cf4370bbf3c08b51a84595243d4dc8ff2R31), but it looks like we can remove that altogether now (done).

I don't see the downside to running tests with both configurations temporarily, but I also don't mind removing `make ci` here if it seems unnecessary.

.github/workflows/docker-test-build.yml
tobiasmcnulty: Good catch, I think it's not needed.

.github/workflows/docker-test-build.yml
tobiasmcnulty: Any other feedback on this (either way), @django/django-website ?

SaptakS:
@tobiasmcnulty makes sense to me. I think you need to rebase and resolve conflicts though. Once that is done, I think should be ready for approval.

.github/workflows/docker-test-build.yml
ulgens: My vote here would be to make Docker-based testing the main/only testing suite in CI in long term but I'm not sure it in this PR's scope.

.github/workflows/docker-test-build.yml
```diff
@@ -25,10 +25,21 @@ jobs:
...
 
-            - name: Test docker image build (local development)
+            - if: matrix.req_file != 'tests.txt'
+              name: Test docker image build (${{ matrix.req_file }})
               uses: docker/build-push-action@v6
               with:
                   context: .
                   push: false
                   build-args: |
                       REQ_FILE=requirements/${{ matrix.req_file }}
```
CuriousLearner, line 35: For production, we just build the image, and then nothing else? 🤔

CuriousLearner: I think just building the image will be a weak test (and burn CI minutes). Since we are building the image, we might also want to check the startup, something like:

```
- if: matrix.req_file == 'prod.txt'
  name: Test production image starts successfully
  run: |
    docker compose up -d
    sleep 5
    docker compose ps
    docker compose logs
    # Maybe curl a health endpoint
```

tobiasmcnulty: @CuriousLearner Thanks for the feedback. Having some tests for the production image sounds like a positive change to me. I think it could be its own PR as its not really related to the local docker compose setup.

.github/workflows/docker-test-build.yml
tobiasmcnulty: FWIW, I agree.

ulgens: To unblock the progress, I'm okay with merging as is. We can revisit the topic later when we are confident in Docker-based testing setup.

Makefile
ulgens: I'm okay with merging as is. We need to apply some other improvements too and developer ergonomics deserve their own discussion.

PR #2210, opened 2025-09-19
Use sans-serif font everywhere

### LINKED ISSUE #286: Consistent fonts across www and docs
The "Roboto, Corbel, ..." font on the documentation site is really easy on the eyes and conducive to fast reading/scanning (nice choice).

In contrast I find the font used on www looks nice at first glance but is very difficult to read fast/scan, can I suggest that you use the same font as the docs site?

E.g. open these two pages next to each other and read two paragraphs from each, you immediately understand what I mean:
https://docs.djangoproject.com/en/1.7/intro/overview/
https://www.djangoproject.com/weblog/2014/dec/15/announcing-django-website-redesign/

adamzap:
This change was requested for simplicity and accessibility.

Closes #286.

### DIFF
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -13,7 +13,7 @@ CSS rendered with Libsass 0.7.0
 
 //------------------- Globals
 body {
-    @include serif;
+    @include sans-serif;
     @include font-size(18);
     background: var(--sidebar-bg);
     color: var(--body-fg);
@@ -636,7 +636,6 @@ header {
     }
 
     p {
-        @include serif;
         @include font-size(18);
         color: var(--white-color);
         left: -9999px;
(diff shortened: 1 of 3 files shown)


CuriousLearner (commented):
Looks good! But I don’t see a finalized decision to standardize on a sans-serif: #286 is still open, and the 2022 switch to a system sans-serif was reverted (#1243). Are we standardizing on sans-serif now, or should we wait for a design decision?

adamzap:
@CuriousLearner The website working group decided that the next step would be to open a PR like this. We'll see what others think (@sabderemane and @SaptakS especially since the three of us discussed it last month).

pauloxnet (approved):

SaptakS:
@CuriousLearner I think the revert that you mentioned was more related to using system fonts rather than san-serif fonts itself, if I read the PR correctly.

Issue #2190, opened 2025-09-11
Restore `psql` to Docker image


tobiasmcnulty:
The Postgres client was removed from the Docker image in https://github.com/django/djangoproject.com/commit/1bb286daa80cb8b5d5cc863e711499d3c30e60a3

It's necessary to have `psql` in the image for running the database setup instructions on the preview server (e.g., creating the Trac database schema) or performing ad-hoc backups on the production server.

The version of psql should match the current production database instance (currently 14, soon to be 17).

The specific version can be installed from the Postgres apt repo: https://www.postgresql.org/download/linux/debian/

PR #2212, opened 2025-09-20
Use the binary version of psycopg for development

ulgens:
Replaces `psycopg[c]`  with `psycopg[binary]` for development.

Relates to https://github.com/django/djangoproject.com/issues/2215

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,6 +9,7 @@ ENV PYTHONDONTWRITEBYTECODE=1
 ENV PYTHONUNBUFFERED=1
 
 # install deb packages
+# PostgreSQL dependencies are needed for dbshell and backup process
 RUN apt-get update \
     && apt-get install --assume-yes --no-install-recommends \
         gettext \
@@ -23,23 +24,14 @@ RUN apt-get update \
     && apt-get distclean
 
 ARG REQ_FILE=requirements/prod.txt
+ARG BUILD_DEPENDENCIES="g++ gcc libc6-dev libpq-dev zlib1g-dev"
 
 # install python dependencies
 COPY ./requirements ./requirements
 RUN apt-get update \
-    && apt-get install --assume-yes --no-install-recommends \
-        g++ \
-        gcc \
-        libc6-dev \
-        libpq-dev \
-        zlib1g-dev \
+    && apt-get install --assume-yes --no-install-recommends ${BUILD_DEPENDENCIES} \
     && python3 -m pip install --no-cache-dir -r ${REQ_FILE} \
-    && apt-get purge --assume-yes --auto-remove \
-        g++ \
-        gcc \
-        libc6-dev \
-        libpq-dev \
-        zlib1g-dev \
+    && apt-get purge --assume-yes --auto-remove ${BUILD_DEPENDENCIES} \
     && apt-get distclean
 
 # copy project
--- a/requirements/common.txt
+++ b/requirements/common.txt
@@ -14,7 +14,6 @@ Jinja2==3.1.6
 libsass==0.23.0
 Markdown==3.9
 Pillow==11.3.0
-psycopg[c]==3.2.9
 Pygments==2.19.2
 pykismet3==0.1.1
 requests==2.32.5
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,6 +5,7 @@ services:
             dockerfile: Dockerfile
             args:
                 - REQ_FILE=requirements/tests.txt
+                - BUILD_DEPENDENCIES=g++ gcc
         entrypoint: ./docker-entrypoint.dev.sh
         command: python manage.py runserver 0.0.0.0:8000
         volumes:
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -1,4 +1,5 @@
 -r common.txt
 django-debug-toolbar==6.0.0
 pre-commit~=4.3.0
+psycopg[binary]==3.2.10
 watchdog==6.0.0
--- a/requirements/prod.txt
+++ b/requirements/prod.txt
@@ -1,4 +1,5 @@
 -r common.txt
 gunicorn==23.0.0
+psycopg[c]==3.2.10
 redis==6.4.0
 sentry-sdk==2.38.0


Dockerfile
```diff
@@ -13,13 +13,8 @@ RUN apt-get update \
     && apt-get install --assume-yes --no-install-recommends \
         gettext \
         git \
-        libpq5 \
-        postgresql-common \
```
pauloxnet, line 17: Have you checked if we need some of these packages content (es: psql) for some Django code (es: dbshell) ?

BTW, this package has just been added by @tobiasmcnulty in this commit https://github.com/django/djangoproject.com/commit/1c5891bf4af534ebbb3298c906bf5d1743492404 few days ago. I would ask him an opinion on its removal.

Dockerfile
```diff
@@ -13,13 +13,8 @@ RUN apt-get update \
...
         make \
         rsync \
-    && /usr/share/postgresql-common/pgdg/apt.postgresql.org.sh -y \
-    && apt-get install --assume-yes --no-install-recommends \
-        postgresql-client-17 \
```
pauloxnet, line 22: Same as above.

Dockerfile
```diff
@@ -30,16 +25,10 @@ RUN apt-get update \
     && apt-get install --assume-yes --no-install-recommends \
         g++ \
         gcc \
-        libc6-dev \
-        libpq-dev \
-        zlib1g-dev \
```
pauloxnet, line 35: If finally we're going to use psycopg[binary] check if this depenencies are not needed to the compilation of other dependencies (eg: pillow).

If not, and we don't need any more of these system dependencies to compile python packages, I'll invite you to get rid of the whole section of installing debian packages (also compilers), and removing  an leaving only the pip install line.

Dockerfile
ulgens: > Have you checked if we need some of these packages content (es: psql) for some Django code (es: dbshell) ?

Thanks for the PR ref, I missed that. I'll check what we can do without hurting the trac setup. I'm not sure we need the whole suite of pg deps just to import a dump, but in any case, this wouldn't directly correlate with the [binary] package selection - only the dependencies here.

Dockerfile
ulgens: gcc and g++ are required for libsass. I looked for an alternative that doesn't require compiling but couldn't find one yet.

Dockerfile
ulgens: I guess we also need to add tests that validate the import works, considering it's an important part of the setup.

Dockerfile
ulgens: Responded above 🖖

Dockerfile
ulgens: @pauloxnet Sorry for another notification but I checked it again and found this open: https://github.com/django/djangoproject.com/pull/2207 The importing of the dump is handled via the database image, which already has the psql in it. The other reason mentioned in https://github.com/django/djangoproject.com/issues/2190 is

> performing ad-hoc backups on the production server.

I'll need to ask a reference/doc about how we use it because I couldn't find any public information about the mentioned process.

tobiasmcnulty: The playbooks and documentation for deploying djangoproject.com are stored in a private repo.

What problem is removing `psql` from the image solving? That feels like a separate issue from switching to a pre-compiled version of `psycopg2`.

I also think being able to run `python -m manage dbshell` during local dev is handy. Adding a test for that to ensure it's not removed in a future PR sounds like a good idea. I can update #2207 with that, unless you want to handle it here, since as you mention it changes the Trac schema import from the Django container to the Postgres container.

requirements/common.txt
```diff
@@ -14,7 +14,7 @@ Jinja2==3.1.6
 libsass==0.23.0
 Markdown==3.9
 Pillow==11.3.0
-psycopg[c]==3.2.9
+psycopg[binary]==3.2.9
```
tobiasmcnulty: > A “Local installation” [the `[c]` variant] results in a performing and maintainable library. The library will include the speed-up C module and will be linked to the system libraries (libpq, libssl…) so that system upgrade of libraries will upgrade the libraries used by Psycopg 3 too. This is the preferred way to install Psycopg for a production site.

From: https://www.psycopg.org/psycopg3/docs/basic/install.html#local-installation

I understand wanting to avoid compiling `psycopg` for non-Docker local dev, however. What if we split the requirement into `dev.txt` and `prod.txt`, using the `[binary]` and `[c]` variants, respectively, with comments to ensure that the installed versions stay in sync between the two files?

tobiasmcnulty (changes requested):
Thanks for working on this.

As mentioned by @pauloxnet, `psql` was only recently removed and re-added, and should be restored before this is merged.

Also, it appears we can solve the compilation issue for local dev and still use the `[c]` variant in the production image.

Dockerfile
ulgens: > The playbooks and documentation for deploying djangoproject.com are stored in a private repo.

Can you please invite me to that repo? I feel like all website WG members should have access, but we can discuss that in a meeting.

> What problem is removing psql from the image solving? That feels like a separate issue from switching to a pre-compiled version of psycopg2.

My goal was to remove unused dependencies, but @pauloxnet helped me to realise there is a hidden context there. I'll revert that hunk from the pr.

> I also think being able to run python -m manage dbshell during local dev is handy.

I have mixed feelings about this but better to handle in a separate discussion.

> I can update https://github.com/django/djangoproject.com/pull/2207 with that

Thank you. #2207 doesn't make use of psql in the Django image, so I think it wouldn't fit the context there. I'll create a separate issue about this, after checking the Django docs.

requirements/common.txt
ulgens: > A “Local installation” [the [c] variant] results in a performing and maintainable library.

Do we or the PostgreSQL docs have any data supporting this? "It is more performant because you compile it for the CPU arch you need" is a thing people always say, but also almost never the case (because it's hardly ever the bottleneck and modern binaries already include improvements for common archs etc)

> What if we split the requirement into dev.txt and prod.txt, using the [binary] and [c] variants

We can do that but what would be the benefit? And that would also a bit more complex with `pyproject.toml` file.

I'd be okay with adding a note like 

```
# TODO: Check the potential performance improvement with [c]
```
with the dependency definition. I think we should avoid premature optimization, and we can revisit the topic when we have the need and the data.

Dockerfile
pauloxnet: Is libsass requiring both gcc and g++ ?
Since you studied it, can we open a PR for the libsass repo to ask a wheel for recent python versions?

Dockerfile
pauloxnet: > ... I'm not sure we need the whole suite of pg deps just to import a dump ...

BTW libpq or postgres-comm are small libraries, not the whole Postgresql suite.

Dockerfile
pauloxnet: Or we can find an alternate package to replace it.

ulgens: I'll check this again. 

https://github.com/sass/libsass-python says

> Provides prebuilt [wheel](https://www.python.org/dev/peps/pep-0427/) binaries for Linux, Windows, and Mac.

but our build fails with 

```
77.32   × Building wheel for libsass (pyproject.toml) did not run successfully.
...
77.32       error: command 'gcc' failed: No such file or directory
```

when those dependencies are not installed.

This feels more a like bug, rather than a missing feature in the libsass package.

pauloxnet: Thanks for checking.
It seems there isn't a valid python alternative to libsass-python that I know.

requirements/common.txt
tobiasmcnulty: The section I quoted is from the [psycopg documentation](https://www.psycopg.org/psycopg3/docs/basic/install.html#local-installation). I don't personally feel the need to prove or disprove that, when an option exists to solve the local development hurdle without impacting deployed environments.

Dockerfile
ulgens: I meant the whole

```
      libpq5 \
      postgresql-common \
...
    && /usr/share/postgresql-common/pgdg/apt.postgresql.org.sh -y \
    && apt-get install --assume-yes --no-install-recommends \
        postgresql-client-17 \
```

Which can also be a lightweight setup, but it definitely feels heavier when you think it can be removed.

requirements/common.txt
ulgens: Thank you. It seems we have no data to prove that the doc's suggestion applies to our case and I'll say that our default position should be to use the simplest config unless the need for a complex one is justified.

pauloxnet: At this point I would suggest having specific development requirements with the psycopg binary only to help people install it locally (because using Docker the dependencies are already installed) and leave the production requirements as they are now, since there isn't data that disproof what the psycopg developer wrote in the docs.

ulgens: @pauloxnet I still think that our default should be the simplicity and ease of maintenance and not "let's keep it as is if we're neutral" but I'll apply your recommendation so we can proceed with the PR. I think we can revisit the topic when there is data, and I can conduct the necessary benchmark but i'm still worried about our approach here:

* Do we even have database performance issues? Did anyone measure them in the past? Why do we even go the extra mile for?
* Even if, let's say, [c] is 3% more performant, does it actually help us? Can our performance bottleneck be improved by slightly more performant database connections?
* Do we even know what the docs mean by "performing", as in the part saying 

> A “Local installation” results in a performing and maintainable library.

I'd like to know what would be an acceptable data point for us so that we can have a healthier discussion.

pauloxnet: I understand that Tobias, as a member of the Ops team, is in favor of keeping things as they are for the production environment, and I think it's best to leave this decision up to him. In my opinion, the current setup is already simple, because it's already working; it's been in production for years, it works well, and the system libraries are already installed, so I see no obvious reason to make changes that would risk potential performance reductions.

For local development, the situation is different: the only reason to use the binary psycopg is to help new developers with the local installation who aren't familiar with these dependencies and don't use Docker, and for them, I agree we could simplify things. But precisely for these cases, I would differentiate the development and production requirements.

However, I'd be very interested in seeing a data-based comparison from yours to understand the performance penalty and the difference in terms of security.  Personally, I don't have time to do these checks. My experience with many projects tells me that the current installation is problem-free, and I trust the psycopg maintainers. However, that doesn't mean I want to prevent anyone from running tests or demonstrating improvements. 🙂

tobiasmcnulty: It's not merely a performance question. Anecdotally, the version of `psycopg` we're using right now [reverted the build of `libpq`](https://www.psycopg.org/psycopg3/docs/news.html#psycopg-3-2-9) bundled with the `[binary]` package due to connection crashes. There's another warning about the binary packages in 3.2.10 (probably inapplicable to us, but still a warning someone has to read and think about nonetheless). We circumvent that set of problems by simply building the package ourselves using the `[c]` variant and the system libraries in the container where it's deployed.

> I'd like to know what would be an acceptable data point for us so that we can have a healthier discussion.

Again, I don't really feel the need to validate or disprove the `psycopg` documentation here. If you are suggesting that said documentation is misleading or incorrect, the [psycopg](https://github.com/psycopg/psycopg) project is probably a better place for that discussion. Perhaps the authors would consider some documentation changes to clarify when and how the various packages should be used. I looked around a little, and found [an issue on a similar topic](https://github.com/psycopg/psycopg/issues/1156) from a few weeks ago which could be a good starting point.

Good luck, and thanks again for working on this!

ulgens: > It's not merely a performance question. 

Can we please be upfront about our concerns? I'm definitely not saying the conversation shouldn't progress and people shouldn't bring out new points, but if you keep presenting them only when they render the previous discussion meaningless, we can't have a proper conversation.

> the version of psycopg we're using right now [reverted the build of libpq](https://www.psycopg.org/psycopg3/docs/news.html#psycopg-3-2-9) bundled with the [binary] package due to connection crashes

https://github.com/psycopg/psycopg/issues/1088

The issue is Windows and conda-related. We can try to consider that single bug as "binary package is bad," but it would be a long stretch.

> probably inapplicable to us, but still a warning someone has to read and think about nonetheless

I'm trying to understand our use cases, our problems, and the improvements that apply to us. I don't see the benefit of nitpicking issues that we are not even sure are actually related to our case and presenting them as "probably inapplicable but nonetheless...". Let's all try to stay within the scope.

 >  If you are suggesting that said documentation is misleading or incorrect

I was asking what our interpretation of the mentioned docs is, and that question can still be answered. The doc doesn't say it's faster or lighter; it uses a vague term without providing any explanation or data. Yes, I consider that misleading, but definitely not incorrect, because it contains no factual information. Fixing that misleading & vague sentence would be something I'd like to handle separately - not in the context of this PR.  Thanks for pointing out the "an issue on a similar topic", but again, I will prefer to focus on the scope of this PR.

> so I see no obvious reason to make changes that would risk potential performance reductions.

@pauloxnet My point was that there is no actual risk of performance reductions here, and our interpretations of the document we keep pointing to differ. I do think that some rare cases would benefit from "don't fix if it's not broken" approach, especially when the project doesn't have any resources for any additional maintenance work but it's not our case and in our case, it roughly translates to "don't try to improve". I agree that we have different understandings of what can be or should be improved, and that's completely natural. We all provide a unique expertise.

In the end, I see two main takes here (except mine):

* The overall perception of things being too fragile and risky to even experiment with is something we need to get rid of. So far, I have seen similar words under multiple different PRs, but in most cases, we changed those things, and all continued to work fine. We can minimize the perceived risk. I am more than willing to listen if there are any ideas to ease the anxiety of breaking things while keeping the development active (like increasing the coverage, simplifying the workflows, adding new types of checks, automating the deployment, adding a rollback mechanism, etc.)
* Not feeling obligated to validate a point but having authority on it and telling others what to do or not is not healthy. I understand there are risks and some history, but we can't overcome that baggage unless we are willing to research, experiment, and, most importantly, communicate.

ulgens:
>  Is there a reason to keep it at 3.2.9?

No, I don't intend to keep the version outdated. We can update this PR and close the Dependabot PR, or we can merge this PR, and Dependabot will create another PR to update it. My point was that we don't need both PRs to be merged.

requirements/common.txt
ulgens: Now that we have different definitions in prod.txt and dev.txt, I'm marking this as resolved. Thanks for the input.

Dockerfile
ulgens: I reverted the runtime dependency changes. Marking as resolved.

Dockerfile
ulgens: I reverted the runtime dependency changes. Marking as resolved.

ulgens:
I updated the dependency version, separated the definitions for dev and prod, reverted the removal of runtime dependencies, and updated the PR title.

The CI failure doesn't seem related to the changes in the PR and I could use help to handle it: https://github.com/django/djangoproject.com/actions/runs/17961730284#summary-51086239582

Dockerfile
ulgens: Is there any action needed here?

tobiasmcnulty:
> The CI failure doesn't seem related to the changes in the PR and I could use help to handle it:

I reverted the offending commit for now. Hopefully the build will pass if you rebase again.

ulgens:
> I reverted the offending commit for now. Hopefully the build will pass if you rebase again.

Thanks you. CI is passing now. I moved the PR out of draft status.

Dockerfile
dvarrazzo: The currently distributed version of libsass only includes binary wheels for Python 3.8: https://pypi.org/project/libsass/0.23.0/#files

So in theory this could be installed with no C compiler but in practice it would require the authors to upgrade their CI.

pauloxnet: > The currently distributed version of libsass only includes binary wheels for Python 3.8: https://pypi.org/project/libsass/0.23.0/#files
> 
> So in theory this could be installed with no C compiler but in practice it would require the authors to upgrade their CI.

I think this is the point for libsass

@dvarrazzo since you are here what can you suggest to us for the choice to use or not psycopg[binary] locally and psycopg[c] in production?

dvarrazzo: `[binary]` works well these years and it's supported on major platforms but it's not supported everywhere and wheels creation can't be guaranteed, so I advise against using it as a library dependency. It's only suggested when you know exactly where an app will run (a requirements file, not an abstract dependency).

I was answering about the matter here too: https://github.com/psycopg/psycopg/issues/1156#issuecomment-3369178267

alexgmin: I'd add that this is not a library, this is the project for the Django website.
I'm not sure the website needs the same level of compatibility as the Django framework.

ulgens: > so I advise against using it as a library dependency

This repo is not a library.

>  It's only suggested when you know exactly where an app will run

I'd say this is a given for a website project. I agree with @alexgmin

dvarrazzo: > This repo is not a library.

Oh sorry, I didn't understand it. In this case no problem at all to use `paycopg[binary]` and you can drop the libpq5/libpq-dev dependencies too.

tobiasmcnulty: Hi @dvarrazzo, thank you for joining this conversation and for the in-depth discussion on https://github.com/psycopg/psycopg/issues/1156!

To clarify one point, with this PR, we segmented the requirements so we can use `psycopg[binary]` for local development (to simplify the local setup process) and `psycopg[c]` in production, as this is how I interpreted the recommendation in the [local installation](https://www.psycopg.org/psycopg3/docs/basic/install.html#local-installation) instructions:

> This is the preferred way to install Psycopg for a production site.

Could you clarify if `psycopg[binary]` is also intended for use in production environments? We control the environment (currently `python:3.12-slim-trixie`, plus pgdg for the correct `postgresql-client` version if that matters), but that will of course change as we upgrade Python & Debian releases over time.

Thanks so much for your feedback!

ulgens: About the problem with libsass, I created an issue on their repo, and it got closed as "completed" but without a fix.

- https://github.com/sass/libsass-python/issues/486

ulgens: libsass follow up: https://github.com/django/djangoproject.com/pull/2248#issuecomment-3389897033

dvarrazzo: Hello, sorry for the radio silence, I'm crushed by my backlog...

The reason for declaring `[c]` preferred is mostly related to release upgrades: updating the system libpq or openssl `[c]` will use the updated libraries, `[binary]` will not and a new release of Psycopg would be required.

Performance and stability-wise I don't expect differences.

If your service is controlled by a grumpy, customarily bearded, sysadmin who wants to have control on every system library, using `[c]` you will use the system ones; using `[binary]` you will get whatever packaged in the `psycopg-binary` build.

ulgens: > The reason for declaring [c] preferred is mostly related to release upgrades: updating the system libpq or openssl [c] will use the updated libraries

> If your service is controlled by a grumpy, customarily bearded, sysadmin who wants to have control on every system library, using [c] you will use the system ones

> Performance and stability-wise I don't expect differences.

@django/django-website My understanding is unless we need to manage libpq and openssl versions, [c] doesn't seem necessary. Do we still want to keep the current config or would we like to proceed with simplifaction / cleaning?

PR #2213, opened 2025-09-20
Added `LOCALE_PATHS` setting

### LINKED ISSUE #2174: Add Swedish translation for djangoproject.com

tobiasmcnulty:
Fixes #2174

Screenshot 2025-09-20 at 8 18 29 AM

### DIFF
--- a/djangoproject/settings/common.py
+++ b/djangoproject/settings/common.py
@@ -217,8 +217,16 @@ TEMPLATES = [
 
 TIME_ZONE = "America/Chicago"
 
+# Internationalization settings
+
 USE_I18N = True
 
+LOCALE_PATHS = [
+    # List apps first so their translations take precedence.
+    # This list should match LOCALE_DIRS in update-translations.sh.
+    str(BASE_DIR.joinpath(path))
+    for path in ("dashboard/locale/", "docs/locale/", "locale/")
+]
 
 USE_TZ = False
 
--- a/update-translations.sh
+++ b/update-translations.sh
@@ -6,6 +6,7 @@
 
 set -ex
 
+# This list should match the LOCALE_PATHS in djangoproject/settings/common.py
 LOCALE_DIRS="dashboard/locale/ docs/locale/ locale/"
 
 tx pull -a -f --minimum-perc=70


djangoproject/settings/common.py
```diff
@@ -217,8 +217,16 @@
...
 
+# Internationalization settings
+
 USE_I18N = True
 
+LOCALE_PATHS = [
+    # List apps first so their translations take precedence.
+    # This list should match LOCALE_DIRS in update-translations.sh.
+    str(BASE_DIR.joinpath(path))
+    for path in ("dashboard/locale/", "docs/locale/", "locale/")
```
bmispelon: Django already loads `locale` directories from installed apps: https://docs.djangoproject.com/en/5.2/topics/i18n/translation/#how-django-discovers-translations

So I think this can be simplified to:
```suggestion
    BASE_DIR.joinpath("locale"),
```

bmispelon (commented):
I've tested this both locally and on the preview server and it appears to work, but I think the `LOCALE_PATHS` setting is more verbose than it needs to be.

To test it I used the [locale switcher add-on](https://addons.mozilla.org/en-US/firefox/addon/locale-switcher/) for firefox and the swedish (`sv`) language.

djangoproject/settings/common.py
tobiasmcnulty: Thanks for the feedback. I intended to address that with this comment:

```
# List apps first so their translations take precedence.
```

But, simpler sounds good and we can add them if that ever actually becomes an issue.

djangoproject/settings/common.py
```diff
@@ -217,8 +217,14 @@
...
 TIME_ZONE = "America/Chicago"
 
+# Internationalization settings
+
 USE_I18N = True
 
+# Django discovers locale directories in the dashboard and docs apps
+# on its own, but the main project locale/ directory needs to be
+# explicitly listed here.
+LOCALE_PATHS = [str(BASE_DIR.joinpath("locale/"))]
```
ulgens: I'd go with 

```python
LOCALE_PATHS = [
    BASE_DIR / "locale/",
]
```

path syntax for the sake of simplicity and a trailing comma so that the diff for the next change will be smaller.

ulgens: I think we don't even need [] here, a tuple would solve but completely okay if we don't want to push that.

tobiasmcnulty: `[str(DIR.joinpath("..."))]` is the form used elsewhere in the file (though I guess I have an unnecessary trailing slash that can be removed).

I don't mind changing it to `BASE_DIR / "locale"`, but do we want to change the other instances of `joinpath()` to use this syntax too?

ulgens: >  but do we want to change the other instances of joinpath() to use this syntax too?

My personal opinion is to update all instances. Pathlib is introduced in 3.4, and in general, it makes things easier to read and maintain.

I can handle the other instances if that's okay.

ulgens (approved):

djangoproject/settings/common.py
ulgens: Handled in https://github.com/django/djangoproject.com/pull/2312

PR #2216, opened 2025-09-21
Pin Django to patch version

ulgens:
Helps with reproducible builds


SaptakS:
I think we have enough approval for this PR and haven't heard anyone opposing, so going to merge this.

PR #2222, opened 2025-09-22
Remove SafeText from url so it can be escaped

### LINKED ISSUE #2131: HTML escaping issue with Raw HTML copy button for image uploads
@vladimir-slavov-iob

Providence-o:
Fixes #2131 

`format_html` marks the url string as a SafeText which prevents it from being escaped when the html is rendered


adamzap:
I'm going to close this for now. @Providence-o please reopen it if you plan to address the comment.

PR #2229, opened 2025-09-23
Fix blog archive heading to show "Month Year archive"

ManshaAgarwal716:
### Summary
Fixes the heading on monthly blog archive pages, which was showing only "archive" instead of "Month Year archive".

### Details
- Previously, the h1 tag rendered as just "archive".
- Now it correctly displays e.g. "August 2025 archive".
- This regression was introduced in commit 910e7c8 (#1669).

### Related
Regression from #1669


SaptakS:
@ManshaAgarwal716 can you rebase the PR on main. That should hopefully resolve the failing CI. Let me know if you need any help.

ManshaAgarwal716:
Have done sir, you can once checkout the changes and if anything else is required do let me know.

djangoproject/templates/blog/entry_archive_month.html
```diff
@@ -7,7 +7,7 @@
 
 {% block content %}
 
-  <h1>{% blocktranslate %}{{ month|date:"YEAR_MONTH_FORMAT" }} archive{% endblocktranslate %}</h1>
+  <h1>{{ month|date:"F Y" }} {% translate "archive" %}</h1>
```
bmispelon, line 10: This is not the correct way to make this title translatable, because different languages will use different word order and might want to have the word "archive" appear before the date for example.

The correct fix is to use `{% blocktranslate with ... %}`: https://docs.djangoproject.com/en/5.2/topics/i18n/translation/#blocktranslate-template-tag

SaptakS:
@ManshaAgarwal716 can you take a look and address the changes requested by @bmispelon ?

djangoproject/templates/blog/entry_archive_month.html
bmispelon: The issue still hasn't been resolved, did you forget to push a commit?

michalpokusa:
I would like to add that the same issue is present in a per day view, it uses a separate template so it might be worth fixing it also there.
https://www.djangoproject.com/weblog/2025/nov/7/

image

adamzap:
@ManshaAgarwal716 Do you plan to address [this comment](https://github.com/django/djangoproject.com/pull/2229#discussion_r2408259249)?

sarahboyce:
I am closing this in favor of https://github.com/django/djangoproject.com/pull/2380 
I have addressed the comments and written a test :+1:

PR #2233, opened 2025-09-30
Automatically generate Django series roadmaps from Release data

nessita:
This branch adds a new view that replaces the Roadmap wiki pages, e.g.:

  * https://code.djangoproject.com/wiki/Version5.2Roadmap
  * https://code.djangoproject.com/wiki/Version6.0Roadmap
  * https://code.djangoproject.com/wiki/Version6.1Roadmap

The wiki pages were almost identical except for version-specific details and target dates. This information is now generated dynamically from the Release model in the database.

The supported and future Roadmap tables were reorganized so they follow a more intuitive timeline (from older to newer versions).

### DIFF
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -2529,14 +2529,48 @@ dl.data {
     white-space: normal;
 }
 
+table {
+    th {
+        background: var(--sidebar-bg);
+        font-weight: bold;
+        text-align: left;
+    }
+
+    td {
+        border-bottom: 1px solid var(--hairline-color);
+    }
+
+    td,
+    th {
+        padding: 0.5em 1em;
+    }
+}
+
 table.foundation td {
-    border-bottom: 1px solid var(--hairline-color);
     padding: 0 5px;
 }
 
-table.docutils td,
-table.docutils th {
-    border-bottom: 1px solid var(--hairline-color);
+table.django-supported-versions,
+table.django-unsupported-versions {
+    border: 1px solid var(--hairline-color);
+    color: var(--table-color);
+
+    th,
+    td {
+        border-bottom: none;
+        padding: 5px;
+        text-align: center;
+    }
+}
+
+table.django-supported-versions th,
+table.django-supported-versions tr {
+    background-color: var(--secondary-accent);
+}
+
+table.django-unsupported-versions th,
+table.django-unsupported-versions tr {
+    background-color: var(--error-light);
 }
 
 .list-links {
@@ -3418,26 +3452,6 @@ form .footnote {
         display: inline-block;
         margin-left: 1em;
     }
-}
-
-table.django-supported-versions,
-table.django-unsupported-versions {
-    border: 1px solid black;
-    text-align: center;
-    color: var(--table-color);
-
-    th,
-    td {
-        padding: 5px;
-    }
-}
-
-table.django-supported-versions tr {
-    background-color: var(--secondary-accent);
-}
-
-table.django-unsupported-versions tr {
-    background-color: var(--error-light);
 }
 
 /* Corporate membership list page */
--- a/djangoproject/templates/releases/download.html
+++ b/djangoproject/templates/releases/download.html
@@ -85,10 +85,10 @@
       <th>End of extended support<sup><a href="#ft2">2</a></sup></th>
     </tr>
     <tr>
-      <td>5.2 LTS</td>
-      <td>{% get_latest_micro_release '5.2' %}</td>
-      <td>December 2025</td>
-      <td>April 2028</td>
+      <td>4.2 LTS</td>
+      <td>{% get_latest_micro_release '4.2' %}</td>
+      <td>December 4, 2023</td>
+      <td>April 2026</td>
     </tr>
     <tr>
       <td>5.1</td>
@@ -97,10 +97,10 @@
       <td>December 2025</td>
     </tr>
     <tr>
-      <td>4.2 LTS</td>
-      <td>{% get_latest_micro_release '4.2' %}</td>
-      <td>December 4, 2023</td>
-      <td>April 2026</td>
+      <td>5.2 LTS</td>
+      <td>{% get_latest_micro_release '5.2' %}</td>
+      <td>December 2025</td>
+      <td>April 2028</td>
     </tr>
   </table>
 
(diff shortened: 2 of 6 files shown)


nessita:
Screenshots for the dark theme follows:
image
image

nessita:
Screenshots for the light theme follows:
image
image

nessita:
Small screens:
image

releases/views.py
```diff
@@ -31,6 +31,17 @@ def index(request):
     return render(request, "releases/download.html", context)
 
 
+def roadmap(request, series):
+    major, minor = series.split(".")
+    releases = Release.objects.filter(major=major, minor=minor, micro=0)
```
jacobtylerwalls: Are you comfortable without an upper bound here, i.e. letting /download/99.2/roadmap/ be crawlable? It might be quick & easy to just render a different template? I'd be fine with "that version of Django is not yet planned" or similar.

A lower bound could be good, too, since these roadmaps don't apply to Django 1.2 (or Django 1.3, which won't even reach this view)

releases/urls.py
```diff
@@ -1,10 +1,11 @@
...
 
-from .views import index, redirect
+from .views import index, redirect, roadmap
 
 urlpatterns = [
     path("", index, name="download"),
     re_path(
         "^([0-9a-z_.-]+)/(tarball|wheel|checksum)/$", redirect, name="download-redirect"
     ),
+    re_path(r"^(?P<series>\d{1,2}\.[0-2])/roadmap/$", roadmap, name="roadmap"),
```
jacobtylerwalls, line 10: Note to self: we should update link to the roadmap in the django docs to point here instead.

djangoproject/templates/releases/download.html
```diff
@@ -85,10 +85,10 @@ <h2 id="supported-versions">Supported Versions</h2>
       <th>End of extended support<sup><a href="#ft2">2</a></sup></th>
     </tr>
     <tr>
-      <td>5.2 LTS</td>
```
jacobtylerwalls, line 88: Nice, I see now how the prior version "zizagged" against itself and the chart.

djangoproject/scss/_style.scss
```diff
@@ -2530,14 +2530,51 @@ dl.data {
     white-space: normal;
 }
 
+table {
+    color: var(--body-fg);
+
+    th {
+        background: var(--sidebar-bg);
+        font-weight: bold;
+        text-align: left;
```
jacobtylerwalls, line 2536: This change looks a bit off:

Image

djangoproject/scss/_style.scss
```diff
@@ -2530,14 +2530,51 @@ dl.data {
     white-space: normal;
 }
 
+table {
+    color: var(--body-fg);
+
+    th {
+        background: var(--sidebar-bg);
```
jacobtylerwalls, line 2534: Lifting these styles up is going to affect flatpages, too, but I guess we can keep an eye on any unexpected changes.

djangoproject/scss/_style.scss
```diff
@@ -2530,14 +2530,51 @@ dl.data {
     white-space: normal;
 }
 
+table {
+    color: var(--body-fg);
```
jacobtylerwalls: Can you add the reason why this color rule is here? Just want to check it.

djangoproject/scss/_style.scss
nessita: Yeah exactly, I want a global CSS for tables otherwise each flatpage needs its own CSS style for pages (most recent case being the cve/cna page at https://www.djangoproject.com/foundation/cve-numbering-authority/ (see source file).

I checked the other (non flatpages) that we have though.

djangoproject/scss/_style.scss
nessita: How do you mean? off in aligment or content or..?

djangoproject/templates/releases/download.html
nessita: Yeah, exactly!!! It was hard to follow a timeline top-bottom.

releases/urls.py
nessita: Yes! Also the checklist generator. And I was planning on replacing all wiki pages content (2.0 and above) with links to their corresponding roadmap pages in djangoproject.com)

releases/views.py
nessita: Good points. About upper bounds, I figured it wouldn't be so bad but I did not think about crawlers. Still these would only be 99*3 requests... does not seem like much. But I'm happy to limit, did you have a upper limit in mind?

For lower bound, I'll require >= 2.0.

djangoproject/scss/_style.scss
nessita: I was going to answer "this is to have decent text color in the tables for both light and dark themes" but I can't reproduce having incorrect text colors in the table no matter what theme. On a second thought, I think we can remove it (I'll do so now.)

djangoproject/scss/_style.scss
jacobtylerwalls: Alignment -- the values are centered but the header is left-aligned.

releases/views.py
jacobtylerwalls: I leave it to you, I don't have a strong view. We don't want to have to maintain this frequently so idk.

nessita: As agreed with Jacob will add a lower limit but not upper.

djangoproject/scss/_style.scss
nessita: Great eye! It was very subtle. Fixed now (high zoom is intentional to see the alignments better):
image

djangoproject/templates/releases/roadmap.html
```diff
@@ -0,0 +1,120 @@
...
+          </tr>
+        </tbody>
+      </table>
+    </div>
+  </section>
+
+  <section id="process">
+    <h2>Process</h2>
+    <p>Any features not completed by the feature freeze date won't make it into {{ series }}.</p>
+    <p>The release manager will keep the schedule updated and ensure efficient
```
jacobtylerwalls, line 73: I see we're no longer documenting the release manager anywhere, yes?

nessita: For now, we are not. When we merge the checklist generator in djangoproject.com, we could make use of the "releaser" info from a release (the final one, for example) and include that here perhaps? What do you think?

jacobtylerwalls: Oh I didn't realize that would be blocked on ingesting the checklist generator. Sounds good!

nessita: Right, my point is that "release manager" info is not stored anywhere AFAIK.

PR #2234, opened 2025-10-01
Upgrade Python to 3.13

ulgens:
Coming from https://github.com/django/djangoproject.com/issues/1890#issuecomment-3167984510

> > If that's okay, I can proceed with 3.13 update and demonstrate why we don't need the matrix anymore.
> 
> Absolutely. It should be a separate PR than removing tox though, the two things might be connected but we should still do them one at a time.

cc @bmispelon

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,7 +8,7 @@ on:
     pull_request:
 
 env:
-    PYTHON_VERSION: 3.12
+    PYTHON_VERSION: 3.13
 
 jobs:
     tests:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -32,7 +32,7 @@ repos:
       rev: v3.21.2
       hooks:
           - id: pyupgrade
-            args: [--py312]
+            args: [--py313]
 
     - repo: https://github.com/adamchainz/django-upgrade
       rev: 1.29.1
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # pull official base image
-FROM python:3.12-slim-trixie
+FROM python:3.13-slim-trixie
 
 # set work directory
 WORKDIR /usr/src/app
--- a/README.rst
+++ b/README.rst
@@ -12,7 +12,7 @@ To run locally, you can either:
 Install and run locally from a virtual environment
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-#. Create a `Python 3.12 virtualenv and activate it <https://docs.python.org/3/library/venv.html>`_
+#. Create a `Python 3.13 virtualenv and activate it <https://docs.python.org/3/library/venv.html>`_
 
 #. Install dependencies::
 


ulgens:
@pauloxnet Thanks for the exteremely quick response 😄 I was waiting to the see the CI results but I guess one potential issue would be: https://github.com/django/djangoproject.com/pull/2105 That PR needs a rebase but I don't have the permission to do that.

alexgmin:
@bmispelon Will have to confirm, but from my memory, the reason we're not using 3.13 is that Trac doesn't work on that version since it used some of the standard library removals, and Trac and the Website run on the same python binary right now.

Unless that has changed at the operations level, I don't think we can merge this.

Edit: I can confirm from an old forum message of mine the trac issue
> Trac right now supports Python 3.5-3.11, and the current version won't run in 3.13 due to modules like crypt being removed.

ulgens:
@alexgmin 

> but from my memory, the reason we're not using 3.13 is that Trac doesn't work on that version since it used some of the standard library removals, and Trac and the Website run on the same python binary right now.

You are right, but after that discussion, we also agreed that it's not that important and we can proceed with the update. I don't remember the exact thread where it happened so I can't share the link, sorry. In any case, I'm okay with @bmispelon 's decision.

bmispelon:
I remember writing a more detailed explanation somewhere, but I can't find it at the moment 🤷🏻 

I _think_ it's safe for the site and Trac to run separate Python versions. It's more important that they run the same Django version (because they share a database, and technically Django doesn't support running two different versions talking to the same db), but even that is not a hard requirement.

Stormheg:
Sounds like Trac could use some attention as well then to keep it up-to-date with the latest Python - is there a tracking issue for that?

ulgens:
> Sounds like Trac could use some attention as well then to keep it up-to-date with the latest Python - is there a tracking issue for that?

The last time I checked, it seemed like Trac was in a maintenance-only mode, and we were already discussing the move of issues to GitHub. I tried to find the old discussions and found these: 
- https://github.com/django/code.djangoproject.com/pull/219#issuecomment-2452482455
- https://github.com/django/code.djangoproject.com/pull/219#issuecomment-2453461847

It seems [the Trac issue](https://trac.edgewall.org/ticket/13774) @pauloxnet commented on had no updates for the last 11 months and Trac's release calendar has a delay of 21 months: https://trac.edgewall.org/milestone/1.6.1 The Github repo seems like it was getting new commits until 6 months ago. All in all, I think it's hard to say Trac is in active development, and I don't expect to get any new Python version support in the foreseeable future.

pauloxnet (approved):

pauloxnet:
> > Sounds like Trac could use some attention as well then to keep it up-to-date with the latest Python - is there a tracking issue for that?
> 
> The last time I checked, it seemed like Trac was in a maintenance-only mode, and we were already discussing the move of issues to GitHub. ...
> It seems [the Trac issue](https://trac.edgewall.org/ticket/13774) @pauloxnet commented on had no updates for the last 11 months and Trac's release calendar has a delay of 21 months: https://trac.edgewall.org/milestone/1.6.1 The Github repo seems like it was getting new commits until 6 months ago. 

Since Trac has the history of the Django project we can think to contribute to it to let it survive, at least the Python version upgrade.

We can also ask the Django community to help the project if we can't.

The goal here is to upgrade Trackat least to the Python 3.12 to be able to upgrade to Django 6.x

ulgens:
> Since Trac has the history of the Django project we can think to contribute to it to let it survive, at least the Python version upgrade.

I mostly agree with this, but do we have a long-term plan here? What will happen next year? I feel like the more we postpone ripping off the band-aid, the worse the situation gets.

ulgens:
About

>  It's more important that they run the same Django version (because they share a database, and technically Django doesn't support running two different versions talking to the same db)

https://github.com/django/code.djangoproject.com/pull/274

pauloxnet:
> > Since Trac has the history of the Django project we can think to contribute to it to let it survive, at least the Python version upgrade.
> 
> I mostly agree with this ...

:)

> ... but do we have a long-term plan here? What will happen next year? I feel like the more we postpone ripping off the band-aid, the worse the situation gets.

I was answering for the short-term issue, but this is an interesting point to propose for a further discussion.

ulgens:
> ...I guess one potential issue would be: https://github.com/django/djangoproject.com/pull/2105

django-hosts update is handled in https://github.com/django/djangoproject.com/pull/2253

bmispelon:
I've just deployed this on the preview server for a final check, but I think this looks good to merge. 3.14 next? 😁

ulgens:
@bmispelon Thank you 🌻

I can help with 3.14 update too but is there any concerns about Python - Django - Trac version compability between this repo and code.djangoproject.com? I feel like there will be a hard limit to the updates we can do, unless Trac project makes a move.

ulgens:
@bmispelon Hey again and sorry for the ping, would you mind giving your approval if everything is okay on preview server?

PR #2247, opened 2025-10-08
dev: shell script to aid development with docker

marksweb:
This adds a shell script to aid in the development of the project using docker.

Commands would look something like;

```
./do.sh build
./do.sh start
```

Then if you need to interact with the app or migrate when it's running there's things available like;

* `./do.sh shell`
* `./do.sh makemigrations`
* `./do.sh migrate`

There is some talk on slack about removing the entrypoint script, which currently runs `migrate`. But with this approach migrating is the responsibility of the developer, rather than an automatic startup job.

### DIFF
--- a/do.sh
+++ b/do.sh
@@ -1,0 +1,73 @@
+#!/bin/bash
+
+COLUMNS="`tput cols`"
+LINES="`tput lines`"
+BOLD=$(tput bold)
+NORMAL=$(tput sgr0)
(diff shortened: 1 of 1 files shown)


marksweb:
> Why not just extend the existing Makefile? Or better yet, migrate to Just.

Because I've already got the script in another project so it's easy to bring across.

PR #2248, opened 2025-10-10
Remove build dependencies to test libsass

ulgens:
This PR is to trigger the CI to understand the issue with the libsass build. Not intended to be merged.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,9 +29,7 @@ ARG BUILD_DEPENDENCIES="g++ gcc libc6-dev libpq-dev zlib1g-dev"
 # install python dependencies
 COPY ./requirements ./requirements
 RUN apt-get update \
-    && apt-get install --assume-yes --no-install-recommends ${BUILD_DEPENDENCIES} \
     && python3 -m pip install --no-cache-dir -r ${REQ_FILE} \
-    && apt-get purge --assume-yes --auto-remove ${BUILD_DEPENDENCIES} \
     && apt-get distclean
 
 # copy project


ulgens:
@bmispelon @pauloxnet

* Prod image builds fail because the OS-level packages I removed for testing are required for psycopg[c], you can ignore that for now.
* Test image build works fine in CI. My memory about this was that it was failing in CI, but apparently it's not the case.
* Test image build fails locally with a ARM-family processor, saying

```
error: command 'gcc' failed: No such file or directory
...
30.57 ERROR: Failed to build installable wheels for some pyproject.toml based projects (libsass)
```

I think this is the result of Aarch64 wheels being missing and not about Python versions. With the current maintenance status of libsass-python, I don't think this issue will be fixed in a foreseeable future.

libsass is not a blocker for now, and having it doesn't hurt as in a way that matter, but I think this is something we will need to replace soon. Considering libsass itself is deprecated

> Warning: [LibSass is deprecated](https://sass-lang.com/blog/libsass-is-deprecated). While it will continue to receive maintenance releases indefinitely, there are no plans to add additional features or compatibility with any new CSS or Sass features. Projects that still use it should move onto [Dart Sass](https://sass-lang.com/dart-sass).

https://github.com/sass/libsass?tab=readme-ov-file#libsass---sass-compiler-written-in-c

We can start looking for alternatives when we have the time / energy.

---

@dvarrazzo I'm not sure you wanted to be included in this discussion but pinging you for just this once, related to [your comment](https://github.com/django/djangoproject.com/pull/2212#discussion_r2404088471).

dvarrazzo:
@ulgens Sorry for the late reply: I am having bandwidth problems... Answered now 🙂

pauloxnet:
As I already commented, I will explore the removal of SASS and switch to standard CSS3 i the project.

Dockerfile
```diff
@@ -29,9 +29,7 @@ ARG BUILD_DEPENDENCIES="g++ gcc libc6-dev libpq-dev zlib1g-dev"
 # install python dependencies
 COPY ./requirements ./requirements
 RUN apt-get update \
-    && apt-get install --assume-yes --no-install-recommends ${BUILD_DEPENDENCIES} \
     && python3 -m pip install --no-cache-dir -r ${REQ_FILE} \
-    && apt-get purge --assume-yes --auto-remove ${BUILD_DEPENDENCIES} \
     && apt-get distclean
```
pauloxnet, line 33: I know that this is only a draft PR, but when we'll be removing compilation needs for every requirements, we can remove also all `apt` calls and leave only this line:
```docker
RUN python3 -m pip install --no-cache-dir -r ${REQ_FILE}
```

adamzap:
@pauloxnet Let me know if you want some help removing SASS. I want to as well!

pauloxnet:
> @pauloxnet Let me know if you want some help removing SASS. I want to as well!

It would be great. 🤗

I'm not sure if @ulgens is already working on the sass removal, otherwise if you want create a draft PR we can then review it.

ulgens:
@pauloxnet @adamzap I'm not working on it, and won't be able to. I can help with the cleaning of dependencies tho.

I created an issue to follow this idea and implementation: https://github.com/django/djangoproject.com/issues/2251

ulgens:
Now that we have the testing result from CI and https://github.com/django/djangoproject.com/issues/2251 is ready, I'm closing this PR.

PR #2253, opened 2025-10-11
Bump django-hosts to 7.0.0

ulgens:
Supersedes:
- https://github.com/django/djangoproject.com/pull/2105

I tested different versions and commits from the django-hosts repo and found that https://github.com/jazzband/django-hosts/commit/591bf5dd8de7e6046c4ce3a89aec0bf522b70bcf creates a behavioral change that results in some responses we test to switch from FOUND to OK. I'm not exactly sure why and I could some help to understand that.

### DIFF
--- a/djangoproject/tests.py
+++ b/djangoproject/tests.py
@@ -104,7 +104,8 @@ class ExcludeHostsLocaleMiddlewareTests(TestCase):
         "We get no Content-Language or Vary headers when docs host is excluded"
         with self.settings(LOCALE_MIDDLEWARE_EXCLUDED_HOSTS=[self.docs_host]):
             resp = self.client.get("/", headers={"host": self.docs_host})
-        self.assertEqual(resp.status_code, HTTPStatus.FOUND)
+
+        self.assertEqual(resp.status_code, HTTPStatus.OK)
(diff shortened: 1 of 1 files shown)


bmispelon:
> I'm not exactly sure why and I could some help to understand that.

I don't understand why this PR was merged if nobody has been able to provide an explanation as to the behavior change. Did I miss a discussion somewhere?

SaptakS:
Apologies, I overlooked that this PR was looking for further clarification. I reviewed, and status OK made sense to me (since I don't think those URLs are supposed to be redirected), and I saw the [commit linked](https://github.com/jazzband/django-hosts/commit/591bf5dd8de7e6046c4ce3a89aec0bf522b70bcf) by @ulgens and it seems to add support for considering `PARENT_HOST` which was not supported before, which is what I assume caused the change in status code.

@bmispelon I am happy to revert the PR, if you think that makes more sense. Apologies on my part for missing that this PR was waiting on more discussion.

ulgens:
I believe we won't need to revert the PR - it doesn't look like it's breaking or dangerous in any way, but I'd like to know more about what's causing the change on our end.

PR #2254, opened 2025-10-11
Update single quoted dosctrings to triple quotes

ulgens:
This is recommend by PEP257 for both single and multi line docstrings, and detected as a warning by Pycharm's code inspection
- https://peps.python.org/pep-0257/#one-line-docstrings
- https://peps.python.org/pep-0257/#multi-line-docstrings

### DIFF
--- a/releases/models.py
+++ b/releases/models.py
@@ -20,7 +20,7 @@ from .utils import get_loose_version_tuple
 # about older release candidates. Safe to use Django's copy of get_version()
 # when upgrading this website to use Django 1.10.
 def get_version(version=None):
-    "Return a PEP 386-compliant version number from VERSION."
+    """Return a PEP 386-compliant version number from VERSION."""
     version = get_complete_version(version)
 
     # Now build the two parts of the version number:
--- a/djangoproject/tests.py
+++ b/djangoproject/tests.py
@@ -101,7 +101,7 @@ class ExcludeHostsLocaleMiddlewareTests(TestCase):
         DocumentRelease.objects.create(lang="en", release=r2, is_default=True)
 
     def test_docs_host_excluded(self):
-        "We get no Content-Language or Vary headers when docs host is excluded"
+        """We get no Content-Language or Vary headers when docs host is excluded"""
         with self.settings(LOCALE_MIDDLEWARE_EXCLUDED_HOSTS=[self.docs_host]):
             resp = self.client.get("/", headers={"host": self.docs_host})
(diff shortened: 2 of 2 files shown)


releases/models.py
```diff
@@ -20,7 +20,9 @@
 # about older release candidates. Safe to use Django's copy of get_version()
 # when upgrading this website to use Django 1.10.
 def get_version(version=None):
-    "Return a PEP 386-compliant version number from VERSION."
+    """
+    Return a PEP 386-compliant version number from VERSION.
+    """
```
pauloxnet:
```suggestion
    """Return a PEP 386-compliant version number from VERSION."""
```

I agree with this PR but let's follow also other PEP-257 rules for example

> One-liners are for really obvious cases. They should really fit on one line.

https://peps.python.org/pep-0257/#one-line-docstrings

And it's applicable to all changes in this PR.

ulgens: Done.

bmispelon:
I don't have any opposition to this change, but this should really be done automatically via a linter/formatter, not manually.

Does Black not care about PEP257?

ulgens:
> Does Black not care about PEP257?

Yep, it doesn't. I was expecting this to be covered by pycodestyle but it doesn't seem to be the case either. Couldn't find anything related on https://docs.astral.sh/ruff/rules/#pydocstyle-d

SaptakS:
@ulgens I think https://docs.astral.sh/ruff/rules/triple-single-quotes/ is the rule relevant to applying triple double quotes. But I guess it doesn't check for whether it's a single double quote.

SaptakS:
> I don't have any opposition to this change, but this should really be done automatically via a linter/formatter, not manually.

I agree that this should be in linter/formatter. But most linters like black, ruff seem to care whether its a triple single quote or triple double quote. It seems like no linter actually takes into account whether a docstring is just a double quote. So I think this PR still makes sense.

ulgens:
@SaptakS 

> @ulgens I think https://docs.astral.sh/ruff/rules/triple-single-quotes/ is the rule relevant to applying triple double quotes. But I guess it doesn't check for whether it's a single double quote.

You are right. The description says

> Checks for docstrings that use '''triple single quotes''' instead of """triple double quotes""".

And I expected it only converts `'''` to `"""` but it seems it handles `"`  to `"""` too.

We can revisit this when we are ready to revamp the linting setup.

---
I created an issue on ruff repo to report the misleading description on D300: https://github.com/astral-sh/ruff/issues/20838

Issue #1499, opened 2024-03-26
Site-wide search


thibaudcolas:
> Part of recommendations from user research report in #1496

Currently search is only available for documentation. Integrate this site-wide / for all site content.

Relevant parts of the report:

> Improving internal search tools could further simplify the user experience, enabling users to quickly find what they need without having to resort to external search engines.

And:

> **Improving internal search tools** could further simplify the user experience, enabling users to quickly find what they need without having to resort to external search engines.

brainwane:
I'd personally appreciate this - I just looked on djangoproject.com and was trying to look around for some materials related to volunteering for the Foundation, and was having a hard time. Sitewide search would be really helpful.

thibaudcolas:
Discussed at today’s WG meeting:

- We want to proceed with a combined docs + "other content sources" search
- BE implementation is reasonably close to accommodating this already
- Needed sources: static Django templates; flatpages; meeting notes; blog posts; etc
- UX task: decide on the search experience with mixed results across different content types (faceted search?)

jacklinke:
Here is the list of views with associated templates and/or models (with the potentially-useful model fields) to consider for indexing.

The listed templates are those containing significant text content, or which serve as  endpoints with content loaded from models - I did not include base templates and snippets.

 I am sure we can whittle this list down, but I hope it works as a starting point.

## aggregator/

- LocalDjangoCommunitiesListView: views.py
    - https://github.com/django/djangoproject.com/blob/main/djangoproject/templates/aggregator/local-django-community.html
    - LocalDjangoCommunity: models.py
        - name
        - description
        - slug
        - city
        - country
        - continent
        - website_url
        - event_site_url

## blog/

- BlogArchiveIndexView, BlogYearArchiveView, BlogMonthArchiveView, BlogDayArchiveView, BlogDateDetailView: views.py
    - https://github.com/django/djangoproject.com/blob/main/djangoproject/templates/blog/entry_archive.html
    - https://github.com/django/djangoproject.com/blob/main/djangoproject/templates/blog/entry_archive_day.html
    - https://github.com/django/djangoproject.com/blob/main/djangoproject/templates/blog/entry_archive_month.html
    - https://github.com/django/djangoproject.com/blob/main/djangoproject/templates/blog/entry_archive_year.html
    - https://github.com/django/djangoproject.com/blob/main/djangoproject/templates/blog/entry_detail.html
    - Entry: models.py
        - headline
        - slug
        - summary
        - summary_html
[...]

PR #2262, opened 2025-10-16
Moved docs search bar to top navigation in prep for global search.

sarahboyce:
Relates to https://github.com/django/djangoproject.com/issues/1499

This moves the "docs" searchbar to the header of the website so that searching (which currently **only searches the docs, the ecosystem page, and some blog posts**) is available everywhere **from a UI perspective**.

The next piece of work is extending the search for the website. But we can review whether these design changes are acceptable before we get into the weeds of this :grin: 

## Notable consequence

As the search requires information of the release to be searched, this now means there is an extra query required to render the header (and so 1 query added to a page load in most cases, the docs already had this).
This also means that in order to test a view, you should have a default `DocumentRelease` object created. I have added this into the existing view tests with a `ReleaseMixin`

## UPDATE

Note that the images and video are slightly out of date as I have made it that the search results doesn't say the release version being searched (as we no longer only search the docs) and "Documentation" is no longer highlighted in the header. Otherwise, the search looks similar :+1: 

## Mobile 412x915

Screenshot from 2025-10-16 20-41-07

## Tablet 810x1080

Screenshot from 2025-10-16 20-43-03

## Laptop 1492x1080

Screenshot from 2025-10-16 20-44-17


docs/templatetags/docs.py
```diff
@@ -17,17 +17,21 @@
 register = template.Library()
 
 
-@register.inclusion_tag("docs/search_form.html", takes_context=True)
+@register.inclusion_tag("search_form.html", takes_context=True)
 def search_form(context):
     request = context["request"]
```
marksweb, line 35: This is failing the tests, which looking at it just here, I find surprising. If we're taking the context, why isn't the request there 🤔

sarahboyce: The error was a bit misleading and a consequence of how error pages are loaded without context. I added a fix to make the errors more useful with a comment. Also updated the existing tests :+1:

docs/templatetags/docs.py
```diff
@@ -17,18 +17,37 @@
...
-    release = DocumentRelease.objects.get_by_version_and_lang(
-        context["version"],
-        context["lang"],
-    )
-    return {
+    lang = context.get("lang", "en")
+
+    render_context = getattr(context, "render_context", None)
+    if render_context is not None and "search_form_data" in render_context:
+        return render_context["search_form_data"]
```
bmispelon: Interesting, I hadn't seen this pattern before. I guess this avoids doing the db queries when `{% search_form %}` is called multiple times in a template?

If so:

1. A test for that would be nice (I guess an `assertNumQueries(1)` block that renders a template with two `{% search_form %}` would work).
2. Could we abstract out this `render_context` caching logic into a generic decorator? Something like `@cached_inclusion_tag("search_form.html")` (we can assume `takes=context=True`)

docs/templatetags/docs.py
```diff
@@ -17,18 +17,37 @@
...
+        # variable is not present in the template context.
+        return {"form": DocSearchForm(release=None), "version": "dev", "lang": "em"}
+
     request = context["request"]
-    release = DocumentRelease.objects.get_by_version_and_lang(
-        context["version"],
-        context["lang"],
-    )
-    return {
+    lang = context.get("lang", "en")
```
bmispelon: I know you didn't start this, so feel free to push back if that's too much faff, but we're starting to have many places that assume English is the default language. Could we refactor that assumption into a constant somewhere? (if it helps, I used `git grep -w en '*.py'` to get a not-too-noisy list of places that make this assumption).

docs/templatetags/docs.py
```diff
@@ -17,18 +17,37 @@
 register = template.Library()
 
 
-@register.inclusion_tag("docs/search_form.html", takes_context=True)
+@register.inclusion_tag("search_form.html", takes_context=True)
 def search_form(context):
+    if "request" not in context:
```
bmispelon, line 25: This `if` branch needs a unit test too.

docs/templatetags/docs.py
```diff
@@ -17,18 +17,37 @@
...
+        # Django's built-in error views (like django.views.defaults.server_error)
+        # render templates without attaching a RequestContext — meaning the 'request'
+        # variable is not present in the template context.
+        return {"form": DocSearchForm(release=None), "version": "dev", "lang": "em"}
```
bmispelon: I don't believe `em` is a valid language code 😁  (see my other comments about tests and having a `DEFAULT_LANGUAGE` constant)

djangoproject/templates/includes/header.html
```diff
@@ -6,7 +7,10 @@
     <a class="logo" href="{% url 'homepage' %}">Django</a>
     <p class="meta">The web framework for perfectionists with deadlines.</p>
     <div class="mobile-toggle">
-      {% include "includes/toggle_theme.html" %}
+      {% search_form %}
+      <div class="dark-mode">
```
bmispelon: I'm a bit confused with the naming of this `class`. Does `dark-mode` here mean that this div contains the light/dark/system toggler?

djangoproject/tests.py
```diff
@@ -10,6 +10,13 @@
 from docs.models import DocumentRelease, Release
 
 
+class ReleaseMixin:
+    @classmethod
+    def setUpTestData(cls):
+        r2 = Release.objects.create(version="2.0")
+        DocumentRelease.objects.create(lang="en", release=r2, is_default=True)
```
bmispelon: I wonder if this could create problems if some `TestCase` using this mixin were to create another `is_default` release before calling `super().setUpTestData()`.

Would that create problems? Could we get around it using `get_or_create()` somehow?

docs/forms.py
```diff
@@ -1,7 +1,7 @@
 from django import forms
 from django.utils.translation import gettext_lazy as _
 
-search_placeholder = _("Search %s documentation")
+search_placeholder = _("Search")
```
bmispelon: Since English loves having words that can be either nouns or verbs, I suggest adding a [context marker](https://docs.djangoproject.com/en/5.2/topics/i18n/translation/#contextual-markers) here to help future translators:
```suggestion
search_placeholder = pgettext("Search", "Action to search the whole site")
```

docs/forms.py
```diff
@@ -13,6 +13,6 @@ def __init__(self, data=None, **kwargs):
         self.fields["q"].widget = forms.TextInput(
             attrs={
                 "type": "search",
-                "placeholder": search_placeholder % self.release.human_version,
+                "placeholder": search_placeholder,
```
bmispelon: Not sure why the `search_placeholder` variable is needed. It doesn't seem to be reused anywhere and could just as well be inlined here I think.

djangoproject/templates/includes/header.html
sarahboyce: Yes, I've renamed to something hopefully clearer

sarahboyce:
Note that I plan to review the scss over the weekend and add some comments to explain what things are :+1: 
I believe I have resolved the Python code comments

djangoproject/static/img/logo-dj.svg
sarahboyce, line 1: This is used in the header for mobile screens to give enough room for the searchbar

djangoproject/scss/_dark-mode.scss
```diff
@@ -346,20 +346,28 @@ html[data-theme="light"] .theme-toggle .theme-label-when-light {
...
+// Section of the header that is only visible on mobile.
+.header-mobile-only {
     line-height: 45px;
     text-align: center;
-    margin: 4px 10px;
+    display: flex;
+    align-items: center;
 
-    @include respond-min(768px) {
+    @include respond-min(1080px) {
```
sarahboyce, line 357: I have updated the point where the header would like the list items into a menu from 768px to 1080px and have used this consistently in header related styling. 1080px is not special, this was roughly what felt like the right point when testing.

Here is a video to show how the header is changing as the screen gets smaller:

https://github.com/user-attachments/assets/dbf2ba1e-68f1-41a7-be3d-e891331128f2

sarahboyce (commented):
I have added some commentary to the scss updates.
I have roughly tried to acheive something that looks reasonable on multiple screen sizes
Most updates are around having vertical alignment of header items and tweaking margin

I am not a css expert. I tried to remove things I deemed as unneeded (from testing) . Happy for any comments around better ways to achieve a "good enough" header

djangoproject/scss/_dark-mode.scss
SaptakS: I think this makes sense to me with the search field added to the header.

SaptakS:
@bmispelon I see you had requested some changes which seem to have been addressed already. Did you want to review this again? Else I think this should be ready for merging.

bmispelon:
Amazing work everyone! 💚 💪🏻

Issue #2264, opened 2025-10-16
Search says there's 1 result, but no results are shown


bmispelon:
When searching for "ananas" on the french 5.2 documentation ([link](https://docs.djangoproject.com/fr/5.2/search/?q=ananas)), the title of the page says "1 result found" (in French), but no results are actually listed:

<img width="867" height="811" alt="Image" src="https://github.com/user-attachments/assets/d04af56a-f9a5-4734-a6fd-b75edf932582" />

This seems like a bug: the result count in the title should match the actual list of results displayed.

PR #2270, opened 2025-10-18
Fix search results count mismatch bug

rrrr08:
## Problem
Search results occasionally showed a count header (e.g., "1 result") but displayed an empty list due to a race condition in Django ORM lazy evaluation.

## Root Cause
Race condition in Django ORM lazy evaluation with paginator:
- Paginator's `paginator.count` runs `SELECT COUNT(*)` at time T1
- Paginator's `page.object_list` runs `SELECT * LIMIT` at time T2
- Database state could change between T1 and T2, causing mismatched results

## Solution
- **Backend**: Force queryset evaluation with `results = list(results)` **before** passing to `Paginator`, ensuring both `paginator.count` and `page.object_list` see the same evaluated data
- **Template**: Add `{% if paginator.count > 0 %}` check to show "No results" message when count is 0

## Files Changed
- `docs/views.py`
- `docs/templates/docs/search_results.html`

### DIFF
--- a/docs/templates/docs/search_results.html
+++ b/docs/templates/docs/search_results.html
@@ -18,21 +18,35 @@
         <a{% if active_category == category %} aria-current="page"{% else %} href="{% querystring category=category.value page=None %}"{% endif %}>{{ category.label }}</a>
       {% endfor %}
     </search>
-    <h2>
-      {% if release.is_dev %}
-        {% blocktranslate count num_results=paginator.count trimmed %}
-          Only 1 result for <em>{{ query }}</em> in the development version
-        {% plural %}
-          {{ num_results }} results for <em>{{ query }}</em> in the development version
-        {% endblocktranslate %}
-      {% else %}
-        {% blocktranslate count num_results=paginator.count trimmed %}
-          Only 1 result for <em>{{ query }}</em> in version {{ version }}
-        {% plural %}
-          {{ num_results }} results for <em>{{ query }}</em> in version {{ version }}
-        {% endblocktranslate %}
-      {% endif %}
-    </h2>
+    {% if paginator.count > 0 %}
+      <h2>
+        {% if release.is_dev %}
+          {% blocktranslate count num_results=paginator.count trimmed %}
+            Only 1 result for <em>{{ query }}</em> in the development version
+          {% plural %}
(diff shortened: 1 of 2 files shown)


ulgens:
Could you share a case - instructions to replicate the issue? I'm also confused about the template changes and why did we have a special block for "1 result".

rrrr08:
Replication Steps:

Start the Django server: docker compose up
Navigate to search: http://localhost:8000/en/5.1/search/?q=test
Apply a category filter (e.g., "Guides")
Search for obscure/rare terms where results might be empty
Observe the bug: Header shows "N results for <query>" but the list is empty . 

// This issue was mainly in "fr" language for words like "ananas" etc

This was the issue and was due to the difference between the value of paginator.count and page.object_list. Actually they queried the database at two different instance of time which lead to some inconsistency when the data in the databased changed during that interval.

So to fix it I made changes so that result is calculated once from the database(the result line which was added) and that is used by both the paginator and page.object_list so that there is  no mismatch. Also i added the if condition in template file to handle the 0 item case separately. Its not absolutely necessary but visually "NO items found" is better than "0 items found". Also the special block for the "1 result" is purely for grammar like "1 result found" and "2 results found". This was already in the repository and is not changed or added by me. I just added a precautionary if condition to handle the "0 item found" case better and make it more grammatically correct.

Mainly I fixed the issue in view.py by making a single database query for the paginator and page.object_list so that they work on the same snapshot of the database.

ulgens:
Thank you for the details. I'll check the condition and the change in the Python file, but I think the template changes deserver its own PR.

rrrr08:
Sir I have added a separate PR only for the python changes as you asked. https://github.com/django/djangoproject.com/pull/2274.

sarahboyce:
> This was the issue and was due to the difference between the value of paginator.count and page.object_list. Actually they queried the database at two different instance of time which lead to some inconsistency when the data in the databased changed during that interval.

Given that this is behaving consistently (see https://docs.djangoproject.com/fr/5.2/search/?q=ananas), this doesn't really hold true

sarahboyce:
Closing in favor of https://github.com/django/djangoproject.com/pull/2274

Issue #2263, opened 2025-10-16
Can't change language from the search result page


bmispelon:
When you [make a search](https://docs.djangoproject.com/en/5.2/search/?q=ananas), the language switcher is displayed on the bottom right (at least on desktop), but hovering or clicking on it does nothing.

Instead, the workaround is to either:
1. Manually change the language in the URL;
2. Or navigate away from search to the root of the documentation, change the language there, then do the search again.

It should be possible to change the language directly from the search result page, or if not then the language switcher should not be displayed.

PR #2272, opened 2025-10-18
Fix language switcher on search results pages

rrrr08:
### Problem
The language switcher appeared on search results pages but had no clickable language options, preventing users from switching languages after performing a search.

### Root Cause
The `search_results` view was missing the `available_languages` context variable that the template requires to render language links.

### Solution
Added `available_languages` to the view context by calling `get_available_languages_by_version()`, matching the implementation in the `document()` view.

Fixes https://github.com/django/djangoproject.com/issues/2263


Makefile
sarahboyce, line 1: Please revert deleting the Makefile

rrrr08: Done👍. You can review it.

bmispelon:
I've tested this locally and it works. Thanks!

Issue #2266, opened 2025-10-17
Make the search result page filter tabs responsive on mobile


sarahboyce:
<img width="494" height="530" alt="Image" src="https://github.com/user-attachments/assets/f1c18d0b-1e08-46d0-8d56-63199748c3aa" />

The search result page filter tabs require a horizontal scroll on small screens. I think having the filters be a dropdown on smaller screens would be better

keranbyge:
Hi @sarahboyce 

I’d like to work on this issue if it’s available!  
I have experience working with front-end design and responsiveness using HTML, CSS, and Tailwind. I’ve already contributed a couple of PRs recently to the Django and djangoproject.com repos, so my local setup is ready.  

From what I understand, this issue involves making the search result filter tabs mobile-responsive — possibly by turning them into a dropdown or a collapsible layout on smaller screens.  

Please let me know if I can take this up — I’ll make sure to follow the project’s style guidelines and test across screen sizes before submitting the PR.

sarahboyce:
This is already being worked on: https://github.com/django/djangoproject.com/pull/2273

There's been engagement on that PR relatively recently. I think it would be available for someone else to pick up if there's no further engagement in a month's time

keranbyge:
Got it, thank you @sarahboyce! 

I’ll keep an eye on #2273, and if there’s no further progress after some time, I’ll be happy to take it over then.  
Meanwhile, I’ll continue contributing to other open issues in the project.

PR #2273, opened 2025-10-18
Fix horizontal scroll on mobile search filters

rrrr08:
Problem
Filter tabs on the search results page require horizontal scrolling on mobile devices, creating a poor user experience.

Solution
Implement responsive design with:

Desktop (>768px): Horizontal tabs (unchanged)
Mobile (≤768px): Dropdown menu
Both UIs exist in HTML; CSS media queries control visibility.

Changes
docs/templates/docs/search_results.html: Add filter-tabs div and filter-dropdown select
djangoproject/scss/_style.scss: Add responsive styles with @media queries

Benefits
No horizontal scrolling on mobile
Touch-friendly dropdown interface
Maintains accessibility (ARIA labels, keyboard navigation)
Desktop experience unchanged

Fixes https://github.com/django/djangoproject.com/issues/2266

### DIFF
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -2667,7 +2667,7 @@ search.filters {
     // Tablet/Mobile: Show hybrid (All + selected + More dropdown)
     .filter-tabs-responsive {
       display: none; // Hide on desktop
-      
+
       @media (max-width: 1024px) {
         display: flex;
         gap: 10px;
@@ -2714,7 +2714,7 @@ search.filters {
 
       &[aria-expanded="true"] {
         color: var(--body-fg);
-        
+
         .filter-more-icon {
           transform: rotate(180deg);
         }
@@ -2789,7 +2789,7 @@ search.filters {
         box-shadow: none !important;
 
         // Remove underline from all child elements
-        *, 
+        *,
         .filter-checkmark {
           text-decoration: none !important;
           text-decoration-line: none !important;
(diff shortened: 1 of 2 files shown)


sarahboyce:
Thank you for the updates! This is for https://github.com/django/djangoproject.com/issues/2266

For others to see, this is what it looks like on mobile:
Screenshot from 2025-10-18 15-58-49

I think:
- the bottom border shouldn't exist for the dropdown
- we might want some top margin/padding
- possibly, we want a filter icon :thinking: 

Also, we still require a horizontal scroll for tablet screen/small laptop screens
Screenshot from 2025-10-18 16-01-29

So we might want this to snap a little earlier.

A different option could be that we keep the tabs, but on mobile or tablet we have it snap into something like:
All | Option 1 | More (dropdown)
Where Option 1 would be whatever is selected or the first option (if no filter is applied)

rrrr08:
I have improved the UI and the mobile/tablet dropdown logic as you suggested. Please review the changes once.

sarahboyce:
I think it will help the review if you include screenshots

This is what I see:

Screenshot from 2025-10-19 14-39-46

This is closer to what I recently suggested thank you
I think the main issue here is although "API Reference" is selected, it is not clear visually that this is selected. When I made the suggestion, I expected "All" to always be available (which you have done) but also 1 other tab (in this case it would be "API Reference" and clearly selected) and then the dropdown always says "More"
I also personally don't like having the burger menu when it is a dropdown

adamzap:
@rrrr08 Do you plan to continue working on this pull request?

adamzap:
@rrrr08 I'm going to close this for now. Please reopen it if you plan to complete it. Thanks!

PR #2274, opened 2025-10-18
Fix #2264 :  Fix search results count race condition

### LINKED ISSUE #2264: Search says there's 1 result, but no results are shown

rrrr08:
## Problem
Search results page shows incorrect count when paginator.count and
page.object_list query the database at different times, causing a
race condition. Header may display 'N results' while showing an
empty list.

## Solution
Convert QuerySet to list immediately after search to force single
database query. Both paginator.count and page.object_list now use
the same data snapshot from memory.

## Changes
- docs/views.py: Add results = list(results) after search query

### DIFF
--- a/docs/views.py
+++ b/docs/views.py
@@ -168,6 +168,11 @@ def search_results(request, lang, version, per_page=10, orphans=3):
                 q, release, document_category=doc_category
             )
 
+            # Force queryset evaluation to prevent race conditions between
+            # paginator.count and page.object_list accessing the database
+            # at different times with potentially different transaction states
+            results = list(results)
+
             page_number = request.GET.get("page") or 1
             paginator = Paginator(results, per_page=per_page, orphans=orphans)
 


sarahboyce:
Note that I don't follow that this is the solution here (see https://github.com/django/djangoproject.com/pull/2270#issuecomment-3425071896)

Lots of hand waving here but... I'm currently wondering if the mismatch could be related to some PostgreSQL full-text search quirks and how stored GeneratedFields affect query planning. 
In `.search()`, `if not vector_qs:` might trigger some partial queryset evaluation and cache elements of the query. So a possible fix is to update `.search()` to use `if not vector_qs.exists():` to force a fresh query and maybe help make pagination more consistent 🤔 

But I'm not sure and it's not easy to write a regression test for. Also locally I only have English docs to test with and I haven't found a search term within the English docs that replicates this issue 😓

@bmispelon I see a preview environment has been setup. Does this have the full french docs and is the issue reproducable there?

bmispelon:
> @bmispelon I see a preview environment has been setup. Does this have the full french docs and is the issue reproducable there?

It does, and it is 😁  https://docs.preview.djangoproject.com/fr/5.2/search/?q=ananas

bmispelon:
Hi and thanks for your attempt at solving this issue!

Can you clarify whether this fix is working for you locally? I just tried it using the steps outline in the bug report and it's not any better: it's still saying "1 result" but listing none.

As Sarah pointed out above, this is unlikely to be a race condition since it happens consistently and without parallel access to the database.

I'm going to close this but please let me know if I've missed something and we can reopen.

PR #2276, opened 2025-10-19
Added static views to sitemap.

sarahboyce:
This relates to https://github.com/django/djangoproject.com/issues/1499

This is motivated by the following PR: https://github.com/django/djangoproject.com/pull/2284

The idea is that if we want to add more to search, we extend or improve the sitemap rather than needing to add more logic to the management command

Comments on the approach are welcome. This PR likely still has some value even if we do not follow this approach for search

### DIFF
--- a/djangoproject/sitemaps.py
+++ b/djangoproject/sitemaps.py
@@ -1,0 +1,56 @@
+from dataclasses import dataclass
+
+from django.contrib import sitemaps
+from django_hosts.resolvers import reverse
+
+
+@dataclass
+class URLObject:
+    name: str
+    host: str = "www"
+
+
+class StaticViewSitemap(sitemaps.Sitemap):
+    priority = 0.5
+    changefreq = "monthly"
+
+    def items(self):
+        return [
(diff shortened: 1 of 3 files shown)


djangoproject/sitemaps.py
```diff
@@ -0,0 +1,56 @@
...
+    priority = 0.5
+    changefreq = "monthly"
+
+    def items(self):
+        return [
+            # accounts
+            URLObject("registration_register"),
+            # aggregator
+            URLObject("community-index"),
+            URLObject("community-ecosystem"),
```
sarahboyce, line 23: May need to remove this as it would be included in the docs sitemap after the change in https://github.com/django/djangoproject.com/pull/2136

sarahboyce: I have concluded that website Document objects should be excluded from the docs sitemap and so this can be included here

sarahboyce:
Moved commit to https://github.com/django/djangoproject.com/pull/2284 :+1:

PR #2284, opened 2025-10-21
Synced webpages from sitemap to Document model for search.

sarahboyce:
Fixes https://github.com/django/djangoproject.com/issues/1499

This PR enables the whole website to be searched.
If anything is missing from the search, it is missing from a sitemap. Going forward we should make sure pages we want to be searched are included in our sitemaps.

Screenshot from 2025-11-22 08-32-44

For testing, recommend running `python manage.py update_docs --force --language=en dev`
On deployment we should run the first one with `--force`. We may want to run `python manage.py update_docs --force --language=en dev` (ie. english dev with force) on production every month to update existing searchable website content. Existing website content (static pages, old blog posts) are mostly static so we don't need to run with force very often


docs/models.py
```diff
@@ -206,16 +215,66 @@ def sync_to_db(self, decoded_documents):
...
+    def sync_from_sitemap(self, force=False):
+        from djangoproject.urls.www import sitemaps
+
+        if not self.is_dev:
+            return
+
+        if force:
+            Document.objects.filter(
+                metadata__parents=DocumentationCategory.WEBSITE
+            ).delete()
```
sarahboyce, line 239: In future we could add a `last_modified` field to Document and check the sitemap last modified and make updates according to this.
But for now, I think having a cron job to build the English dev docs with force (`update_docs --force --language=en dev`) running every 2 weeks (or even every month) will be good enough to update changes to the webiste pages for search and remove any stale pages. This content doesn't change very often, nor do we remove content often

g7aaa (approved):

sarahboyce:
@django/django-website if anyone has time, I would love a review for this PR 💚

blog/models.py
```diff
@@ -37,9 +37,6 @@ def published(self):
     def active(self):
         return self.filter(is_active=True)
 
-    def searchable(self):
```
alexgmin, line 40: Since `is_searchable` is not used anymore it should be removed.

djangoproject/urls/www.py
```diff
@@ -21,6 +22,7 @@
 sitemaps = {
     "weblog": WeblogSitemap,
     "flatpages": FlatPageSitemap,
+    "static": StaticViewSitemap,
```
alexgmin:
```suggestion
    "templates": "TemplateViewSitemap"`
```

`static` can be confused with static assets, so maybe calling it templates makes more sense.

docs/models.py
```diff
@@ -216,81 +217,63 @@ def sync_to_db(self, decoded_documents):
...
-                Document(
+        for sitemap in sitemaps.values():
+            for url in sitemap().get_urls():
+                path = url["location"]
+                if path in doc_urls:
+                    continue
+                try:
+                    page_html = fetch_html(path)
+                except requests.RequestException:
+                    continue
```
alexgmin, line 256: We should log the failures. 
This was failing in my docker setup with no indication of why, for example.

docs/models.py
```diff
@@ -351,6 +334,14 @@ def search(self, query_text, release, document_category=None):
...
             )
             base_filter = Q(release_id=release.id)
+            if release.lang == settings.DEFAULT_LANGUAGE_CODE and not release.is_dev:
+                dev_release = DocumentRelease.objects.get_by_version_and_lang(
+                    version="dev", lang=settings.DEFAULT_LANGUAGE_CODE
+                )
+                base_filter |= Q(
+                    release_id=dev_release.id,
+                    metadata__parents=DocumentationCategory.WEBSITE,
+                )
```
alexgmin, line 353: This probably needs a comment. It took a lot of thinking to find why this is `if` is needed.
Mentioning why dev documents are being added to for example 5.2 searches.

docs/models.py
```diff
@@ -351,6 +334,14 @@ def search(self, query_text, release, document_category=None):
                 config=models.F("config"),
             )
             base_filter = Q(release_id=release.id)
+            if release.lang == settings.DEFAULT_LANGUAGE_CODE and not release.is_dev:
+                dev_release = DocumentRelease.objects.get_by_version_and_lang(
+                    version="dev", lang=settings.DEFAULT_LANGUAGE_CODE
+                )
+                base_filter |= Q(
+                    release_id=dev_release.id,
```
alexgmin, line 351: I was going to suggest this way: `release__version="dev", release__lang=settings.DEFAULT_LANGUAGE_CODE` vs retrieving the `dev_release`.

However doing it this way is probably more optimal, so we should add a comment since an extra query may be interpreted as less optimal initially.

docs/search.py
```diff
@@ -81,25 +78,31 @@ def parse(cls, value, default=None):
...
+        "User-Agent": (
+            "Mozilla/5.0 (Windows NT 10.0; Win64; x64) "
+            "AppleWebKit/537.36 (KHTML, like Gecko) "
+            "Chrome/123.0.0.0 Safari/537.36 Edg/123.0.0.0"
+        ),
+        "Accept": "text/html",
+        "Accept-Language": "en-US,en;q=0.9",
+    }
+
+    response = requests.get(url, headers=headers, timeout=timeout)
```
alexgmin: Instead of doing external requests, you could use the Django test Client, like this. Based on my basic tests, it should work.

```
from django.test import Client
from urllib.parse import urlparse

client = Client()

parsed = urlparse(url)
response = client.get(parsed.path, headers={'HOST': parsed.netloc})  #Using netloc to include the port

return response.text
```

Doing it this way makes it work in a local docker setup (Django inside the container listens on 0.0.0.0, not localhost), and it's faster.
It also means we don't need all of the mocking in the tests

alexgmin (commented):
This was a pair review done by the French @bmispelon and the Spanish me.

docs/search.py
sarahboyce: I have updated to use the test client and removed the mocking.

The tests that run `sync_from_sitemap()` now have some new error logs. This is because some template views included in the sitemap are no longer mocked and we are missing the data to render these pages correctly (specifically we need the release pages represented in the `Document` model).
Options are: ignore these logs, create the `Document` instances (a little challenging), or mock the responses.

Also I think we should check in the preview environment if the move to the test client works :+1:

bmispelon:
Progress update: I've deployed this to the preview server, then ran `python -m django update_docs --force en/dev` but it failed while trying to index the download page:
```
django.urls.exceptions.NoReverseMatch: Reverse for 'document-detail' with keyword arguments '{'lang': 'en', 'version': '', 'url': 'releases/'}' not found. 1 pattern(s) tried: ['(?P<lang>[a-z-]+)/(?P<version>[\\w.-]+)/(?P<url>[\\w./-]*)/$']
```

I believe this is because there's no `previous` version available on the preview server, which is causing a crash in the `release_notes` template tag. I'll try to populate the database so the page doesn't crash anymore, then I'll attempt reindexing again.

Thanks for bearing with me! 🐻

docs/models.py
```diff
@@ -216,81 +219,64 @@ def sync_to_db(self, decoded_documents):
...
-                Document(
+        for sitemap in sitemaps.values():
+            for url in sitemap().get_urls():
+                path = url["location"]
+                if path in doc_urls:
+                    continue
+                try:
+                    page_html = fetch_html(path)
+                except Exception as e:
+                    logger.error(str(e))
```
alexgmin:
```suggestion
                except ValueError:
                    logger.exception("Error indexing template view for search")
```
Two changes here:
1. The only exception that `fetch_html` can return is `ValueError` when the page is not found, so only that one needs to be captured.
2. `logger.exception` prints the exception nicer and with a traceback, which may be useful for Sentry.

sarahboyce:
Thank you both for the reviews and testing :star: 
I know this isn't perfect and we might rip it out later on, but it adds some form of website wide search :slightly_smiling_face: 

@bmispelon if you can manage the merge, deploy, management command run - that would be great :star: 
Otherwise I should be able to do this in April :grin:

PR #2285, opened 2025-10-22
Fixed translations of results count for languages where 0 is singular rather than plural.

sarahboyce:

Got suspicious when every French 0 term had this issue:
- https://docs.djangoproject.com/fr/5.2/search/?q=rubbish
- https://docs.djangoproject.com/fr/5.2/search/?q=potato

Turns out that 0 things is singular, not plural in French and so 0 results would translate to "Only 1 result" as 0s need the singular translation.

Happy to tweak this however thought best

### DIFF
--- a/docs/templates/docs/search_results.html
+++ b/docs/templates/docs/search_results.html
@@ -21,13 +21,13 @@
     <h2>
       {% if release.is_dev %}
         {% blocktranslate count num_results=paginator.count trimmed %}
-          Only 1 result for <em>{{ query }}</em> in the development version
+          {{ num_results }} result for <em>{{ query }}</em> in the development version
         {% plural %}
           {{ num_results }} results for <em>{{ query }}</em> in the development version
         {% endblocktranslate %}
       {% else %}
         {% blocktranslate count num_results=paginator.count trimmed %}
-          Only 1 result for <em>{{ query }}</em> in version {{ version }}
+          {{ num_results }} result for <em>{{ query }}</em> in version {{ version }}
         {% plural %}
           {{ num_results }} results for <em>{{ query }}</em> in version {{ version }}
         {% endblocktranslate %}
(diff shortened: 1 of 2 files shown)


docs/templates/docs/search_results.html
```diff
@@ -20,17 +20,29 @@
...
-        {% plural %}
-          {{ num_results }} results for <em>{{ query }}</em> in the development version
-        {% endblocktranslate %}
+        {% if paginator.count == 0 %}
+          {% blocktranslate trimmed %}
+            No results for <em>{{ query }}</em> in the development version
+          {% endblocktranslate %}
+        {% else %}
+          {% blocktranslate count num_results=paginator.count trimmed %}
+            Only 1 result for <em>{{ query }}</em> in the development version
```
bmispelon: You should still be using `{{ num_result }}` in the singular form, in case we add support for a language where the singular form is used outside of 0 and 1 (you might also want to consider dropping the "only" for that same reason, but I'm not sure if that's as much of an issue):
```suggestion
            Only {{ num_results }} result for <em>{{ query }}</em> in the development version
```

sarahboyce: Good point, let's drop "Only" it causes problems :+1:

bmispelon (approved):
This looks good, nice work! 🌌

I'll make sure it works locally later today, then merge it.

bmispelon:
Funnily enough the bug is present in [Django's docs](https://docs.djangoproject.com/en/5.2/topics/i18n/translation/#blocktranslate-template-tag):
```
{% blocktranslate count counter=list|length %}
There is only one {{ name }} object.
{% plural %}
There are {{ counter }} {{ name }} objects.
{% endblocktranslate %}
```

sarahboyce:
Yeah I was reading the docs trying to figure out if we'd done something wrong or if there was a wrong translation in transifex :sweat_smile: 
It's probably a good idea for us to update the example, as this pattern doesn't always work well (or add an admonition or something)

sarahboyce:
Created a ticket https://code.djangoproject.com/ticket/36681 :+1:

PR #2291, opened 2025-10-25
Populate description meta tag in base_weblog template (fix #2250)

### LINKED ISSUE #2250: Set description meta in blog posts
Our blog posts currently get generated with an empty `description` meta tag. [DSF Board nominations](https://www.djangoproject.com/weblog/2025/oct/11/2026-dsf-board-nominations/) as an example:

```html
<meta name="description" content="" />
```

It seems we need to populate the `description` block:

https://github.com/django/djangoproject.com/blob/8c1c01d5d4d9e81708c58744fe9da86cf3b1c339/djangoproject/templates/base.html#L11

See how `og:description` is set for the expected value for `description`:

https://github.com/django/djangoproject.com/blob/8c1c01d5d4d9e81708c58744fe9da86cf3b1c339/blog/models.py#L211-L213

---

We do need both tags as some crawlers only support "description" (Mastodon for example)

if-i:
Added a description block to base_weblog.html so that the meta description tag is populated for blog pages. This aligns the description with the page title, ensuring better previews on social platforms and search engines.

### DIFF
--- a/djangoproject/templates/base_weblog.html
+++ b/djangoproject/templates/base_weblog.html
@@ -1,6 +1,7 @@
 {% extends "base.html" %}
 {% load fundraising_extras i18n weblog %}
 {% block layout_class %}sidebar-right{% endblock %}
+{% block description %}{% translate "News & Events" %}{% endblock %}
 {% block title %}{% translate "News &amp; Events" %}{% endblock %}
 
 {% block og_title %}{% translate "News &amp; Events" %}{% endblock %}


sarahboyce:
Thank you for the PR :+1: 
I feel what was required here was slightly misinterpreted. The issue mentioned using the `og:description` value and was referring to blog entry detail pages, rather than the blog base template. 
Also going forward, I recommend writing a test for contributions :heavy_check_mark:  

I created a PR for this issue https://github.com/django/djangoproject.com/pull/2292 and will be closing this PR

Issue #2130, opened 2025-07-19
New base template for one-off flat pages


thibaudcolas:
It’d be very useful to have a new "lite" base template for flat pages that are very different from the rest of the site. 

We’re using flat pages for the [2024 Annual Impact Report](https://www.djangoproject.com/foundation/reports/2024/), which uses different stylesheets from anywhere else on the site as the report’s styles are very different. Repo: [dsf-annual-report](https://github.com/thibaudcolas/dsf-annual-report).

---

Longer-term this report might be moved to be part of the main site, but we’ll still have an (occasional) need for one-off documents that are styled differently and still need to be HTML

anya-xcode:
@thibaudcolas I want to work on this issue. Could you please assign it to me?

thibaudcolas:
@anya-xcode we don’t assign issues, everyone can just submit a PR. Providing more details for people who want to work on this at PyConFR - this task involves:

1. (Initial website project setup)
2. Making a new flatpage in the admin of your local project so you can try out how it works
3. Creating a new "empty.html" template for flat pages, in the same folder as https://github.com/django/djangoproject.com/blob/main/djangoproject/templates/flatpages/default.html
4. That template should only contain `{{ flatpage.content }}` so nothing else but the flatpage content is displayed
5. Test it in the Django admin
6. If it all works as expected, submit a PR

sarahboyce:
Completed by https://github.com/django/djangoproject.com/pull/2300

PR #2300, opened 2025-10-30
feat: Add an empty template for flatpage

antoinehashbang:
Related to this issue:  https://github.com/django/djangoproject.com/issues/2130

- Add a template "empty.html" who render only the content of a flatpage, without extending base.html


sabderemane:
> Tested and works as I would expect @thibaudcolas this means we ignore the title added in the admin but I think that's what you want?

Yes, Thibaud confirmed in person (PyCon FR)

thibaudcolas:
Thanks all! I just migrated the [2024 annual impact report](https://www.djangoproject.com/foundation/reports/2024/) to this new setup, all working as expected.

PR #2310, opened 2025-11-02
Merge db settings in settings/docker.py

ulgens:
* We don't benefit from having separate definitions
* Pycharm was marking this code block with "replace dictionary creation" recommendation


ulgens:
Superseded by https://github.com/django/djangoproject.com/pull/2311

PR #2311, opened 2025-11-02
Improve docker db settings

ulgens:
Small cleanup in `settings/docker.py` file. Details can be found in commit messages.

### DIFF
--- a/djangoproject/settings/docker.py
+++ b/djangoproject/settings/docker.py
@@ -12,7 +12,8 @@ DATABASES = {
     "trac": {
         "ENGINE": "django.db.backends.postgresql",
         "NAME": os.environ["TRAC_DATABASE"],
-        "USER": os.environ["TRAC_USER"],
+        # "code.djangoproject" value is hardcoded in trac.sql
+        "USER": "code.djangoproject",
         "PASSWORD": os.environ["TRAC_PASSWORD"],
         "HOST": os.environ["TRAC_HOST"],
         "PORT": os.environ["TRAC_PORT"],
--- a/djangoproject/settings/docker.py
+++ b/djangoproject/settings/docker.py
@@ -2,20 +2,20 @@ from .dev import *  # noqa: F403
 
 DATABASES = {
     "default": {
-        "ENGINE": os.environ.get("SQL_ENGINE"),
-        "NAME": os.environ.get("SQL_DATABASE"),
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -23,7 +23,8 @@ services:
             - SQL_HOST=db
             - SQL_PORT=5432
             - TRAC_DATABASE=code.djangoproject
-            - TRAC_USER=code.djangoproject
+            # "code.djangoproject" value is hardcoded in trac.sql, no need to pass it via env vars
+            # - TRAC_USER=code.djangoproject
```
pauloxnet, line 27: Is `trac.sql` the only place where `code.djangoproject` is used?

djangoproject/settings/docker.py
```diff
@@ -2,24 +2,24 @@
...
 DATABASES = {
     "default": {
-        "ENGINE": os.environ.get("SQL_ENGINE"),
-        "NAME": os.environ.get("SQL_DATABASE"),
-        "USER": os.environ.get("SQL_USER"),
-        "PASSWORD": os.environ.get("SQL_PASSWORD"),
-        "HOST": os.environ.get("SQL_HOST"),
-        "PORT": os.environ.get("SQL_PORT"),
-    }
+        "ENGINE": os.environ["SQL_ENGINE"],
```
pauloxnet, line 5: Better tha before :)

pauloxnet (approved):

djangoproject/settings/docker.py
ulgens: The project won't work with empty database credentials. The updated approach allows it to fail with a key error instead of database connection issues.

docker-compose.yml
ulgens: I'm not 100% confident but it seems so. Regardless, the project won't work with a different value, so I thought it's better not to parameterize.

pauloxnet (approved):

alexgmin (approved):

docker-compose.yml
```diff
@@ -23,7 +23,8 @@ services:
...
             - TRAC_PASSWORD=secret
```
marksweb, line 28: If we're changing this, we could, arguably, remove all the `TRAC_` values and have them as defaults to the `os.environ.get()` calls.

PR #2313, opened 2025-11-02
Remove redundant if condition from django debug toolbar init

ulgens:
DEBUG is defined as True in the same file. Because the variable value is hardcoded, `if DEBUG` will always give same result - hence not needed

### DIFF
--- a/djangoproject/settings/dev.py
+++ b/djangoproject/settings/dev.py
@@ -44,25 +44,24 @@ PARENT_HOST = "djangoproject.localhost:8000"
 
 PUSH_SSL_CALLBACK = False
 
-# Enable optional components
+# django-debug-toolbar initialization
+try:
+    import debug_toolbar  # NOQA
+except ImportError:
+    pass
+else:
+    INSTALLED_APPS.append("debug_toolbar")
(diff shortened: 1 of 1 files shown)


djangoproject/settings/dev.py
```diff
@@ -44,25 +44,24 @@
...
-        MIDDLEWARE.insert(
-            MIDDLEWARE.index("debug_toolbar.middleware.DebugToolbarMiddleware") + 1,
-            "djangoproject.middleware.CORSMiddleware",
-        )
-
+# django-debug-toolbar initialization
+try:
+    import debug_toolbar  # NOQA
+except ImportError:
+    pass
```
ulgens, line 51: Do we have a case where developers use the dev settings without installing the dev requirements?

SaptakS: I am not sure, but I think for this PR, makes sense to keep the `try-except` block for now?

SaptakS (approved):

djangoproject/settings/dev.py
ulgens: Yep, not touching it in this PR.

PR #2321, opened 2025-11-05
Migrate to uv

ulgens:
Resolves https://github.com/django/djangoproject.com/issues/1901

Depends on:
- https://github.com/django/djangoproject.com/pull/2448

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -41,7 +41,6 @@ repos:
       rev: 25.12.0
       hooks:
           - id: black
-            args: ["--target-version=py312"]
 
     - repo: https://github.com/rbubley/mirrors-prettier
       rev: v3.7.4


ulgens:
@django/django-website Is there a way to disable pre-commit.ci?

ulgens:
@pauloxnet The setup is working now, and I think it can be reviewed. I still need to understand why black is generating the trailing commas, and I can organize the commits a bit better, but other than those, it's pretty much done. Please let me know if you see anything that you want to discuss about or needs to be updated.

pauloxnet:
> ... I still need to understand why black is generating the trailing commas ...

Do we still use black? Maybe it's time to migrate linting and formatting to ruff :smile:

ulgens:
> Do we still use black? Maybe it's time to migrate linting and formatting to ruff 😄

+100 🌻 I was a bit afraid of starting that discussion but there is no harm in creating an issue.

pauloxnet:
> > Do we still use black? Maybe it's time to migrate linting and formatting to ruff 😄
> 
> +100 🌻 I was a bit afraid of starting that discussion but there is no harm in creating an issue.

Yes, an issue would be the starting point ;-)

ulgens:
@pauloxnet #2446 🪷

ulgens:
After the last review from @pauloxnet , I pushed some changes:
- Rebased from main after #2448 , no more mysterious trailing commas
- Moved pyproject-fmt settings to pyproject.toml file
- Replaced the ulgens/uv-lock-check git hook with astral-sh/uv-pre-commit
- Updated makefile content and wrapped the necessary entries with `uv run`

pauloxnet:
Looking at it now I realized that the change is quite big in terms of lines of code.

Do you think that using the iterative approach will help everyone to review and split the complexity of this PR:
- phase 1 (compile textual requirement files with uv pip compile and install them with pip)
- phase 2 (the current one)

I'm using the phase 1 approach in some simple repository (see https://github.com/pauloxnet/djangoqueries/blob/master/Makefile)

P.S. avoid force pushing changes only because it' more difficult to see what has been changed since last review :)

ulgens:
> Do you think that using the iterative approach will help everyone to review and split the complexity of this PR

I can try that 👍🏻 

> P.S. avoid force pushing changes only because it' more difficult to see what has been changed since last review :)

I prefer push force - rebase to keep the PR history clean but I see your point, will be more careful with them.

ulgens:
Closing the PR, I'll create 2 new ones with smaller contexts.

PR #2325, opened 2025-11-06
Enable codecov

ulgens:
Coming from the discussion we had during the removal of coveralls. We are already using Sentry and it looks like codecov doesn't need any extra config.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -62,6 +62,13 @@ jobs:
                   echo '"trac_db_host": "localhost", ' >> conf/secrets.json
                   echo '"trac_db_password": "secret", ' >> conf/secrets.json
                   echo '"secret_key": "a"}' >> conf/secrets.json
+
             - name: Run tests
               run: |
                   make ci
+                  coverage xml
+
+            - name: Upload coverage reports to Codecov
+              uses: codecov/codecov-action@v5.5.1
+              with:
+                  token: ${{ secrets.CODECOV_TOKEN }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -69,6 +69,6 @@ jobs:
                   coverage xml
 
             - name: Upload coverage reports to Codecov
-              uses: codecov/codecov-action@v5.5.1
+              uses: codecov/codecov-action@v5
               with:
                   token: ${{ secrets.CODECOV_TOKEN }}


.github/workflows/tests.yml
```diff
@@ -59,6 +59,13 @@ jobs:
                   echo '"trac_db_host": "localhost", ' >> conf/secrets.json
                   echo '"trac_db_password": "secret", ' >> conf/secrets.json
                   echo '"secret_key": "a"}' >> conf/secrets.json
+
```
codingjoe, line 65: We should add https://github.com/google/yamlfmt to the pre-commit hooks ;)

.github/workflows/tests.yml
```diff
@@ -59,6 +59,13 @@ jobs:
...
             - name: Run tests
               run: |
                   make ci
+                  coverage xml
+
+            - name: Upload coverage reports to Codecov
+              uses: codecov/codecov-action@v5.5.1
```
codingjoe: Do we need the bugfix version pin? None of the other actions have it? (Though, codecov has a poor track record when it comes to versioning their action).
```suggestion
              uses: codecov/codecov-action@v5
```

codingjoe (commented):
🚀

.github/workflows/tests.yml
pauloxnet: I agree

ulgens: I prefer to use hash pinning, as recommended by zizmor, but pinned to the patch here hoping that it will comply better with the existing convention. I'm okay with pinning to the major.

.github/workflows/tests.yml
ulgens: yamlfmt can remove or keep the existing newlines, but I didn't know you could enforce adding new ones. That's aside, I like and actively use yamlfmt, I'll create a separate PR for it.

.github/workflows/tests.yml
ulgens: Fixed.

.github/workflows/tests.yml
ulgens: I created #2527

pauloxnet (approved):
LGTM

codingjoe (approved):
🥳

sarahboyce (approved):
Thank you!

sarahboyce:
@ulgens the links on the CodeCov reports are not currently working. Do you know what we need to set up?

PR #2328, opened 2025-11-09
Update --selection bg-color when default browser theme is dark

ZougaRida:
The `--selection` bg-color when default theme for browser is dark (and the user hasn't changed the theme at all, thus theme-icon-when-auto shows up) is **$(green-medium)** but the contrast is way too terrible (the actual value is 2.35 which is way too low)
django-project


SaptakS (approved):
Thanks for the PR @ZougaRida . Looks good to me!

Issue #2340, opened 2025-11-16
Rename "About" to "DSF" or "Foundation"


sarahboyce:
Python has PSF in the navigation bar: https://www.python.org/

Rails has Foundation in the navigation bar: https://rubyonrails.org/

I think these are clearer than "About" which I feel could be interpreted more about the Django framework (code). I think I have a slight preference to change to "DSF"

SaptakS:
I like the idea of having it renamed to "DSF". Makes sense to me.

PR #2342, opened 2025-11-17
Renamed "About" in header to "Foundation".

sarahboyce:
Fixes https://github.com/django/djangoproject.com/issues/2340

Screenshot from 2025-11-18 08-23-34

Did a poll on Mastodon and "Foundation" was the most popular with 28 votes, "DSF" had 2, and "About" also had 2 votes.

### DIFF
--- a/djangoproject/templates/includes/header.html
+++ b/djangoproject/templates/includes/header.html
@@ -27,14 +27,14 @@
         <li{% if 'weblog' in request.path %} class="active"{% endif %}>
           <a href="{% url 'weblog:index' %}">News</a>
         </li>
-        <li{% if 'community' in request.path or 'conduct' in request.path %} class="active"{% endif %}>
-          <a href="{% url 'community-index' %}">Community</a>
-        </li>
         <li>
           <a href="https://github.com/django/django" target="_blank" rel="noopener">Code</a>
         </li>
         <li>
           <a href="https://code.djangoproject.com/">Issues</a>
+        </li>
+        <li{% if 'community' in request.path or 'conduct' in request.path %} class="active"{% endif %}>
+          <a href="{% url 'community-index' %}">Community</a>
         </li>
         <li{% if 'foundation' in request.path %} class="active"{% endif %}>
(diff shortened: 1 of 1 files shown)


sarahboyce:
I see we got a thumbs up from Paulo on the issue, let's maybe aim for 1 more board member to thumbs up (cc @django/dsf-board )

knyghty:
I wonder if we should maybe spell it out as "foundation" but yes I have wanted to change this wording for a while.

sarahboyce:
Foundation was my other suggestion 😁
I think both work, happy to go for whatever is most popular 👍

sarahboyce:
Did a poll on Mastodon and "Foundation" was the most popular with 28 votes, "DSF" had 2, and "About" also had 2 votes. So updated the PR to rename to "Foundation"

SaptakS:
Makes sense to me. I am happy to merge this if there is no objection.

jefftriplett:
Python.org goes with PSF, so DSF felt natural. The trouble is that foundation reads differently in a tech context than in a legal one. I don’t love nonprofit either (but it's less confusing), and that’s why the DSF, while not as popular, feels less confusing.

sarahboyce:
https://mastodon.social/@sarahboyce/115566516022543101
The poll is now closed 44 votes for Foundation, 9 votes for DSF, and 3 votes to leave as About.
Some commentary includes concerns that folks won't know what the DSF is, some suggestions to have 'The DSF' or 'The Foundation'
Note that Rails also uses 'Foundation': https://rubyonrails.org/ and our url itself is 'foundation'

benjaoming:
Great change! If "Community" and "Foundation" are next to each other, I think they spell out an interesting relationship :heart: (edit: but I certainly also see why Foundation is at the end, just before Donate - that's probably best)

sarahboyce:
> If "Community" and "Foundation" are next to each other, I think they spell out an interesting relationship ❤️

I think that's a good idea 😁 if I get a few thumbs up to make this change, I'm happy to do it

PR #2345, opened 2025-11-20
Removed PyCharm 2025 discount banner.

sarahboyce:
I assume we should remove the banner :+1:


sabderemane:
@django/ops-team could you please deploy this when you have some time 🙏🏽

Issue #1955, opened 2025-02-22
The language and docs version switcher is impossible to use with a keyboard


sarahboyce:
https://docs.djangoproject.com/en/5.1/

You cannot tab to focus on the language and documentation version widget.
This means the options, which are only visible on hover, are not available to keyboard users.

PR #2347, opened 2025-11-22
Made the language and documentation version switcher accessible by keyboard.

sarahboyce:
Fixes https://github.com/django/djangoproject.com/issues/1955

Note that I removed the titles as I don't believe they are worth it

### DIFF
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -2286,88 +2286,97 @@ h2+.list-link-soup {
     z-index: 1;
     pointer-events: none;
 
-    list-style: none;
-
-    .icon {
-        margin-right: 4px;
-    }
+    display: flex;
+    flex-direction: column;
+    align-items: flex-end;
+    gap: 0.6em;
 
     ul {
         text-align: right;
+        margin: 0;
+        padding: 0;
+        list-style: none;
+        display: inline-flex;
+        align-items: center;
+        justify-content: flex-end;
+        gap: 3px;
     }
 }
 
-#doc-versions,
-#doc-languages,
-#faq-link,
-#backtotop-link {
-    margin: 0;
-    padding-top: 0.25em;
-    padding-bottom: 0.25em;
+#doc-versions li,
+#doc-languages li,
+#faq-link li,
+#backtotop-link li {
+    margin: 0 3px;
+    background: var(--body-bg);
+    border: 1px solid var(--hairline-color);
+    border-radius: 4px;
+    cursor: pointer;
+    pointer-events: auto;
 
-    &.open {
-        li {
-            display: inline-block;
+    a,
+    button {
+        display: inline-block;
+        background: var(--body-bg);
+        padding: 8px 15px;
+        border-radius: 4px;
+        border: none;
+        @include sans-serif;
+        font-size: 12px;
+        line-height: inherit;
+        color: var(--body-fg);
+    }
 
-            &.current {
-                border: 1px solid $green-light;
-            }
+    a {
+        text-decoration: none;
+        font-weight: 700;
+        color: var(--primary);
+    }
+
+    &:hover,
+    &:focus-within {
+        border-color: $green-light;
+        a {
+            color: var(--secondary);
         }
     }
+}
+
+#doc-languages,
+#doc-versions {
+    position: relative;
 
     li {
-        display: none;
-        margin: 0 3px;
-        @include sans-serif;
-        color: var(--body-fg);
-        @include font-size(12);
-        pointer-events: auto;
+        display: inline-block;
+        padding: 0;
+        position: relative;
 
         &.current {
-            display: inline-block;
-            background: var(--body-bg);
-            padding: 8px 15px;
-            border: 1px solid var(--hairline-color);
-            border-radius: 4px;
-            cursor: pointer;
-        }
-
-        &.current-link {
-            display: inline-block;
-        }
-
-        a {
-            display: inline-block;
-            background: var(--body-bg);
-            color: var(--primary);
-            text-decoration: none;
-            font-weight: 700;
-            padding: 8px 15px;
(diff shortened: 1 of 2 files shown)


docs/templates/docs/doc.html
```diff
@@ -56,12 +56,15 @@
   <div id="version-switcher">
     <ul id="faq-link">
       <li class="current-link">
-        <a href="{% url 'document-detail' lang=lang version=version url='faq/help' host 'docs' %}">
+        <a tabindex="0" href="{% url 'document-detail' lang=lang version=version url='faq/help' host 'docs' %}">
```
SaptakS: I don't think a `tabindex=0` is needed here given this is an `<a>` which is already focusable. It seems to be keyboard accessible to me in the current website too.

docs/templates/docs/doc.html
```diff
@@ -56,12 +56,15 @@
...
           <span>{% trans "Getting Help" %}</span>
         </a>
       </li>
     </ul>
     <ul id="doc-languages" class="language-switcher doc-switcher">
+      <li class="current" tabindex="0">
+        <span>{% trans "Language:" %} <strong>{{ lang }}</strong></span>
```
SaptakS: What if we made this `<span>` into a `<button>`? That would mean we won't forcibly make the listitems focusable elements given buttons are already focusable, and I feel it semantically makes more sense.

I know this would also probably mean a lot of CSS changes, and I am not sure if there are other implications that require us to keep this as `<span>` instead of `<button>`

docs/templates/docs/doc.html
```diff
@@ -74,14 +77,15 @@
...
-        <span>{% trans "Language:" %} <strong>{{ lang }}</strong></span>
-      </li>
     </ul>
 
     {% get_all_doc_versions docurl as other_versions %}
     <ul id="doc-versions" class="version-switcher doc-switcher">
+      <li class="current {% if release.is_dev %}dev{% endif %}" tabindex="0">
+        <span>{% trans "Documentation version:" %}
+          <strong>{% if release.is_dev %}development{% else %}{{ version }}{% endif %}</strong>
+        </span>
```
SaptakS: Similar comment as https://github.com/django/djangoproject.com/pull/2347/files#r2555889212

docs/templates/docs/doc.html
```diff
@@ -94,22 +98,10 @@
...
-                 {% endblocktrans %}">
-        <span>{% trans "Documentation version:" %}
-          <strong>{% if release.is_dev %}development{% else %}{{ version }}{% endif %}</strong>
-        </span>
-      </li>
     </ul>
     <ul id="backtotop-link">
       <li class="current-link">
-        <a href="#top" aria-label="Back to top" class="icon-chevron-up-align"><i class="icon icon-chevron-up"></i></a>
+        <a tabindex="0" href="#top" aria-label="Back to top" class="icon-chevron-up-align"><i class="icon icon-chevron-up"></i></a>
```
SaptakS: Same comment as https://github.com/django/djangoproject.com/pull/2347/files#r2555874182

sarahboyce:
Lovely suggestions thank you! I've tried to incorporate each of them :+1:

djangoproject/scss/_style.scss
```diff
@@ -2286,88 +2286,96 @@ h2+.list-link-soup {
...
-    &:hover {
-        pointer-events: auto;
-        padding: 2em 0;
-        margin: -1.75em 0;
-
-        li {
-            display: inline-block;
-        }
+    &:hover li.other,
+    &:focus-within li.other {
```
SaptakS: I think instead of using `focus-within` and removing `.open` usage in SCSS, it might be better to leave it the other way (i.e. use `&.open li.other` to actually show the other languages, instead of `&:focus-within li.other`).

My reasoning for this is, with &.open, the language options are shown on button press, whereas with focus-within, it shows as soon as the button is focused. I think the button press UX is better in keyboard, than automatically showing on focus, because with automatic showing with `focus-within` user will have to tab through all the languages and all the documentation version to reach the "top of the page" button. But with button press, the language and documentation versions only show if enter pressed, and otherwise can be easily tabbed through.

Thoughts?

This is how the interaction looks to me with `.open`
https://github.com/user-attachments/assets/7b616c81-2bf0-4fa9-8af8-4415816c20f2

sarahboyce: I think this is a nicer keyboard UX as well :+1: pushed a tweak

SaptakS (approved):
LGTM. Just one thing I noticed (which I don't think really breaks anything) is that the text colors on the language button and documentation version switcher was changed to black from the green in this PR. Was that intentional. Honestly, I do feel it helps increase the color contrast with the black text.

sarahboyce:
> Just one thing I noticed (which I don't think really breaks anything) is that the text colors on the language button and documentation version switcher was changed to black from the green in this PR. Was that intentional.

I hadn't noticed I'd changed it so it wasn't intentional 😅
Happy to either change it back or leave it if you prefer it

SaptakS:
In that case, maybe change it back for now, to keep it consistent. I think its ready to be merged after that.

sarahboyce:
Actually it looks the same to me locally :thinking: 

djangoproject.com

Screenshot from 2025-12-15 12-18-02


My branch

Screenshot from 2025-12-15 12-18-42

Which bit is black on my branch which is not black on djangoproject.com?

laymonage:
I think it's the "Language: en" and "Documentation version: 5.2" elements

SaptakS:
The text "Language: en" if you see in the screenshots, in your branch it is black, while in the site it is `var(--body-fg)`
image

sarahboyce:
Ah sorry thought something used to be green! Well spotted! Would have broken dark mode - pushed up a change :+1:

SaptakS:
Looks good to me. Merging.

Issue #2349, opened 2025-11-24
Hotkey for focus-on-search broken


jonbiemond:
The hotkey, `CMD` + `K`,  to focus on the search input element, seemingly has no affect on desktop. But works fine on mobile.

Commit https://github.com/django/djangoproject.com/commit/e81dd718cba062f6ace41428278621429cb2d7fc adds two search fields to the DOM, one displayed only on mobile.

```diff
--- a/djangoproject/templates/includes/header.html
+++ b/djangoproject/templates/includes/header.html
@@ -5,8 +6,11 @@
   <div class="container container--flex--wrap--mobile">
     <a class="logo" href="{% url 'homepage' %}">Django</a>
     <p class="meta">The web framework for perfectionists with deadlines.</p>
-    <div class="mobile-toggle">
-      {% include "includes/toggle_theme.html" %}
+    <div class="header-mobile-only">
+      {% search_form %}
+      <div class="light-dark-mode-toggle">
+        {% include "includes/toggle_theme.html" %}
+      </div>
     </div>
     <button class="menu-button">
       <i class="icon icon-reorder"></i>
@@ -42,6 +46,9 @@
         <li{% if 'fundraising' in request.path %} class="active"{% endif %}>
           <a href="{% url 'fundraising:index' %}">&#9829; Donate</a>
         </li>
+        <li>
+          {% search_form %}
+        </li>
         <li>
           {% include "includes/toggle_theme.html" %}
         </li>
```
The selector for the search-focus event listener returns the first search input element, and not the element displayed on desktop.
```js
  const el = document.querySelector('#id_q');
```

sarahboyce:
Thank you for raising! Opened a PR 👍

SanthoshPari:
Hi! I'd like to fix. The issue mentions two search fields – I'll update the JS selector to target the desktop one.

adamzap:
@SanthoshPari Thank you, but this is already handled in #2355.

PR #2355, opened 2025-11-24
Restored search hotkey for desktop.

sarahboyce:
Fixes https://github.com/django/djangoproject.com/issues/2349

I also added playwright to have a way to test JavaScript (might be controversial and can remove if wanted)


djangoproject/static/js/djangoproject.js
```diff
@@ -65,9 +65,15 @@ document.querySelector('.menu-button').addEventListener('click', function () {
...
+function querySelectorVisible(selector) {
+  return [...document.querySelectorAll(selector)].find(
+    (el) => el.offsetParent !== null,
+  );
+}
+
 // Update search input placeholder text based on the user's operating system
 (function () {
-  const el = document.getElementById('id_q');
+  const el = querySelectorVisible('#id_q');
```
adamzap: Would you be opposed to updating the placeholder on both elements?

It would simplify the code here and allow us to factor the `querySelectorVisible` function that you created out into the other handler. Before this commit, this file didn't add anything to the global namespace.

sarahboyce:
> I wonder if there's a way to achieve the same behavior with a single element and responsive styling. Do you or others think this is worth pursuing?

I originally tried doing this but found it challenging. The search bar duplication aligns with the light/dark mode toggle duplication and show/hide behavior, aligning to that was easier to acheive given the existing html structure

djangoproject/static/js/djangoproject.js
sarahboyce: Good idea 👍

SaptakS: I feel like its usually a good idea to use a class or some other selector and not id, if we expect multiple elements to have that id. Or maybe have id like `id_q_mobile` and `id_q` for mobile and desktop respectively. Multiple elements having same id is usually considered invalid HTML. I do like the `querySelectorVisible` function though, and I don't have a super strong opinion.

sarahboyce: This was a bit fiddly to pull off but I think I've managed it. I pushed up a second commit in case the previous version is preferred but now have unique html ids :+1:

djangoproject/static/js/djangoproject.js
```diff
@@ -94,7 +98,17 @@ window.addEventListener('keydown', function (e) {
...
+    return el && el.offsetParent !== null;
+  }
+
+  const inputs = [
+    document.getElementById('id_desktop-q'),
+    document.getElementById('id_mobile-q'),
+  ];
+
+  const el = inputs.find(isVisible);
+  if (!el) return;
```
adamzap, line 107: [`checkVisibility`](https://developer.mozilla.org/en-US/docs/Web/API/Element/checkVisibility) is baseline 2024, so maybe we can replace the `isVisible` definition and call with:

```javascript
  const el = inputs.find((el) => el.checkVisibility());
```

Or is that too modern? [This](https://github.com/django/djangoproject.com/?tab=readme-ov-file#supported-browsers) section of the README seems outdated. I'll follow up.

sarahboyce: I think we should use it for both updating the placeholder and adding the hotkey, then if it's not supported, it won't be advertised in the placeholder

djangoproject/static/js/djangoproject.js
SaptakS: It is a bit fiddly. Not to be a pain but I wonder if should have just used a class instead of an id in the first commit and just gone with that. Do you think that would be cleaner and maybe makes more sense?

sarahboyce: If I remove the id, I think I need to remove the label as this is linked to the field using the field id 🤔 
Is that acceptable?

SaptakS: Yeah you are right. That's not acceptable as well. In that case, I think the fiddly bit maybe makes sense and is worth the effort. Thanks!

djangoproject/static/js/djangoproject.js
SaptakS: I think either way is fine. @sarahboyce are you thinking of using `checkVisibility`? Apart from that, I think this PR is ready to be merged.

sarahboyce: I have pushed up a commit with this change :+1:

djangoproject/static/js/djangoproject.js
```diff
@@ -67,16 +67,18 @@ document.querySelector('.menu-button').addEventListener('click', function () {
...
+  const inputs = [
+    document.getElementById('id_desktop-q'),
+    document.getElementById('id_mobile-q'),
+  ];
 
-  if (!el) {
-    return;
-  }
+  const el = inputs.find((el) => el.checkVisibility());
+  if (!el) return;
```
adamzap: Is there a case where neither element will be rendered on the page? I'm trying to understand why we need this line.

djangoproject/tests.py
```diff
@@ -211,3 +214,64 @@ def test_single_h1_per_page(self):
...
+        expect(desktop_search_bar).to_have_attribute("placeholder", "Search (Ctrl+K)")
+        is_focused = page.evaluate("document.activeElement.id === 'id_desktop-q'")
+        self.assertFalse(is_focused)
+
+        page.keyboard.press("Control+KeyK")
+        is_focused = page.evaluate("document.activeElement.id === 'id_desktop-q'")
+        self.assertTrue(is_focused)
+        page.close()
+
+    def test_search_ctrl_k_hotkey_mobile(self):
```
adamzap, line 256: I think you need to set an explicit user agent on the two hotkey tests. I'm getting failures because I'm running them locally on a Mac.

Makefile
```diff
@@ -4,7 +4,7 @@ APP_LIST ?= accounts aggregator blog contact dashboard djangoproject docs founda
 SCSS = djangoproject/scss
 STATIC = djangoproject/static
 
-ci: compilemessages test
+ci: compilemessages collectstatics test
```
adamzap, line 7: Is this needed because of the new JavaScript tests?

djangoproject/static/js/djangoproject.js
adamzap: Thanks!

djangoproject/static/js/djangoproject.js
sarahboyce: I think this is incase there are any pages without the header. For example, I think the DSF annual report doesn't have the header

Makefile
sarahboyce: Yes 👍 
(I think this also mostly fixes https://github.com/django/djangoproject.com/issues/1831 but still need the tests to use the production STORAGES setting, can have someone check if that can be pulled into the common settings file 🤔)

djangoproject/static/js/djangoproject.js
adamzap: Ah ok, thanks!

djangoproject/static/js/djangoproject.js
```diff
@@ -67,16 +67,18 @@ document.querySelector('.menu-button').addEventListener('click', function () {
...
 
-  const original_placeholder = el.getAttribute('placeholder');
   const is_mac = navigator.userAgent.indexOf('Mac') !== -1;
-  const new_value = `${original_placeholder} (${is_mac ? '⌘\u200aK' : 'Ctrl+K'})`;
 
+  const original_placeholder = el.getAttribute('placeholder');
+  const new_value = `${original_placeholder} (${is_mac ? '⌘\u200aK' : 'Ctrl+K'})`;
```
adamzap, line 82: This is a nitpick, but the content of these lines didn't change. They only changed position. If we want to reduce the diff, we could move them back. Up to you!

adamzap (approved):
The JavaScript changes look good to me! Thanks!

I hope we can eventually move to having a single input, but that's out of scope here.

djangoproject/static/js/djangoproject.js
sarahboyce: Reduced the diff :+1:

docs/forms.py
```diff
@@ -16,3 +16,8 @@ def __init__(self, data=None, **kwargs):
                 "placeholder": search_label_placeholder,
             }
         )
+        q_with_prefix = super().add_prefix("q")
```
bmispelon, line 24: I'd love a comment to explain the `add_prefix()` shenanigans. Is this all so that we add the prefix to the `id` but not the `name`?

sarahboyce: Yes exactly, I can add a comment because it's definitely not standard 👍

adamzap:
I'll try to take another look at this by the end of the week.

adamzap:
I'd like to defer to others on this PR if someone is available to review. I'm unsure about the changes around template tags and would like to assess the difficulty of removing the duplicate navigation elements.

SaptakS:
> assess the difficulty of removing the duplicate navigation elements.

I think the way our navigation is designed, duplicating actually makes the most sense. Without duplication, the CSS would be a little complicated / fiddly. I do think that the template tags and non-duplicating ids is slightly complicated, but I think till we have any better solution (and maybe at some point a redesign changes the entire UI anyways), this PR can be merged to at least fix the bug. @adamzap what do you think?

adamzap:
I'm closing this since #2442 was merged.

PR #2358, opened 2025-11-29
Improve spacing on search form field

adamghill:
Currently, there are two spacing issues I see in Vivaldi/Chrome around the searchbar in the top nav. Not sure if this is viewport width thing or a browser thing, though.

Screenshot 2025-11-29 at 8 11 59 AM

This is what it looks like Firefox for me:

image

1. Too much whitespace to the right of the magnifying glass `button`.

Removing the explicit `width: 40px` takes away some of the whitespace around the magnifying button and makes the spacing look a little better to my eyes. When I search the repo for other `form-input` classes, I didn't see anywhere else that uses a `button` element inside the `form`, so I don't think this would affect anything else.

This is what it looks like in Vivaldi with that change:

image

And this is in Firefox:

image

2. The toggle is too close to the searchbar in Vivaldi. This does *not* seem to happen in Firefox. I have some ideas for that fix, but was thinking it might be better as a separate PR. Let me know if that makes sense or if it'd be better in this one.

Tagging @sarahboyce since she might know the most about the search bar because of https://github.com/django/djangoproject.com/pull/2262.

### DIFF
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -581,7 +581,7 @@ header {
                 border: 0;
                 float: left;
                 text-align: left;
-                > :not(.search){
+                > :not(.search) {
                   padding: 10px;
                 }
             }
@@ -3057,12 +3057,9 @@ form {
         width: 20%;
         min-width: 200px;
         margin: 10px;
-        margin-left: 0;
+
         input {
           margin: 0;
-          @include respond-min(1080px) {
-            margin: 0 10px;
-          }
         }
     }
 }


sarahboyce:
I agree this looks better for desktop screens but tablet and mobile now look a little close to me

Screenshot from 2025-11-29 18-36-37

I feel like the 40px width is fine as the height is also 40px, the main issue to me is that it is not positioned at the end of the input and the same with mobile. On mobile it is closer to the end than it is on desktop screens

Screenshot from 2025-11-29 18-59-52

adamghill:
>the main issue to me is that it is not positioned at the end of the input and the same with mobile

Ah, hmm, ok let me try keeping the 40px width, but a different tweak.

sarahboyce:
I have created an alternative PR https://github.com/django/djangoproject.com/pull/2365 - @adamghill can you test and see what you think?

adamghill:
> I have created an alternative PR #2365 - @adamghill can you test and see what you think?

I replied! I think moving to flexbox is probably the better approach. But, if we don't want to do that I just pushed a change to this branch if you want to check it out. @sarahboyce

sarahboyce:
Tested and this looks great! It's also a much smaller change so I like it :grin: 
I will close the other PR :+1:

bmispelon:
I've tested it locally and the spacing does look better.

Thanks everyone!

PR #2365, opened 2025-12-02
Fixed global search button spacing.

sarahboyce:
Alternative to https://github.com/django/djangoproject.com/pull/2358
Uses flexbox


adamghill:
flexbox is probably the right way to do this, but the top nav is now much taller for me (in both Firefox and Vivaldi at desktop size). Does this happen for you as well?

image

Issue #2220, opened 2025-09-22
Monthly archive heading only shows up as "archive"


laymonage:
The `<h1>` in the monthly blog archive pages only shows up as "archive" instead of e.g. "August 2025 archive":

https://www.djangoproject.com/weblog/2025/aug/

<img width="391" height="291" alt="News & Events, 'archive' heading" src="https://github.com/user-attachments/assets/b3dc1783-35da-4ec1-92d9-180b6ef05f82" />

It still worked fine on November 5 2024: https://web.archive.org/web/20241105093852/https://www.djangoproject.com/weblog/2024/nov/

But it already broke on December 8 2024: https://web.archive.org/web/20240801000000*/https://www.djangoproject.com/weblog/2024/nov/

This likely was a regression in 910e7c862c4e3863024b94d6746b3ce4e0a47050 in #1669.

ManshaAgarwal716:
Sir,I have raised the pull request (https://github.com/django/djangoproject.com/pull/2229), you can once please review it.

PR #2380, opened 2025-12-08
Fixed blog archive headers.

sarahboyce:
Fixes https://github.com/django/djangoproject.com/issues/2220

### DIFF
--- a/blog/tests.py
+++ b/blog/tests.py
@@ -493,7 +493,7 @@ class ViewsTestCase(ReleaseMixin, DateTimeMixin, TestCase):
             {
                 "view": "weblog:archive-day",
                 "kwargs": {"year": year, "month": month, "day": day},
-                "header": "<h1>July 21 archive</h1>",
+                "header": "<h1>July 21, 2005 archive</h1>",
             },
         ]:
             with self.subTest(view=testcase["view"]):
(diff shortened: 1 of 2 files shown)


blog/tests.py
```diff
@@ -466,6 +466,44 @@ def test_user_cannot_see_unpublished_entries(self):
...
+            },
+            {
+                "view": "weblog:archive-month",
+                "kwargs": {"year": year, "month": month},
+                "header": "<h1>July 2005 archive</h1>",
+            },
+            {
+                "view": "weblog:archive-day",
+                "kwargs": {"year": year, "month": month, "day": day},
+                "header": "<h1>July 21 archive</h1>",
```
adamzap: This is probably out of scope, but I wonder if we need to support the day archive view at all. If we do, should we add the year to the header?

adamzap (approved):
👍 Looks great!

My comment is not blocking.

blog/tests.py
sarahboyce: Good point! I don't think it makes much sense and I agree about the header

I'll add the year to the header and let the @django/django-website wg discuss if we should just get rid the day archive view in their next meeting 👍

adamzap: Thanks!

PR #2221, opened 2025-09-22
Updates minutes links

CaptainJack2491:
fixes #2079 
The new link points to the new minutes repo. Also updated the tests related to it.

Issue #1880, opened 2025-01-19
Broken links on Donate page


pbratkowski:
On the page <https://docs.djangoproject.com/foundation/donate/>, the "get in touch links" under the following two sections are broken:

- "Do you accept donations from companies/corporations?"
- "Can I choose a specific feature or project to support?"

Both point to <https://docs.djangoproject.com/contact/foundation/>, which returns a `404`.

pbratkowski:
Upon further investigation, I believe that this page should not be accessed on the `docs` subdomain at all.

If you open the Django website (<https://www.djangoproject.com/>), everything works as expected.

However, if you open <https://docs.djangoproject.com/en/5.1/>, the link **Benevity Workplace Giving Program** in the footer, using a relative URL, links to <https://docs.djangoproject.com/foundation/donate/#benevity-giving>

I guess the fix is to ensure the footer link uses the absolute URL, including the host (and potentially adding a redirect to the `www` subdomain for `docs.djangoproject.com/foundation/` to preserve existing links).

pbratkowski:
Also when clicking the Corporate Membership links in the footer.

https://github.com/django/djangoproject.com/blob/20cec4557cf4a2d92e8bce48c7426607d32f3c83/djangoproject/templates/includes/footer.html#L60-L62

bmispelon:
I think this was reported before as #1590, but your issue has more information so I'll keep this one open and close the other.

As you pointed out, this has to do with the `django-hosts` setup and some links being domain-relative when they should be absolute.

I'm not too familiar with how `django-hosts` works, but I think using `{ url ... %}` in the templates (rather than hardcoding the URL) should work to fix this issue. I don't know if that's even possible, but it would be nice to be able to check if there are similarly broken links in other templates 🤔

AhmedNassar7:
Hi @bmispelon , 

I’d like to work on this issue. Just to confirm, should I proceed with updating the links using `{ url ... %}` in the templates, or is there another preferred approach? 

Thanks!

PR #2384, opened 2025-12-08
Rebase meeting minutes pr2221

jcjudkins:
This PR rebases and continues the work from #2221 by @CaptainJack2491.

## Changes
- Updates Django meeting minutes links to point to the new minutes repository
- Removes old meeting minute references and template tags
- Adds redirects for old minute URLs
- Updates tests to reflect the new structure

## Context
The original PR (#2221) needed rebasing due to conflicts with recent changes. This PR applies those changes on top of the current main branch.

Closes #2221


SaptakS:
@knyghty can you take a look at this rebase and if all looks good to you. This is the rebased #2221 that you and @CaptainJack2491 had been working on.

knyghty (approved):
It looks correct to me. Thanks for the extra work here.

sarahboyce:
I just want to question whether we want to drop the tables around terms and board members. 
We have some existing functionality where user accounts have a profile page. Adding the past/existing board members terms/responsibilities onto these pages seems quite a reasonable request and relatively easy thing to acheive currently. Are we sure we don't want to keep these existing tables/data before we drop it all?

knyghty:
@sarahboyce I think you're right to question that but I can't say I have a very compelling answer or opinion either way. I think it is nice to have this if we show it somewhere. And in order to show it somewhere maybe nice to not delete it and the data with it. I question a bit whether we'd keep a database table up to date that isn't actually displayed anywhere, but I suppose if we did agree to implement something like this in the near future I can definitely see the benefit.

At the moment though will still have the [board listing](https://www.djangoproject.com/foundation/#board-history) as a flatpage and I'd be tempted to say that's enough. But I also see your point about being able to re-use this information elsewhere.

So as usual I just wrote a longwinded way of saying "dunno", which is not extremely helpful for anyone, so maybe we should wait and hope for other opinions.

sarahboyce:
Personally I would remove the deletion of the board and terms from this PR and have this in a follow up PR to be discussed/approved separately.

The information only needs updating once a year, not updating it for 3 years is not too difficult to backfill. There might be no plans to use this for making the board list dynamic, displaying past boards, using for permissions, profiles etc. But I currently see potential, so I think the @django/django-website wg should discuss it before we drop the tables.

knyghty:
Makes sense to me.

SaptakS:
@jcjudkins can you make the changes as discussed in the above comments and also resolve the conflict?

ulgens:
> Personally I would remove the deletion of the board and terms from this PR and have this in a follow up PR to be discussed/approved separately.

> But I currently see potential, so I think the @django/django-website wg should discuss it before we drop the tables.

+1 on both points.

ulgens:
I was checking #2221 and came back to this PR. Apart from the existing change request, I'll have two recommendations:

* Please install git hooks in your local following https://github.com/django/djangoproject.com?tab=readme-ov-file#git-hooks and make sure your commits pass the checks. pre-commit.ci's fix commits pollute the history. A cleaner history with atomic commits is easier to review.
* Please update your branch by rebasing and not the "merge x branch in" approach. It pollutes the history, breaks its linearity, and sometimes causes unrelated changes to appear in the PR diff.

jcjudkins:
Updated the branch with the following changes:

Removed the deletion of board/terms tables as discussed — this will be handled in a separate PR
Resolved merge conflicts

Removed the deletion migration (0008_remove_approvedcorporatemember_approved_at_and_more.py)

Restored missing imports to foundation/models.py

Cleaned up git history — removed pre-commit.ci fix commits, added .venv/ and .idea/ exclusions to pre-commit config

Installed git hooks locally so pre-commit.ci won't need to intervene going forward

All 3 foundation tests passing

Ready for re-review.

jcjudkins:
@SaptakS @sarahboyce @knyghty — branch has been rebased onto main and is ready for re-review. Summary of changes since last review:

Removed the board/terms table deletion (scope reduced per discussion — to be handled in a separate PR)

Resolved all merge conflicts

Removed the deletion migration

Cleaned up views.py, tests.py, admin.py, urls/www.py, and sitemaps.py to properly reflect the removal of meeting minute views/feeds while keeping board-related functionality intact

All 456 tests passing

djangoproject/templates/base_foundation.html
```diff
@@ -15,24 +15,22 @@
...
     <ul class="list-links">
-      <li><a href="{% host_url 'homepage' host 'www' %}foundation/faq/">FAQ</a></li>
-      <li><a href="{% host_url 'homepage' host 'www' %}foundation/records/">Records</a></li>
+      <li><a href="/foundation/faq/">FAQ</a></li>
+      <li><a href="/foundation/records/">Records</a></li>
       <li><a href="{% url 'contact_foundation' %}">Contact us</a></li>
-      <li><a href="{% host_url 'homepage' host 'www' %}foundation/cla/">Contributor license agreements</a></li>
-      <li><a href="{% host_url 'homepage' host 'www' %}foundation/conferences/">Organizing a Django conference</a></li>
+      <li><a href="/foundation/cla/">Contributor license agreements</a></li>
+      <li><a href="/foundation/conferences/">Organizing a Django conference</a></li>
```
sarahboyce: Please revert these changes, this was added in a29ae32066bf63d2dfabaefe9ea227d4cb63fc0f to resolve https://github.com/django/djangoproject.com/issues/1880

djangoproject/templates/base_foundation.html
```diff
@@ -15,24 +15,22 @@
 {# Always include <h2> label and <div> with aria role. #}
   <div role="complementary">
     <h2 class="visuallyhidden" id="aside-header">{% translate "Additional Information" %}</h2>
-
-    {% block content-related-extra %}{% endblock %}
-
```
sarahboyce, line 21: This change should also be reverted, this was added as part of 4ff6a93bf4c5f2193e020c151ff8d6b6fe8746d2

accounts/tests.py
```diff
@@ -220,17 +219,6 @@ def test_deletion_staff_forbidden(self):
...
-    def test_user_with_protected_data(self):
-        form = self.create_user_and_form()
-        form.user.boardmember_set.create(
-            office=foundationmodels.Office.objects.create(name="test"),
-            term=foundationmodels.Term.objects.create(year=2000),
-        )
-        form.delete()
-        self.assertFormError(
-            form, None, ["User has protected data and cannot be deleted"]
-        )
```
sarahboyce, line 232: We can keep this test

accounts/forms.py
```diff
@@ -64,13 +64,6 @@ def clean(self):
             raise forms.ValidationError(_("Staff users cannot be deleted"))
         return cleaned_data
 
-    def add_errors_from_protectederror(self, exception):
-        """
-        Convert the given ProtectedError exception object into validation
-        errors on the instance.
-        """
-        self.add_error(None, _("User has protected data and cannot be deleted"))
-
```
sarahboyce, line 73: I don't think we want to delete this, we use this below

sarahboyce (commented):
Thank you for the updates @jcjudkins !

Issue #2391, opened 2025-12-13
[Proposal] Scheduled daily github workflow to run tests using Django's main branch


sarahboyce:
Was listening to DjangoChat and how we don't get enough testers for pre-releases. It was also mentioned that folks may not know how to do this

I think djangoproject.com should lead by example. We should have our tests run against Django main everyday with warnings enabled. Other folks can then look at this workflow as an example for their own project. We are also contributing to testing pre-released Django 👍

SaptakS:
I like the idea.

ulgens:
> I think djangoproject.com should lead by example

Yes, but I don't think that best practice applies to this project. One important point is that this repo is a project, not a library. It will not be used with multiple Django versions at any given time, so I think it's perfectly okay to have a single version in tests.

ulgens:
Also, it's not clear how long we will be able to keep this project up to date with the latest version of Django, due to Python compatibility issues with Trac. I'll recommend finalizing

- https://github.com/django/djangoproject.com/pull/2234
- https://github.com/django/djangoproject.com/issues/2265

first so that we can have a healthier discussion.

sarahboyce:
Django projects, not just libraries, should test pre-releases ideally on a regular basis 🙂 
I think a test matrix doesn't fit this use-case so well, as there is no need to have this run on every PR. It's more about having our tests run periodically against Django main (could even be once a week, once a month etc) to test that Django itself hasn't accidentally broken something for our project

ulgens:
I don't disagree with the intention, but I still don't think it's needed or beneficial. We can always create a new branch with a new Django version (or main) and can run tests on it.

sarahboyce:
I agree that it's not needed. More of a nice-to-have than a true need 👍 

In case we are discussing the benefit of doing anything at all, then the benefit is, if something got merged into Django that broke some existing functionality accidentally that we relied on, we would catch this early and raise a ticket. They would then fix it before that breaking change is even included in a release. Otherwise, we would need to wait a month for a patch release of Django before we would upgrade versions (if we were only checking once a final release is out).
That's the benefit of anybody (whether it's a library or a normal Django project) testing against Django main on a regular basis. Having a scheduled job would mean this happens automatically. In theory, someone could do this manually every now and again

However, you may be trying to discuss an implementation detail rather than the benefit of doing something here? I'm open to other ways to approach this :)

Things to consider:
[...]

PR #2395, opened 2025-12-14
Added workflow to have tests run against Django main.

sarahboyce:
See https://github.com/django/djangoproject.com/actions/runs/20209878458/job/58014095773?pr=2395 as an example run.

Fixes https://github.com/django/djangoproject.com/issues/2391

### DIFF
--- a/.github/workflows/tests-django-main.yml
+++ b/.github/workflows/tests-django-main.yml
@@ -1,0 +1,68 @@
+name: Tests (Django main)
+
+on:
+    schedule:
+        - cron: "0 3 * * *"
+    workflow_dispatch:
+
+env:
+    PYTHON_VERSION: 3.12
+    PYTHONWARNINGS: default
+
+jobs:
+    tests:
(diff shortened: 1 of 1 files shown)


ulgens (changes requested):
I don't think this is needed, added my comment on the issue: https://github.com/django/djangoproject.com/issues/2391#issuecomment-3655210284 Also, even if we consider it's needed, I'd recommend adding a test matrix (which we had in the past but removed because it's not needed) and not a new workflow.

ulgens:
One more "change requeest" here would be that we are in the process of moving the test setup to Docker so if we are adding anything, I'd expect them not to ignore the Docker config.

cc @tobiasmcnulty

sarahboyce:
Marked the pr as draft as need the website wg to agree on the proposal 👍

ulgens:
> Marked the pr as draft as need the website wg to agree on the proposal 👍

Adding to the agenda for the next meeting.

PR #2400, opened 2025-12-17
Created a DSF donate banner to encourage end of year donations.

sarahboyce:
Quick banner that mostly reuses what we already have. I changed the banner color to have a different color to the cta button

Desktop 

Screenshot from 2025-12-17 18-53-20

Mobile

Screenshot from 2025-12-17 18-55-02

### DIFF
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -654,7 +654,7 @@ header {
 
 /* banner for billboard */
 #billboard {
-    background-color: $green-medium;
+    background-color: $green-light;
     border-top: 1px solid $black;
 
     * {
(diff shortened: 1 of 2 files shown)


pauloxnet (approved):

ulgens:
Do we any limits - rules about the size of the banner?

sarahboyce:
> Do we any limits - rules about the size of the banner?

I don't think we have a guideline here and I agree we should. I also think we may want to improve the design. I have added an issue to track this request: https://github.com/django/djangoproject.com/issues/2402

SaptakS:
cc-ing @bmispelon @django/ops-team  for deployment to make the changes visible in the website.

PR #2405, opened 2025-12-18
Fixed the version and language switcher for multiple versions/languages.

sarahboyce:
Stupidly, I was only testing with 1 extra language and 1 extra version locally. This is overlapping horribly on production making the switcher unuasable. This is now fixed

### Before on djangoproject.com

Screenshot from 2025-12-18 13-47-30

### After

Screenshot from 2025-12-18 13-48-04


alexgmin (approved):
@django/django-website-mergers This is urgent since we're going to start getting a lot of complaints very soon after the deploy 90 minutes ago.

Issue #2167, opened 2025-08-15
Duplicated dev docs warning on docs theme


thibaudcolas:
On docs pages in the dev version like [Using Django](https://docs.djangoproject.com/en/dev/topics/), our "This document is for Django's development version," warning is duplicated in the page HTML. Once at the top of the `<body>`, and another time at the bottom.

This seems to come from this JS script, that creates a clone of the element and then adds it to the page:

https://github.com/django/djangoproject.com/blob/6fcf41e44aec3ec9d62fef126a8390b94a88224a/djangoproject/static/js/djangoproject.js#L162-L167

I don’t think this is a big issue but definitely not good for users of assistive tech that the same content is on the page multiple times. The JS should be simple to change, but I’m not sure how much the styling of the warning relies on there being two copies like this.

adamzap:
The JavaScript has done it that way for a very long time, and I'd like to change it. See #1988.

I wonder if we'd be open to a new UI/UX for these warnings.

SachinAditya:
Hi mentors,
I’d like to work on this issue. I’m currently reviewing the code to understand the cause of the duplicated warning and will propose a fix that resolves it while keeping accessibility and styling in mind.

I’ll share a pull request once it’s ready. Please let me know if there’s anything specific you’d like me to consider.

Thanks!

adamzap:
@SachinAditya Thanks, but #2408 is already open and will close this issue.

SaptakS:
This is fixed by #2408 . Closing.

PR #2408, opened 2025-12-19
Use CSS instead of JavaScript for floating warning

### LINKED ISSUE #1988: Use CSS instead of JS for floating docs banner

adamzap:
This PR fixes #1988 and #2167 and is a second attempt at #1799. [This comment](https://github.com/django/djangoproject.com/issues/1988#issuecomment-3204118200) by @thibaudcolas was especially helpful in arriving at this implementation.

---

Some of the functionality of the floating warning on unsupported and in-development documentation versions was implemented using JavaScript. This is longer necessary if we use the following approach with modern CSS:

Attach the warning element the top of the page with `position: sticky`. This almost solves the problem, but we still have to scroll past header elements when an internal page link is clicked. This is achieved with a combination of the `:has` pseudo-class and the `scroll-margin-top` property.

We apply `scroll-margin-top` to elements that can be scrolled to by their `id` attribute. This is done with responsive breakpoints to get appropriate spacing when the floating warning has one, two, or three lines of text. Thanks to `:has`, this only happens when a warning element is on the page.

The modern CSS features used are widely-supported:

- https://caniuse.com/css-has
- https://caniuse.com/css-sticky
- https://caniuse.com/mdn-css_properties_scroll-margin-top

Other than making the primary CSS changes and removing the related JavaScript, this patch does the following:

- Simplify the CSS so that the class on all floating warning elements implements the common rules and the CSS for the IDs on the elements implements the colors
- Add a `z-index` to warning elements so that they always display in front of the header element
- Move the floating warning element to the top of the page in the HTML
  - Remove the `body_extra` block that was only being used by the floating warning element
  - Add a new `before_header` block for the floating warning element

Thank you to @bmispelon and @thibaudcolas for helping me arrive at this solution.

---

image

---

To test this PR:

1. Check that the floating warning element behaves as expected on the documentation part of the site. You can see it by navigating to the development version of the documentation or an unsupported version like 3.2.
2. Check that both clicking on and navigating to internal page links keeps the desired header in view.
3. Check that the responsive breakpoints I chose work well for window sizes where the warning text is one, two, or three lines.

### DIFF
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -2383,13 +2383,13 @@ h2+.list-link-soup {
     }
 }
 
-@include respond-max(674px) {
+@include respond-max(767px) {
     body:has(.doc-floating-warning) [id] {
-        scroll-margin-top: 8.4rem;
+        scroll-margin-top: 0;
     }
 }
 
-@include respond-min(675px) {
+@include respond-min(768px) {
     body:has(.doc-floating-warning) [id] {
         scroll-margin-top: 6.3rem;
     }
@@ -2402,7 +2402,6 @@ h2+.list-link-soup {
 }
 
 .doc-floating-warning {
-    position: sticky;
     top: 0;
     width: 100%;
     padding: 8px 20px 8px;
@@ -2414,6 +2413,7 @@ h2+.list-link-soup {
     z-index: 1;
 
     @include respond-min(768px) {
+        position: sticky;
         min-width: 768px;
     }
 }


djangoproject/scss/_style.scss
```diff
@@ -2383,9 +2383,26 @@ h2+.list-link-soup {
     }
 }
 
-#dev-warning,
-#outdated-warning {
-    position: absolute;
+@include respond-max(674px) {
+    body:has(.doc-floating-warning) [id] {
```
adamzap, line 2387: `[id]` feels too generic here, but I can't think of a downside. I think we'd want any element that can be scrolled to to be affected.

pauloxnet (approved):
LGTM

sarahboyce:
Thank you for this! Always love removing JavaScript :heart: 

Links with a header anchor on mobile loose the headers for me

Screenshot from 2025-12-19 21-10-09

Currently on mobile for docs.djangoproject.com, the warning is not sticky. Do you think this should do the same to simplify the scroll-margin-top calculation?

adamzap:
@sarahboyce Thank you! I overlooked the behavior change at mobile width. It should be addressed in 33ba68d.

sarahboyce (approved):
Thank you! This looks great ⭐

Issue #2414, opened 2025-12-22
UI ISSUE: Navbar layout inconsistency: extra line above logo on Documentation page; missing theme toggle on Issues page


kehach07:
The navbar layout is inconsistent across pages. On the Documentation page, the navbar shifts to an extra line above the logo, creating unnecessary vertical space, while this issue does not occur on the Issues page. Additionally, the light/dark mode toggle is missing on the Issues page.

<img width="2789" height="295" alt="Image" src="https://github.com/user-attachments/assets/c1bf81c1-e06a-405b-92f2-a68f2ba19bda" />

<img width="2829" height="404" alt="Image" src="https://github.com/user-attachments/assets/0eb245ef-a2f4-4588-a957-735f52f6800d" />

uzman2406:
Heyy ! I can solve this issue as I had already encountered that type of issue in one of my project .

Proposed approach to fix this issue:

I’d start by comparing the navbar templates used on the Documentation and Issues pages to identify structural or CSS differences. The extra vertical space on the Documentation page seems likely caused by a layout wrapper, flex/grid configuration, or breakpoint-specific styles that push navbar elements onto a separate line.

Steps I would follow:

Inspect the HTML structure to confirm whether both pages use the same base navbar template and container hierarchy.

Compare applied CSS classes and media queries to find what causes the navbar to wrap above the logo on the Documentation page.

Normalize the navbar layout so the logo, links, and controls remain on a single row across all pages.

Investigate why the light/dark mode toggle is missing on the Issues page—likely due to a conditional render, missing include, or template override—and ensure it’s consistently included wherever the navbar is used.

Test the fix across different screen sizes to confirm responsive behavior is preserved.

Happy to take this up and submit a PR if this approach looks good. Please assign this issue to me.

AyushPatiTripathi:
I would like to work on this issue

AyushPatiTripathi:
@kehach07 @sarahboyce  
I’ve opened a PR addressing the docs header layout issue (navbar wrapping/extra spacing). The fix is limited to the documentation header to keep scope small. Please have a look when convenient — feedback welcome. Thanks! #2423

alex:
Do not ping random users, its extremely rude.

I have had no involvement in Django for many many years at this point.

AyushPatiTripathi:
@alex sorry for the above

kehach07:
This is my PR :- https://github.com/django/djangoproject.com/pull/2426
[...]

PR #2423, opened 2025-12-30
Navbar layout inconsistency: issue #2414 fixed

AyushPatiTripathi:
### Summary

This PR fixes a layout inconsistency in the Documentation site header where the
navigation could wrap onto an extra line above the logo, creating unnecessary
vertical spacing. The issue was caused by a redundant Docs-specific header block
being rendered in addition to the global site header.

### What changed

- Removed the redundant documentation-only header block
- Ensured the global site header is used consistently on Docs pages
- Prevented unintended flex wrapping at desktop widths

### Notes on the theme toggle

While investigating this issue, I noticed that the light/dark mode toggle is not
present on the Issues site. After further inspection, the Issues pages are served
by Trac and do not share Django’s template stack or frontend assets.

I have **not attempted to address this in this PR**, as it appears to be outside
the scope of the Django site templates. I’m happy to open a follow-up issue or
work on this if there’s interest in coordinating a solution on the Trac side.

### Screenshots / Verification
Screenshot From 2025-12-30 21-16-21
Screenshot From 2025-12-30 23-24-03

- Docs navbar remains on a single line at desktop widths
- No regression on the main site or other sections

### DIFF
--- a/docs/templates/base_docs.html
+++ b/docs/templates/base_docs.html
@@ -9,8 +9,6 @@
   container--flex container--flex--wrap--mobile
 {% endblock %}
 
-{% block header %}
-  <p><a href="{% block doc_url %}{% url 'homepage' %}{% endblock %}">{% trans 'Documentation' %}</a></p>
-{% endblock %}
+
 
 {% block layout_class %}sidebar-right{% endblock %}


adamzap:
Thanks for your contribution, but I don't think we want to remove the documentation header to fix this issue since that would be a pretty significant design change.

I will close this PR for now. If you feel strongly about removing the documentation header, could you propose it in the issue?

PR #2425, opened 2025-12-31
Removed the donation banner to reach the 2025 fundraising goal.

sarahboyce:
It is already 2026 in some parts of the world. Ideally we should merge and deploy before midnight AoE: https://time.is/Anywhere_on_Earth


SaptakS:
CC-ing @django/ops-team for deploy

PR #2427, opened 2026-01-03
Keep github actions up to date with dependabot

p-r-a-v-i-n:
I think its worth automating github actions with dependabot which reduce manual intervention of creating pr to update github actions.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,3 +7,11 @@ updates:
           day: "monday"
           time: "12:00"
           timezone: "Europe/Oslo" # hipp hipp hurra
+
+    - package-ecosystem: github-actions
+      directory: /
+      schedule:
+          interval: "weekly"
+          day: "monday"
+          time: "12:00"
+          timezone: "Europe/Oslo" # hipp hipp hurra


.github/dependabot.yml
```diff
@@ -7,3 +7,14 @@ updates:
...
+
+    - package-ecosystem: github-actions
+      directory: /
+      groups:
+          github-actions:
+              patterns:
+                  - "*" # Group all Actions updates into a single larger pull request
+      schedule:
+          interval: weekly
+      cooldown:
```
p-r-a-v-i-n: added cooldown period as mention in https://docs.zizmor.sh/audits/#dependabot-cooldown

ulgens (changes requested):
I don't support investing in Dependabot too much, considering how horribly maintained it is, but it works fine with not-pinned actions so this should be okay for now.

Several notes:
- Please install git hooks as described in https://github.com/django/djangoproject.com?tab=readme-ov-file#git-hooks and apply the hooks on your commit.
- We can use the existing grouping and scheduling settings for now, better to be in sync with the existing config.
- I'm supportive of using `cooldown` but it feels like that should be a different PR.

.github/dependabot.yml
p-r-a-v-i-n: I agree

Stormheg (approved):
Nice!

PR #2429, opened 2026-01-04
Redirect /about to /foundation

### LINKED ISSUE #2428: Redirect from /about to /foundation
Coming from Stephanie Goulet's Slack message:

> Not urgent at all, but a recommendation that I wanted to put somewhere so as not to forget:
> * Redirect "https://www.djangoproject.com/about" so it goes to "https://www.djangoproject.com/foundation/" instead of a 404 page

https://django-dsf.slack.com/archives/C090VEVQCQ1/p1765636588262319

The rest of the discussion and a reference to another discussion can be found in the thread.

---

The expected implementation is a new `RedirectView` entry in the urls.

CodeVishal-17:
Adds a RedirectView to redirect `/about` to `/foundation` instead of returning a 404.

Fixes #2428


ulgens (changes requested):
Several notes:

* Please install git hooks as described in https://github.com/django/djangoproject.com?tab=readme-ov-file#git-hooks and apply the hooks on your commit.
* We can use `reverse()` for getting the foundations path URL, instead of hardcoding it.
* Please add tests to validate the new behaviour.

CodeVishal-17:
Thanks for the review!

I’ve updated the redirect to use pattern_name (via reverse()) instead of a hardcoded URL and added tests to validate the new behavior.

I’ll also make sure to have the git hooks installed locally for future commits.

djangoproject/urls/www.py
```diff
@@ -26,6 +26,13 @@
 
 urlpatterns = [
     path("", TemplateView.as_view(template_name="homepage.html"), name="homepage"),
+    path(
+        "about/",
+        RedirectView.as_view(
+            pattern_name="members:developer-members",
+            permanent=True,
```
ulgens, line 35: Not sure about `permanent=True`. @django/django-website Any thoughts?

CodeVishal-17: Good question. I added permanent=True assuming this is intended as a long-term redirect since /about/ shouldn’t be served anymore and should consistently point to /foundation/.

That said, I’m happy to remove it or switch to a temporary redirect if that’s preferred here — happy to follow the project’s convention.

CodeVishal-17:
Thanks again for the review!

I believe I’ve addressed all the feedback and wanted to check if anything else is needed from my side. Happy to adjust permanent=True or anything else based on the project’s preference.

ulgens:
@django/django-website Needs input / review about `permanent` arg.

djangoproject/urls/www.py
SaptakS: We discussed in the Website WG meeting, and we agreed that `permanent=True` makes sense here.

ulgens:
I'd recommend squashing the PRs before merging, LGTM otherwise.

ulgens:
@CodeVishal-17 You don't need to create more `merge branch main into...` commits, what I was looking for is a one or two clean commits in the PR and with linear history if possible. I guess that can be handled by a squash merge, cc @django/django-website-mergers

Issue #2399, opened 2025-12-17
Docs UX: Optional “copy to clipboard” button for code blocks


aadeina:
Hello,

While reading Django’s documentation, I noticed that code examples currently require manual text selection to copy. This isn’t a bug, but a small documentation UX improvement that could make examples easier to use, especially for beginners and mobile users.

If this is considered acceptable for Django’s documentation, I would be interested in working on a lightweight, accessible implementation and submitting a PR for review.

Please let me know if this is something the maintainers would be open to.

Thank you.

sarahboyce:
Note that this is related to and possibly duplicates https://github.com/django/djangoproject.com/issues/1276

aadeina:
Appreciate the note, @sarahboyce . I’d be happy to contribute a simple, user-friendly “copy to clipboard” as related to #1276.

im-anishraj:
Hi maintainer 👋  
I’d like to work on this issue.  Please let me know if I can proceed. Thanks!

VidhiKathiriya05:
Hello @aadeina,
I’m following up here after the reference in #1276. I would love to take on this task.

As I've been examining the HTML structure of the Django documentation, I've developed a strategy to make sure this is both practical and lightweight:
Intelligent Copy Logic: The script will specifically target and remove the 'Generic Prompt' elements (the $ and >>> symbols denoted by the.gp class) rather than just copying them. This guarantees that the code will run instantly in the user's terminal or shell when they paste it.

Zero Dependencies: To keep the documentation quick and manageable, I will use Vanilla JavaScript and the contemporary Clipboard API.

Accessibility: I'll make sure the button offers clear visual feedback (e.g., changing the label to 'Copied!' for a few seconds) and is keyboard-accessible (using <button> tags).

VidhiKathiriya05:
"Hi @aadeina, I received an email notification that I was welcome to work on this, but the comment isn't appearing here on the web interface. I'm ready to start! Could you please assign this issue to me? I'll follow my plan for a lightweight, Vanilla JS implementation. Thanks!"

adamzap:
Closed by #2431.

PR #1434, opened 2023-11-26
Fix #1276: Add copy to clipboard icon for commands

prashantpandey9:
- Fixes #1276 
- This fix utilizes the existing copy to clipboard icon.

## Screenshots for reference

Before:
image

After:
- Light Theme (Windows):
image

- Light Theme (Linux/Mac) 
image

- Dark Theme (Linux/Mac)
image

- Dark Theme (Windows)
image

- Copied Message
 image

PR #2431, opened 2026-01-05
Added copy to clipboard button for console tab commands

aadeina:
## Summary

Adds a **copy-to-clipboard button** to console tab code blocks (Unix / Windows installation commands).

**Fixes:** #1276, #2399  
Builds on feedback from **@sabderemane** in PR #1434.

### JavaScript  
`djangoproject/static/js/djangoproject.js`

- Extends existing clipboard functionality to support **console tabs**
- Adds smart command text extraction that excludes:
  - Shell prompts (`.gp` — e.g. `$`, `...>`)
  - Command output (`.go`)
- Minimal change set (~20 lines added), reusing existing patterns and logic

### SCSS  
`djangoproject/scss/_console-tabs.scss`

- Adds `position: relative` to console tab sections
- Introduces `.btn-clipboard` styling using **absolute positioning** (no `float`)
- Includes hover states and “Copied!” success feedback animation
- Uses existing CSS variables for **light/dark mode compatibility**

### Before
Before: console tabs without copy button

### After
https://github.com/user-attachments/assets/c6c6091e-438c-4e80-b66c-7ed72a591bae

### DIFF
--- a/djangoproject/scss/_console-tabs.scss
+++ b/djangoproject/scss/_console-tabs.scss
@@ -34,7 +34,6 @@
     >section {
         display: none;
         text-align: left;
-        position: relative;
 
         .highlight {
             margin-top: 0;


ulgens:
This looks nice 🌻 I'd recommend installing git hooks (you can follow [the last section of the readme](https://github.com/django/djangoproject.com?tab=readme-ov-file#git-hooks)) and rebasing the PR to have a bit more clear history.

aadeina:
Hi @ulgens ,
I’ve cleaned up the branch and ensured the PR now contains only the intended commit.
Please let me know if anything else is needed.

djangoproject/static/js/djangoproject.js
```diff
@@ -130,25 +134,48 @@ document.querySelectorAll('.btn-clipboard').forEach(function (el) {
       this.remove();
     });
 
-    function on_success(el) {
+    function on_success() {
```
adamzap, line 147: Why add an unused argument here and on line 145?

djangoproject/static/js/djangoproject.js
```diff
@@ -130,25 +134,48 @@ document.querySelectorAll('.btn-clipboard').forEach(function (el) {
...
 
-    const text = this.parentElement.nextElementSibling.textContent.trim();
+    let text;
+    const console_section = this.closest('.console-block > section');
+
+    if (console_section) {
+      // Console tabs: extract text excluding prompts (.gp) and output (.go)
+      const pre_el = console_section.querySelector('.highlight pre');
+      text = '';
+      if (pre_el) {
```
adamzap: In what case is there not a `<pre>` element?

djangoproject/static/js/djangoproject.js
```diff
@@ -130,25 +134,48 @@ document.querySelectorAll('.btn-clipboard').forEach(function (el) {
...
+        pre_el.childNodes.forEach(function (node) {
```
adamzap: Do you think it would be more elegant to grab the text content like the old code was doing and just left trim the prompt?

```javascript
(...).textContent.trim().replace(/^\$ /, '')
```

djangoproject/static/js/djangoproject.js
aadeina: Good catch! Just to clarify — I actually removed the unused `el` parameter here, since it wasn't being used (the functions access `success_el` via closure instead). Happy to keep it if you prefer consistency with other patterns in the codebase.

djangoproject/static/js/djangoproject.js
aadeina: You're right — since Pygments always generates the `<pre>` element, this check isn't strictly necessary. 
I'll remove it to keep the code cleaner. Thanks for pointing this out!

djangoproject/static/js/djangoproject.js
aadeina: Great question! I went with the node-based approach mainly because:

1. It also handles the Windows tab (which uses `...>` prompts)
2. It excludes command output marked with `.go` (in case any examples show output)

That said, if the console blocks only ever show simple `$ ` prompts without output, a regex would definitely be simpler. I'm happy to switch to:

```js
text = pre_el.textContent.trim().replace(/^(\$ |\.\.\.>)/gm, '');

Comment (aadeina) [reply djangoproject/static/js/djangoproject.js]: Let me know which approach you'd prefer!

File: djangoproject/static/js/djangoproject.js (hunk 1 above)
Comment (adamzap) [reply djangoproject/static/js/djangoproject.js:147]: Oh, thanks! Sorry for my misunderstanding here.

**Comment by aadeina** [comment]
Updated per feedback — removed the `if` check. Thanks! Ready for another look when you have time.

File: djangoproject/static/js/djangoproject.js (hunk 3 above)
Comment (adamzap) [reply djangoproject/static/js/djangoproject.js]: Let's see how the regex looks for now. I should have more time to look at this soon. Thanks!

**Comment by adamzap** [comment]
Thank you. In the future, could you wait to squash until the PR is ready? That way I can review new commits in isolation.

**Comment by aadeina** [comment]
Understood, apologies for the premature squash. I’ll keep commits separate going forward and appreciate the guidance on best practices.

File: djangoproject/static/js/djangoproject.js (hunk 3 above)
Comment (aadeina) [reply djangoproject/static/js/djangoproject.js]: I've updated to use the regex approach. Thanks!

File: djangoproject/static/js/djangoproject.js [hunk 4]
```diff
@@ -102,24 +102,39 @@ window.addEventListener('keydown', function (e) {
...
+    }
   });
 })();
 
-// Attach copy functionality to dynamically-created buttons
+// Attach copy functionality to clipboard buttons
 document.querySelectorAll('.btn-clipboard').forEach(function (el) {
   el.addEventListener('click', function () {
+    // Remove any existing success message first
+    const existing = this.querySelector('.clipboard-success');
```
adamzap: Could we call this `success_el` and drop the comment? I guess we'd need a `let`.

djangoproject/static/js/djangoproject.js
```diff
@@ -130,25 +145,35 @@ document.querySelectorAll('.btn-clipboard').forEach(function (el) {
...
     }
 
-    const text = this.parentElement.nextElementSibling.textContent.trim();
+    let text;
+    const console_section = this.closest('.console-block > section');
+
+    if (console_section) {
+      // Console tabs: extract text excluding prompts
+      const pre_el = console_section.querySelector('.highlight pre');
+      text = pre_el.textContent.replace(/^\$ |^\.\.\.\\>/gm, '');
```
adamzap: Could we set the regex to a variable and drop the comment here and in the `else` below? Maybe `prompt_regex`?

djangoproject/static/js/djangoproject.js
```diff
@@ -130,25 +145,35 @@ document.querySelectorAll('.btn-clipboard').forEach(function (el) {
...
+    } else {
+      // Code snippets: get text from next sibling
+      text = this.parentElement.nextElementSibling.textContent;
+    }
 
-    navigator.clipboard.writeText(text).then(on_success, on_error);
+    navigator.clipboard.writeText(text.trim()).then(on_success, on_error);
```
adamzap, line 174: Why is this `trim` call needed?

djangoproject/static/js/djangoproject.js
adamzap: On second thought, I think `existing_el` or `existing_success_el` would be fine if you don't want to use the same variable.

djangoproject/static/js/djangoproject.js
aadeina: The `.trim()` prevents a trailing newline, so the command doesn't execute immediately upon paste.

djangoproject/static/js/djangoproject.js
aadeina: Good call. Extracted this to `const prompt_regex`.

djangoproject/static/js/djangoproject.js
aadeina: Agreed. Renamed to `existing_el` for consistency.

aadeina:
**Visual comparison (1px layout shift fix):**

 **Before:**

https://github.com/user-attachments/assets/e99c2994-09da-4196-9b9a-836e253bc0fe


 **After:**
 

https://github.com/user-attachments/assets/4a9bb731-706e-43dc-9cf9-1e36999ab5f5



 Please let me know if there's anything else you'd like me to adjust!

adamzap:
Thanks for the changes. I'll try to get to this by end of tomorrow!

adamzap (approved):
👍 Looks great!

Thanks for your patience here. I will ask someone on the team to do a CSS review.

aadeina:
Thanks so much for the thorough reviews and guidance, @adamzap! I really appreciate the time you took to help refine this. Looking forward to the CSS review.

djangoproject/scss/_console-tabs.scss
```diff
@@ -34,9 +34,48 @@
     >section {
         display: none;
         text-align: left;
+        position: relative;
```
sabderemane: Why add `position: relative` here?

sabderemane (commented):
Thank you for working on this! I've a question on the CSS to understand what is the purpose of the change, otherwise looks good to me!

djangoproject/scss/_console-tabs.scss
aadeina: Ah, good eye! Leftover from when the button was absolutely positioned — forgot to remove it after switching to flexbox. Fixed now, thanks for the review!

SaptakS (approved):
Thank you @aadeina for working on this. Looks great! Thanks everyone for reviewing the PR. I am going to merge this now.

aadeina:
Thanks for merging @SaptakS! Really appreciate everyone's time on the reviews.
This should close #1276 and #2399 (heads up if GitHub didn't catch them automatically).

noorzaibchohan:
Great. Loved to see this. LGTM because team approved this after verifying carefully !

PR #2434, opened 2026-01-05
Set cooldown in dependabot

p-r-a-v-i-n:
following up from https://github.com/django/djangoproject.com/pull/2427#pullrequestreview-3625122990

 Ref for cooldown: https://docs.zizmor.sh/audits/#dependabot-cooldown

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,6 +7,8 @@ updates:
           day: "monday"
           time: "12:00"
           timezone: "Europe/Oslo" # hipp hipp hurra
+      cooldown:
+          default-days: 7
 
     - package-ecosystem: github-actions
       directory: /
@@ -15,3 +17,5 @@ updates:
           day: "monday"
           time: "12:00"
           timezone: "Europe/Oslo" # hipp hipp hurra
+      cooldown:
+          default-days: 7


ulgens:
+1 on the intention, but I think it's better to have zizmor in place and let it allow to enforce this rule, instead of adding it manually.

p-r-a-v-i-n:
> +1 on the intention, but I think it's better to have zizmor in place and let it allow to enforce this rule, instead of adding it manually.

Thanks @ulgens , this would be new learning for me.

adamzap:
@p-r-a-v-i-n Could you rebase this?

ulgens:
@adamzap How do you feel about zizmor?

adamzap:
@ulgens I haven't looked into it yet. I want to get this PR into a merge-able state to see if we want to move forward with it. Do you think we should close this?

ulgens:
I think it can be merged as is - zizmor decision is not a blocker. My thinking was that this is a part of the security improvements handled by an open source tool and having the tool in place could be more helpful in the long run.

sarahboyce (approved):
Thank you!

ulgens (approved):

PR #2439, opened 2026-01-07
Run Sphinx builds in isolated subprocesses

kehach07:
Sphinx keeps global state and is not designed to be run multiple times in the same Python process.

This change runs each documentation build in a separate subprocess and isolates doctrees per language and builder, preventing translation state from leaking between builds and fixing mixed-language labels.

### DIFF
--- a/docs/management/commands/update_docs.py
+++ b/docs/management/commands/update_docs.py
@@ -4,6 +4,7 @@ app.
 """
 
 import json
+
 # import multiprocessing
 import os
 import shutil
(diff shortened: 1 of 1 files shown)


ulgens:
Would you mind creating an issue and describing the existing problem?

> preventing translation state from leaking between builds and fixing mixed-language labels.

Are we currently suffering from this issue? If so, I think it's better we document it how does it happen and what is the result it's creating.

Issue #2416, opened 2025-12-23
6.0 English docs are using Swedish Sphinx labels


medmunds:
(I'm assuming this is related to the docs.djangoproject.com build and not the Django docs source. Apologies if this ends up belonging in code.djangoproject.com instead.)

Some—but not all—of the `versionchanged` (and maybe `versionadded`?) labels in the *English* docs are showing up in *Swedish.* Example from https://docs.djangoproject.com/en/6.0/topics/email/:

> **Changed in Django 6.0:**
> Support for `MIMEPart` attachments was added.
>
> **Ersatt sedan version 6.0:**
Support for `email.mime.base.MIMEBase` attachments is deprecated. Use `MIMEPart` instead.

Here's a screenshot:
<img width="498" height="204" alt="Screenshot from docs.djangoproject.com/en/6.0/topics/email/ showing text quoted above" src="https://github.com/user-attachments/assets/6a49c97f-d778-47c2-882b-34d93970f328" />

(Search for "Ersatt" in the page to find several other examples.)

DhanashreePetare:
Greetings @medmunds ..I am interested to work on the issue. Can you please assign this to me. ThankYou

adamzap:
@DhanashreePetare As long as there is no pull request for this issue, you are free to attempt a fix. No need to wait for the issue to be assigned to you.

sarahboyce:
I confirmed this issue when this was reported. This often goes away on it's own with a docs rebuild (I believe this is better now).
That we have sparidic mixed translations is still an issue. This was also reported in https://github.com/django/djangoproject.com/issues/1973 which was not really fixed but rather "improved"

adamzap:
@sarahboyce Hmm, you're right. The example posted above is now correct in production.

medmunds:
What's odd is that the exact same Sphinx directive (versionchanged) used twice in a row is ending up with different translations. That feels like some sort of *incremental* parsing/caching problem.

The [`_clean_up_global_state()` workaround](https://github.com/django/djangoproject.com/blob/38220f29619ebe8a706141eb792ad09d6d7a6579/docs/management/commands/update_docs.py#L223-L224) should probably be moved to a `finally` clause. But I'm not sure that would explain this.

[...]

PR #2441, opened 2026-01-07
Run Sphinx builds in isolated subprocesses

kehach07:
Fixed issue :- Docs intermittently mix translations of Sphinx labels #2416

Run each Sphinx build in a separate subprocess to avoid global state leakage
between documentation builds.

### DIFF
--- a/docs/management/commands/update_docs.py
+++ b/docs/management/commands/update_docs.py
@@ -178,6 +178,12 @@ class Command(BaseCommand):
             builders = self.default_builders[:] + ["gettext"]
         else:
             builders = self.default_builders
+        if release.lang != settings.DEFAULT_LANGUAGE_CODE:
+            os.environ["LANGUAGE"] = release.lang
+            os.environ["LC_ALL"] = release.lang
+        else:
+            os.environ.pop("LANGUAGE", None)
+            os.environ.pop("LC_ALL", None)
 
         #
         # Use Sphinx to build the release docs into JSON and HTML documents.
@@ -264,9 +270,6 @@ class Command(BaseCommand):
         builder,
         language,
     ):
-        env = os.environ.copy()
-        env["SPHINXOPTS"] = f"-D language={language}"
-
         subprocess.check_call(
             [
                 sys.executable,
@@ -274,14 +277,15 @@ class Command(BaseCommand):
                 "sphinx",
                 "-b",
                 builder,
+                "-D",
+                f"language={language}",
                 "-c",
                 str(source_dir),
                 "-d",
                 str(doctreedir),
                 str(source_dir),
                 str(build_dir),
-            ],
-            env=env,
+            ]
         )
 
     def update_git(self, url, destdir, changed_dir="."):
--- a/docs/management/commands/update_docs.py
+++ b/docs/management/commands/update_docs.py
@@ -289,7 +289,6 @@ class Command(BaseCommand):
             env=env,
         )
 
-
     def update_git(self, url, destdir, changed_dir="."):
         """
         Update a source checkout and return True if any docs were changed,
--- a/docs/management/commands/update_docs.py
+++ b/docs/management/commands/update_docs.py
@@ -18,7 +18,9 @@ from django.core.management import BaseCommand, call_command
 from django.db.models import Q
 from django.utils.translation import to_locale
 from sphinx.config import Config
+
 from ...models import DocumentRelease
+
 
 class Command(BaseCommand):
     def add_arguments(self, parser):
@@ -258,7 +260,7 @@ class Command(BaseCommand):
         json_built_dir = parent_build_dir / "_built" / "json"
         documents = gen_decoded_documents(json_built_dir)
         release.sync_to_db(documents)
-        
+
     def run_sphinx_build(
         self,
         *,
--- a/docs/management/commands/update_docs.py
+++ b/docs/management/commands/update_docs.py
@@ -265,7 +265,7 @@ class Command(BaseCommand):
         json_built_dir = parent_build_dir / "_built" / "json"
         documents = gen_decoded_documents(json_built_dir)
         release.sync_to_db(documents)
-        def run_sphinx_build(self,*,source_dir,build_dir,doctreedir,builder,language,extensions,):
+    def run_sphinx_build(self,*,source_dir,build_dir,doctreedir,builder,language,extensions,):
             env = os.environ.copy()
             env["SPHINXOPTS"] = " ".join(
                 [
--- a/docs/management/commands/update_docs.py
+++ b/docs/management/commands/update_docs.py
@@ -17,8 +17,6 @@ from django.conf import settings
 from django.core.management import BaseCommand, call_command
 from django.db.models import Q
 from django.utils.translation import to_locale
-from sphinx.config import Config
-
 from ...models import DocumentRelease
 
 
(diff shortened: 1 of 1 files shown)


pauloxnet:
I would suggest removing old lines of code instead of commenting.

kehach07:
> I would suggest removing old lines of code instead of commenting.

thank you for the feedback, i have implemented the suggestion.

medmunds:
This seems to me like a safer way to use Sphinx, and removing the monkeypatching and undocumented Sphinx internals feels much cleaner.

One important test would be whether this generates the same built docs as before. (I'm not set up to run djangoproject.com locally, and it doesn't seem like the PR checks deploy a preview site.)

FWIW, Copilot had some concerns:

> Summary — will the built docs match previous builds (aside from the bug being fixed)?
>
> Not with the code as written. The subprocess approach is the right direction (it properly isolates Sphinx/docutils global state), but the current way extensions are passed to sphinx-build is likely to change behaviour and break loading of extensions, so output may differ. Other behavioral changes (single-threaded build, doctree location) are less likely to change rendered content, but could affect build speed and doctree reuse.

The point about extensions seems valid: they need to be full dotted paths in a command line arg, and there are potential shell quoting issues. But I question some of its other concerns: e.g., preventing doctree reuse was intentional. (Also, the AI mixing British and American spelling in the same paragraph doesn't inspire confidence 😄 .)

Copilot also pointed out logging behavior would change.

[To clarify: I'm not a maintainer on this project, just a human who uses docs.djangoproject.com and reported a bug.]

kehach07:
> This seems to me like a safer way to use Sphinx, and removing the monkeypatching and undocumented Sphinx internals feels much cleaner.
> 
> One important test would be whether this generates the same built docs as before. (I'm not set up to run djangoproject.com locally, and it doesn't seem like the PR checks deploy a preview site.)
> 
> FWIW, Copilot had some concerns:
> 
> > Summary — will the built docs match previous builds (aside from the bug being fixed)?
> > Not with the code as written. The subprocess approach is the right direction (it properly isolates Sphinx/docutils global state), but the current way extensions are passed to sphinx-build is likely to change behaviour and break loading of extensions, so output may differ. Other behavioral changes (single-threaded build, doctree location) are less likely to change rendered content, but could affect build speed and doctree reuse.
> 
> The point about extensions seems valid: they need to be full dotted paths in a command line arg, and there are potential shell quoting issues. But I question some of its other concerns: e.g., preventing doctree reuse was intentional. (Also, the AI mixing British and American spelling in the same paragraph doesn't inspire confidence 😄 .)
> 
> Copilot also pointed out logging behavior would change.
> 
> [To clarify: I'm not a maintainer on this project, just a human who uses docs.djangoproject.com and reported a bug.]

Thanks for the detailed review!

I’ve updated the PR and the implementation to avoid passing extensions via the command line and now rely entirely on conf.py, so extension loading and rendered output should match previous builds (aside from the bug fix).

The subprocess isolation is intentional to avoid shared Sphinx/docutils state; changes like separate doctrees are expected but shouldn’t affect rendered content. Logging differences are noted, but behaviorally this should now be much closer to the existing builds.

sarahboyce:
> the current way extensions are passed to sphinx-build is likely to change behaviour and break loading of extensions

:point_up: these changes currently mean the custom extension is no longer used as it roughly reverts the use added in  https://github.com/django/djangoproject.com/pull/1947

How to test this works locally:
- Build the docs (`python -m manage update_docs`)
- Search for "select"
- You should see code reference search results such as `QuerySet.select_for_update` (currently with these changes, these results are lost)

kehach07:
> ☝️ these changes currently mean the custom extension is no longer used as it roughly reverts the use added in #1947
> 
> How to test this works locally:
> 
> * Build the docs (`python -m manage update_docs`)
> * Search for "select"
> * You should see code reference search results such as `QuerySet.select_for_update` (currently with these changes, these results are lost)

Thanks for the pointer @sarahboyce . I’ve updated the PR to rely entirely on conf.py for extension loading (no overrides), so the custom extension is used again.
I built the docs locally with python manage.py update_docs and verified that searching for “select” returns code reference results like QuerySet.select_for_update, so the search is working as expected now.

sarahboyce:
I get the following error when running `python manage.py update_docs --force` on this branch:
```
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/usr/local/lib/python3.12/site-packages/sphinx/__main__.py", line 7, in <module>
    raise SystemExit(main(sys.argv[1:]))
                     ^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/sphinx/cmd/build.py", line 546, in main
    locale.setlocale(locale.LC_ALL, '')
  File "/usr/local/lib/python3.12/locale.py", line 615, in setlocale
    return _setlocale(category, locale)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
locale.Error: unsupported locale setting
sphinx-build failed (release=fr/5.2, builder=json): Command '['/usr/local/bin/python', '-m', 'sphinx', '-b', 'json', '-D', 'language=fr', '-c', 'data/djangodocs/sources/5.2/docs', '-d', 'data/djangodocs/fr/5.2/_doctrees/fr/json', 'data/djangodocs/sources/5.2/docs', 'data/djangodocs/fr/5.2/_build/json']' returned non-zero exit status 1. 
```

Also I still have the same issue on this branch. I think I didn't explain it well enough.
On the main branch, this is what the results look like when searching for select:

Screenshot from 2026-01-25 10-27-54

This is what the results look like on this branch:

Screenshot from 2026-01-25 10-12-25

`QuerySet.select_for_update` etc. is still missing

medmunds:
(The Swedish labels are back in the English docs site today.)

If I were trying to verify the proposed change (or get an AI to help me with it), I might try to check the generated docs before and after the change are exactly the same. Something roughly like:

```shell
# 1. Start clean:
rm -rf djangodocs

# 2. Build the docs without the change, and keep a copy:
git checkout main
python manage.py update_docs --force --language en
mv djangodocs django-docs-before

# 3. Build the docs with the change:
git checkout fix-sphinx-subprocess-clean
python manage.py update_docs --force --language en

# 4. Compare before and after:
diff --recursive django-docs-before django-docs
```

If there are any differences (other than timestamps or the bug being fixed here), that's a problem.

(The commands above build only the default English language to speed development. Once everything seems to be working, it might also be good to compare builds for all languages by repeating without `--language en`.)

kehach07:
> (The Swedish labels are back in the English docs site today.)
> 
> If I were trying to verify the proposed change (or get an AI to help me with it), I might try to check the generated docs before and after the change are exactly the same. Something roughly like:
> 
> ```shell
> # 1. Start clean:
> rm -rf djangodocs
> 
> # 2. Build the docs without the change, and keep a copy:
> git checkout main
> python manage.py update_docs --force --language en
> mv djangodocs django-docs-before
> 
> # 3. Build the docs with the change:
> git checkout fix-sphinx-subprocess-clean
> python manage.py update_docs --force --language en
> 
> # 4. Compare before and after:
> diff --recursive django-docs-before django-docs
> ```
> 
> If there are any differences (other than timestamps or the bug being fixed here), that's a problem.
> 
> (The commands above build only the default English language to speed development. Once everything seems to be working, it might also be good to compare builds for all languages by repeating without `--language en`.)

Thank you @sarahboyce and @medmunds for checking and for the detailed verification steps — that’s very helpful.

I agree that doing a clean before/after comparison of the generated docs is a good way to validate that this change doesn’t introduce unintended differences.

I’ll reproduce the comparison you suggested (starting from a clean state and comparing main vs this branch) and check whether the Swedish labels still appear in the English build. I’ll also verify whether the issue is intermittent and update the PR with my findings.

Thanks again for taking the time to outline a clear validation approach.

adamzap:
@kehach07 I'm going to close this due to inactivity. Please reopen it if you intend to finish the work here. Thanks!

PR #2442, opened 2026-01-07
Restored search hotkey for desktop. (alternative)

sarahboyce:
Fixes https://github.com/django/djangoproject.com/issues/2349

I also added playwright to have a way to test JavaScript (might be controversial and can remove if wanted)
This is an alternative to https://github.com/django/djangoproject.com/pull/2355. Here, only one search bar exists in the page


djangoproject/templates/includes/header.html
```diff
@@ -46,13 +40,11 @@
...
-          {% search_form %}
-        </li>
-        <li>
-          {% include "includes/toggle_theme.html" %}
-        </li>
       </ul>
     </nav>
+    <div class="header-tools">
+      {% search_form %}
+      {% include "includes/toggle_theme.html" %}
```
adamzap, line 47: I suppose we could remove this include now, but it's probably out of scope here.

sarahboyce:
> We should confirm that this design change on wide viewports is intended [...]
> I think it's a nice improvement!

It wasn't intentional, I also think I prefer it to the current state. I know ülgen pointed out in the slack group that prior to the previous navbar change the navbar was centered on wide viewpoints. So perhaps we can chose what we want and make some adjustments 👍

adamzap:
@sarahboyce We discussed this PR in the WG meeting today. Could you move the Playwright parts of this to another PR so that we can iterate on front-end testing separately?

sarahboyce:
> Could you move the Playwright parts of this to another PR so that we can iterate on front-end testing separately?

Moved to https://github.com/django/djangoproject.com/pull/2444

PR #2444, opened 2026-01-09
Added playwright tests.

sarahboyce:
Note that these tests are expected to fail as these are regression tests for the search hotkey issue. These should pass once a solution (such as https://github.com/django/djangoproject.com/pull/2442) is merged

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,6 +30,10 @@ COPY ./requirements ./requirements
 RUN apt-get update \
     && apt-get install --assume-yes --no-install-recommends ${BUILD_DEPENDENCIES} \
     && python3 -m pip install --no-cache-dir -r ${REQ_FILE} \
+    && if [ "${REQ_FILE}" = "requirements/tests.txt" ]; then \
+        echo "Installing Playwright browsers..."; \
+        python3 -m playwright install --with-deps chromium; \
+    fi \
     && apt-get purge --assume-yes --auto-remove ${BUILD_DEPENDENCIES} \
     && apt-get distclean
 
--- a/djangoproject/tests.py
+++ b/djangoproject/tests.py
@@ -1,12 +1,15 @@
+import os
 from http import HTTPStatus
 from io import StringIO
 
 from django.conf import settings
+from django.contrib.staticfiles.testing import StaticLiveServerTestCase
 from django.core.management import call_command
 from django.test import TestCase
 from django.urls import NoReverseMatch, get_resolver
 from django.utils.translation import activate, gettext as _
 from django_hosts.resolvers import reverse
+from playwright.sync_api import expect, sync_playwright
 
 from docs.models import DocumentRelease, Release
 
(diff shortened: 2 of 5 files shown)


djangoproject/tests.py
```diff
@@ -211,3 +214,52 @@ def test_single_h1_per_page(self):
...
+                is_focused = page.evaluate("document.activeElement.id === 'id_q'")
+                self.assertTrue(is_focused)
+                page.close()
+
+    def test_search_placeholder_mac_mode(self):
+        page = self.browser.new_page(user_agent=self.mac_user_agent)
+        page.goto(self.live_server_url)
+
+        desktop_search_bar = page.locator("#id_q")
+        expect(desktop_search_bar).to_have_attribute("placeholder", "Search (⌘\u200aK)")
```
adamzap, line 282: Should focus also be tested here?

Dockerfile
adamzap, line 1: Could someone that uses Docker review the change to this file?

djangoproject/tests.py
```diff
@@ -211,3 +214,52 @@ def test_single_h1_per_page(self):
...
+        cls.browser = cls.playwright.chromium.launch()
+
+    @classmethod
+    def tearDownClass(cls):
+        super().tearDownClass()
+        cls.browser.close()
+        cls.playwright.stop()
+
+    def setUp(self):
+        self.setUpTestData()
```
adamzap, line 256: Did you mean `self.setUp()` here?

djangoproject/tests.py
```diff
@@ -211,3 +214,52 @@ def test_single_h1_per_page(self):
...
+        self.mac_user_agent = "Mozilla/5.0 (Macintosh) AppleWebKit"
+        self.windows_user_agent = "Mozilla/5.0 (Windows NT 10.0)"
+        self.mobile_linux_user_agent = "Mozilla/5.0 (Linux; Android 10; Mobile)"
```
adamzap: Could these be class attributes or go in `setUpClass`?

djangoproject/tests.py
sarahboyce: I don't think so because this tests this code: https://github.com/django/djangoproject.com/blob/566d8ab8a66be96b5be275930c452fda7002bf3a/djangoproject/static/js/djangoproject.js#L68-L81
This is only to do with the search bar text

The focus is covered in the other test

djangoproject/tests.py
sarahboyce: Moved into `setUpClass` :+1:

djangoproject/tests.py
sarahboyce: Ah no, we need `setUpTestData` to be called as it isn't called from `StaticLiveServerTestCase`

djangoproject/tests.py
adamzap: You're right, thanks!

djangoproject/tests.py
adamzap: Oh, ok, for `ReleaseMixin`? I think I understand now. It feels odd to call `setUpTestData` since it's intended to be used to set up data for the whole test case. It seems safe though.

adamzap (approved):
👍 Looks great!

Dockerfile
```diff
@@ -31,6 +31,10 @@ COPY ./requirements ./requirements
 RUN apt-get update \
     && apt-get install --assume-yes --no-install-recommends ${BUILD_DEPENDENCIES} \
     && python3 -m pip install --no-cache-dir -r ${REQ_FILE} \
+    && if [ "${REQ_FILE}" = "requirements/tests.txt" ]; then \
+        echo "Installing Playwright browsers..."; \
+        playwright install --with-deps; \
+    fi \
     && apt-get purge --assume-yes --auto-remove ${BUILD_DEPENDENCIES} \
```
CuriousLearner, line 37: We may choose to install and test on chromium if not all broswers are necessary:

```
python3 -m playwright install --with-deps chromium; \
```

CuriousLearner: This will reduce the docker image size as well.

pauloxnet: I would test at least for chromium and firefox to at least cover the 2 more used open source browser engine.

adamzap:
@sarahboyce Do you want to rebase this branch and address the comment about browser versions?

Dockerfile
sarahboyce: I have updated to only install chromium
The current design does not allow for testing on multiple browsers, I think we can add that as a followup ticket if we are encountering bugs where this would be useful

pauloxnet: I think testing in more than one browser would be better for diversification. But we can do that in a follow-up issue. Can you create at list an issue foradding Firefox in the future?

pauloxnet (approved):

Dockerfile
sarahboyce: https://github.com/django/djangoproject.com/issues/2613 done :+1:

PR #2447, opened 2026-01-10
Fixed 405 error when deleting community feeds

### LINKED ISSUE #1646: Cannot delete own feed from Django Community Blog
When a logged-in (non-superuser) tries to delete one of their feeds from https://www.djangoproject.com/community/mine/ after the confirmation request, they are met with a 405 response and the feed doesn't get deleted

vidhiii1711:
Fixes a 405 error when non-superusers attempt to delete their own community feeds.
The delete form submits a POST request to a URL without a trailing slash, which triggers a redirect when APPEND_SLASH=True, converting the request to GET. Adding the trailing slash to the URL pattern prevents the redirect and allows the feed to be deleted correctly.

Fixes #1646 

Letme know if there's any other changes needed !!

### DIFF
--- a/aggregator/tests.py
+++ b/aggregator/tests.py
@@ -154,29 +154,28 @@ class AggregatorTests(TestCase):
             ["Approved long URL Item", "Approved Item"],
             transform=attrgetter("title"),
         )
+
     def test_logged_in_user_can_delete_own_feed(self):
-     user = User.objects.create_user(
-        username="deleter",
-        password="pass123",
-     )
+        user = User.objects.create_user(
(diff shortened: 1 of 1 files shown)


ulgens:
I'll recommend adding a test. LGTM otherwise.

vidhiii1711:
Thanks for the review! I’ve added a test covering feed deletion via POST and pushed the update. Please let me know if you’d like any changes.

vidhiii1711:
All changes have been addressed. Please let me know if anything else is needed.

PR #2454, opened 2026-01-13
Rearrange Tasks in Ecosystem page

frankwiles:
- Created new Tasks section and moved the Task app references there
- Spotted two missing Testing related libraries so added them as well


djangoproject/templates/aggregator/ecosystem.html
```diff
@@ -324,6 +319,17 @@ <h3 id="templates">
     </li>
   </ul>
 
+  <h3 id="templates">
+    Tasks
+    <a class="headerlink" href="#templates" aria-label="Link to this heading" aria-describedby="community-ecosystem">¶</a>
```
tim-schilling:
```suggestion
  <h3 id="tasks">
    Tasks
    <a class="headerlink" href="#tasks" aria-label="Link to this heading" aria-describedby="community-ecosystem">¶</a>
```

djangoproject/templates/aggregator/ecosystem.html
```diff
@@ -324,6 +319,17 @@ <h3 id="templates">
     </li>
   </ul>
 
+  <h3 id="templates">
+    Tasks
```
tim-schilling: Minor / potentially nitpick: I think we may benefit from a more descriptive title for this section. Tasks works for Django because we know what we're referring to. However, does every other web framework use this same term to refer to this type of functionality?

I may be bikeshedding here though 🤷

djangoproject/templates/aggregator/ecosystem.html
frankwiles: Doh good catch!

djangoproject/templates/aggregator/ecosystem.html
frankwiles: I think the only other real option is "Async Tasks" which runs the risk of getting confused with `asyncio` bits.

tim-schilling: That's fair. I'll mark this as resolved.

LilyFirefly: We could use "Background Tasks".

frankwiles: ooh good idea.  Changing.

SaptakS (approved):

PR #2456, opened 2026-01-15
Disable cache 404

vidhiii1711:
This PR fixes the issue where 404 responses were being cached, which can result in stale or incorrect responses being served. The change ensures that 404 responses are explicitly excluded from caching.

Updated the caching logic to prevent 404 (Not Found) responses from being cached.
Added a regression test to ensure 404 responses are not cached going forward.
Verified that the change does not affect caching of valid responses.
Added a test covering the 404 caching behavior

Note: I was unable to run the test suite locally due to Docker/PostgreSQL setup issues on Windows.

### DIFF
--- a/aggregator/tests.py
+++ b/aggregator/tests.py
@@ -154,29 +154,28 @@ class AggregatorTests(TestCase):
             ["Approved long URL Item", "Approved Item"],
             transform=attrgetter("title"),
         )
+
     def test_logged_in_user_can_delete_own_feed(self):
-     user = User.objects.create_user(
-        username="deleter",
-        password="pass123",
-     )
+        user = User.objects.create_user(
+            username="deleter",
+            password="pass123",
+        )
 
-     feed = models.Feed.objects.create(
-        title="My Feed",
-        feed_url="https://example.com/rss/",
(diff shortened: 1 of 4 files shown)


aggregator/urls.py
```diff
@@ -19,5 +19,5 @@
     path("<feed_type_slug>/", views.FeedListView.as_view(), name="community-feed-list"),
     path("add/<feed_type_slug>/", views.add_feed, name="community-add-feed"),
     path("edit/<int:feed_id>/", views.edit_feed, name="community-edit-feed"),
-    path("delete/<int:feed_id>", views.delete_feed, name="community-delete-feed"),
+    path("delete/<int:feed_id>/", views.delete_feed, name="community-delete-feed"),
```
adamzap, line 22: This change and the test above seems unrelated to the issue you are trying to fix.

vidhiii1711: Thanks for pointing that out — you’re right.
I accidentally reused the same branch and included commits from a previous, unrelated issue (community feed deletion).
I closed this PR and opened a new one with only the changes related to preventing caching of 404 responses. Apologies for the confusion.

PR #2458, opened 2026-01-15
Fix donation percentage rounding and add on-track messaging

### LINKED ISSUE #2455: Tweaks to donation percentage communication
1. The big percentage number has standard rounding, hence 2.8% is rounded to 3%. However, we should always round down (`floor()`) so that we do not celebrate hitting 100% before we actually reach 100%
2. It would be nice to add a form of judgement as to whether we are on track. This could be updating:
- $8,498.64 donated of US $300,000.00 goal for 2026

  To say:

- $8,498.64 donated of US $300,000.00 goal for 2026. To reach our goal, we should have raised $X by this point

  Where $X is the donation target multiplied by the percentage progress through the year

aadeina:
Fixes #2455

### Summary

  - **Floor rounding:** Use `floor()` so 99.9% shows as 99%, not 100%
  - **On-track message:** Show expected donations based on day-of-year progress
  - **Currency format:** Updated to "$X donated of a $Y USD goal"
  - **Accessibility:** Added `<progress>` element for screen readers

### DIFF
--- a/fundraising/tests/test_templatetags.py
+++ b/fundraising/tests/test_templatetags.py
@@ -1,6 +1,8 @@
 from datetime import date, datetime, timedelta
 from decimal import Decimal
+from unittest.mock import patch
 
+import time_machine
 from django.test import TestCase
 from django.utils.crypto import get_random_string
 
@@ -15,10 +17,30 @@ from ..models import (
     Payment,
 )
 from ..templatetags.fundraising_extras import (
+    as_percentage,
     display_django_heroes,
     donation_form_with_heart,
     top_corporate_members,
 )
+
+
+class TestAsPercentage(TestCase):
+    def test_floor_rounding(self):
+        with self.subTest(case="2.8% should show as 2%, not 3%"):
+            self.assertEqual(as_percentage(Decimal("8400"), Decimal("300000")), "2")
+        with self.subTest(case="99.9% should show as 99%, not 100%"):
+            self.assertEqual(as_percentage(Decimal("299700"), Decimal("300000")), "99")
+        with self.subTest(case="Exact percentage should remain the same"):
+            self.assertEqual(as_percentage(Decimal("150000"), Decimal("300000")), "50")
+
+    def test_zero_and_none_values(self):
+        self.assertEqual(as_percentage(None, Decimal("300000")), "0")
+        self.assertEqual(as_percentage(Decimal("0"), Decimal("300000")), "0")
+        self.assertEqual(as_percentage(Decimal("100"), None), "0")
+        self.assertEqual(as_percentage(Decimal("100"), Decimal("0")), "0")
+
+    def test_over_100_percent(self):
+        self.assertEqual(as_percentage(Decimal("350000"), Decimal("300000")), "116")
 
 
 class TestDonationFormWithHeart(TestCase):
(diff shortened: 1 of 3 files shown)


sarahboyce:
Thank you for the PR! This looks like what I was expecting and love that you've written tests 👏

I think it would make sense to test the `expected_amount` further, perhaps using a combination of `subTest` and `freezegun` (adding this as a test dependency)

For transparency, I go on holiday today and won't get a chance to review in detail for the next two weeks 👍

pauloxnet:
> I think it would make sense to test the `expected_amount` further, perhaps using a combination of `subTest` and `freezegun` (adding this as a test dependency)

Luckily we migrated to `time-machine`, but I like the idea :+1: 

In addition can you attach a screenshot to show us how it looks?

I'm interested in see how the progress element looks.

aadeina:
Thank you @sarahboyce and @pauloxnet for the thoughtful feedback!

  I've addressed your suggestions:

  ✅ Added comprehensive tests for `expected_amount` using `time-machine` and `subTest`

  ## Screenshots

image



  **Progress element (normally hidden, shown here for demo):**

image


  ## About the Progress Element

  The `<progress>` element has `class="visually-hidden"` — it's intentionally hidden from sighted users but accessible to screen readers. Screen readers announce: "Donation progress: X of 300,000".

  ---

  @sarahboyce enjoy your holiday! No rush on the review. 🏖️

  Let me know if you'd like any adjustments!

pauloxnet:
> Thank you @sarahboyce and @pauloxnet for the thoughtful feedback!
> 
> I've addressed your suggestions:
> 
> ✅ Added comprehensive tests for `expected_amount` using `time-machine` and `subTest`
> 
> ## Screenshots
> image
> **Progress element (normally hidden, shown here for demo):**
> 
> image
> ## About the Progress Element
> The `<progress>` element has `class="visually-hidden"` — it's intentionally hidden from sighted users but accessible to screen readers. Screen readers announce: "Donation progress: X of 300,000".
> 
> @sarahboyce enjoy your holiday! No rush on the review. 🏖️
> 
> Let me know if you'd like any adjustments!

Thanks for the update. I would suggest to set the color of the progress bar similar to a green taken from CSS that can have enough contrast with the background to be accessible, or better use one of the "red" similar to the vectorial heart. I'm sure Sarah or other in the accessibility team can help you in choosing the right color to use in the progresss bar better than me.

aadeina:
Thank you for the suggestion, @pauloxnet! I've styled the progress bar using Amaranth (#d9195c) - the most frequently used color in the heart.

 This provides:
 - Visual consistency with the fundraising heart
 - Good contrast against the light gray background for accessibility

 Screenshot:
image

 Please let me know if you'd prefer a different shade from the heart palette, or if the accessibility team has any recommendations!

pauloxnet:
@aadeina very good job 👍

aadeina:
Hi @sarahboyce, hope the holiday was great! Just a friendly ping on this whenever you have a moment. All feedback has been addressed. Thanks!

djangoproject/scss/_style.scss
```diff
@@ -3305,6 +3305,30 @@ hr {
             margin-top: 0;
             @include font-size(28);
         }
+
```
sarahboyce: I actually think we shuld remove the `<progress>` element and styling around this.
The `<progress>` element (https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/progress) should be used for showing progress of a task such as a file download or a form completion. Using this for a metric, which is quite static and does not require live page progress announcements, is not suitable imo.

fundraising/tests/test_templatetags.py
```diff
@@ -42,6 +66,36 @@ def test_donors_count(self):
         self.assertEqual(response["total_donors"], 1)
         self.assertEqual(response["donated_amount"], Decimal("8.00"))
 
+    def test_expected_amount_in_context(self):
+        """Expected amount should be calculated and included in context."""
+        response = donation_form_with_heart({"user": None})
+        self.assertIn("expected_amount", response)
+        self.assertIsInstance(response["expected_amount"], Decimal)
+        # Expected amount should be a positive value based on day of year
+        self.assertGreater(response["expected_amount"], 0)
```
sarahboyce: I would remove this test as it is also tested below

fundraising/tests/test_templatetags.py
```diff
@@ -42,6 +66,36 @@ def test_donors_count(self):
...
+
+    def test_expected_amount_calculation(self):
```
sarahboyce, line 68:
```suggestion
    @patch("fundraising.templatetags.fundraising_extras.GOAL_AMOUNT", 500000)
    def test_expected_amount_calculation(self):
```
The `GOAL_AMOUNT` amount value was updated and so this test and the one below now fail, I think it makes sense to patch these values

aadeina: Done. Added `@patch` to both tests

djangoproject/scss/_style.scss
aadeina: Makes sense, removed it along with the CSS

fundraising/tests/test_templatetags.py
aadeina: Removed

aadeina:
Welcome back, @sarahboyce!
I've applied the fixes, removed the `<progress>` element, cleaned up the tests , and rebased onto main.
Let me know if everything looks good or if you’d like any other tweaks!

PR #2459, opened 2026-01-16
Removes focus on nav links in mobile screen when not visible

SaptakS:
While [reviewing another PR](https://github.com/django/djangoproject.com/pull/2442#pullrequestreview-3670754345), I noticed that all navigation links were receiving focus when they are not visible in mobile screen. Currently, we are just hiding the navigation in mobile screens visually (by making the `max-height: 0`). This causes the links inside the header navigation to receive focus even when they are not visible, which makes the keyboard accessibility bad because there is no visual element getting a focus.

This PR hides it completely (using `display: none`) so that the nav links don't receive focus when they are not visible. I have also tried to keep the transition effect intact using `@starting-style`.

I also added `aria-hidden: true` to the menu icon on the menu button.

### DIFF
--- a/djangoproject/templates/includes/header.html
+++ b/djangoproject/templates/includes/header.html
@@ -7,7 +7,7 @@
     <a class="logo" href="{% url 'homepage' %}">Django</a>
     <p class="meta">The web framework for perfectionists with deadlines.</p>
     <button class="menu-button">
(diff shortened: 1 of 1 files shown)


sabderemane (approved):
LGTM, thanks! ✨

PR #2463, opened 2026-01-19
Add eslint pre-commit hook

codingjoe:


JohananOppongAmoateng:
biomejs is also a js linter/formatter we could consider

codingjoe:
> biomejs is also a js linter/formatter we could consider

Good point, though the current formatter is actually prettier. I was considering ESLint mostly to catch errors. Its high level of maturity was the main consideration.

That being said, there's also rslint, oxlint, jslint and probably some more linters. None of which have the option to target a specific ECMAScript version. Considering that we ship JS without a bundler (which I believe is excellent), that is a fairly mandatory option. Since you don't want to use JS features not yet available in all browsers.

adamzap:
@codingjoe How does this relate to our current JavaScript formatter, `prettier`?

codingjoe:
> @codingjoe How does this relate to our current JavaScript formatter, `prettier`?

One is a linter, and one is a formatter. ESLint will complain about unused variables or other code smells.

You can look at the errors it currently throws and you'll see that this might just be a useful addition.

However, I don't want to fix the code before adding #2537
The contributing guidelines ask for tests, so tests you shall get! 😉

PR #2465, opened 2026-01-19
Add pyproject.toml and basic uv setup

ulgens:
Coming from https://github.com/django/djangoproject.com/pull/2321#issuecomment-3738199681

This PR handles

> phase 1 (compile textual requirement files with uv pip compile and install them with pip)

I preferred `export` over `pip compile`. It doesn't seem there are any functional differences between two approaches, it just felt more uv-native.

It seems we can't use hashes. With hashes, `pip install` fails with

```
ERROR: Can't verify hashes for these requirements because we don't have a way to hash version control repositories:
    django-push@ git+https://github.com/brutasse/django-push.git@22fda99641cfbd2f3075a723d92652a8e38220a5 from git+https://github.com/brutasse/django-push.git@22fda99641cfbd2f3075a723d92652a8e38220a5 (from -r requirements/tests.txt (line 207))
```

I also made use of [uv-pre-commit](https://github.com/astral-sh/uv-pre-commit) hook to ensure output files are always synchronized with the pyproject.toml file.

Overall, the PR introduces a new pyproject.toml based setup for managing dependencies and project settings, without breaking the existing tooling built on `requirements/` folder.


pyproject.toml
```diff
@@ -0,0 +1,74 @@
...
+]
+
+prod = [
+    "gunicorn==23.0.0",
+    "psycopg[c]==3.3.2",
+    "redis==7.1.0",
+    "sentry-sdk==2.49.0",
+]
+
+# TODO: Consider merging with dev group
```
marksweb, line 54: Yeah I think test & dev are really the same thing.

pyproject.toml
```diff
@@ -0,0 +1,74 @@
+[project]
+name = "djangoproject-com"
+version = "0.0.1"
+description = "Source code to djangoproject.com"
+requires-python = ">=3.12"
+classifiers = [
```
marksweb, line 6: Probably don't need these. They're for pypi metadata only I think.

ulgens: The biggest usecase is PyPI metadata but I think they are useful for communicating what is supported. Can also be used for generating badges: https://shields.io/badges/py-pi-versions-from-framework-classifiers

Python classifiers are pushed by the default pyproject-fmt setup. It can be disabled, but I see no harm.

I can remove them, there is not much benefit, but I'm on the side of keeping them documentation/communication purposes.

pyproject.toml
ulgens: @django/django-website How do you feel here?

ulgens:
There is already a conflict with `common.txt` file and #2467 introduces a new dependency. One of the PRs will need to be rebased, depending on which one gets merged first.

pyproject.toml
```diff
@@ -0,0 +1,74 @@
...
+    "requests==2.32.5",
+    "sorl-thumbnail==12.11.0",
+    "sphinx==8.1.3",
+    "stripe==13.2.0",
+    "time-machine==3.2.0",
+]
+
+[dependency-groups]
+dev = [
+    "black==26.1.0",               # Used by Django's find_formatters and run_formatters
```
codingjoe:
```suggestion
    { include-group = "test" },
    "black==26.1.0",               # Used by Django's find_formatters and run_formatters
```

pyproject.toml
```diff
@@ -0,0 +1,74 @@
...
+    "redis==7.1.0",
+    "sentry-sdk==2.49.0",
+]
+
+# TODO: Consider merging with dev group
+tests = [
+    "coverage==7.13.1",
+    "requests-mock==1.12.1",
+    "tblib==3.2.1",
+    { include-group = "dev" },
```
codingjoe: I usually do it the other way your CI suite don't need all thequality-of-life depdencies of a developer.

pyproject.toml
```diff
@@ -0,0 +1,74 @@
...
+]
+
+[tool.uv]
+package = false
+# The earliest version that supports the current lock file revision
+required-version = ">=0.8.4"
+
+[tool.pyproject-fmt]
+indent = 4
```
codingjoe: I prefer 2 spaces, but if we change this, we need to update the `.editorconfig` too.

pyproject.toml
```diff
@@ -0,0 +1,74 @@
+[project]
+name = "djangoproject-com"
+version = "0.0.1"
+description = "Source code to djangoproject.com"
+requires-python = ">=3.12"
```
codingjoe: If you want uv to install the correct version and not jump to 3.14:
```suggestion
requires-python = "==3.12"
```

pyproject.toml
```diff
@@ -0,0 +1,74 @@
...
+classifiers = [
+    "Framework :: Django",
+    "Framework :: Django :: 6.0",
+    "License :: OSI Approved :: BSD License",
+    "Operating System :: OS Independent",
+    "Programming Language :: Python :: 3 :: Only",
+    "Programming Language :: Python :: 3.12",
+]
+dependencies = [
+    "babel==2.17.0",
```
codingjoe: I don't like that all dependencies are pinned there. That will be funny with dependabot. Pinning is done by the lock file. UV defaults to `>=` statements when you add new dependencies. I'd love to use this and only pin, with a comment, when an update isn't possible.

pyproject.toml
```diff
@@ -0,0 +1,74 @@
+[project]
+name = "djangoproject-com"
+version = "0.0.1"
```
codingjoe: It's arbitrary, but:
```suggestion
version = "0.1.0"
```

pyproject.toml
codingjoe: No one asked me, but I'd disagree. Dev should include test dependencies, but your CI system shouldn't install them to be closer to the production environment.

.pre-commit-config.yaml
codingjoe, line 1: You can actually remove the fixed python version argument from `pyupgrage`. It'll pick it up from the project file.

codingjoe (commented):
I love the initiative. You might want to update the PR title, though. Adding UV is a little more than adding a config file 😉 

That aside, I'd love to use the uv actions in the CI workflow too.  And possibly even in Docker. I have done this plenty of times; I am happy to help.

Lastly, maybe open a new PR for just the pyproject.toml file, without UV. Just to get a head start, other people can start moving configurations into the `pyproject.toml` file, like from the liner suite. In turn this PR could focus on UV.

pyproject.toml
ulgens: Fixed.

pyproject.toml
ulgens: > That will be funny with dependabot.

I'm missing how. Dependabot already has multiple issues with UV, and it wasn't in an usable state last time I tried, so I'm not worried about it, but I also didn't experience a case with it having issues with pinned dependencies with other tools. My personal take here is that using the ">=" approach can lead to noisy update PRs, risk reproducibility issues when updating/experimenting with packages, and require a more careful overall approach when working with packages. Meanwhile, the patch pinning has none of the issues, and it doesn't introduce other problems.

Considering that pinning to the patch works fine, and we don't need a broader support range like a package does, I'd prefer to keep them as is.

pyproject.toml
ulgens: Every opinion is appreciated @codingjoe 🌻 

I kinda agree that a CI environment running tests may not need all the dev dependencies, but I couldn't catch what you mean here:

> but your CI system shouldn't install them to be closer to the production environment.

pyproject.toml
ulgens: I'm okay with both, but asking to understand, what is the benefit?

pyproject.toml
ulgens: Makes sense, I'll revisit this.

ulgens:
@codingjoe Thanks for the review.

> You might want to update the PR title, though. Adding UV is a little more than adding a config file 

I'm lost here. The PR title doesn't mention "adding UV".

>  I'd love to use the uv actions in the CI workflow too. And possibly even in Docker. 

The original PR #2321 was handling all related changes, but then I created a smaller one with @pauloxnet 's recommendation. The CI and Docker-related changes will be handled in a follow-up PR.

> maybe open a new PR for just the pyproject.toml file, without UV. 

Coming from #2321, this is pretty much _that_ PR which adds the pyproject.toml file without the overall uv - pip replacement. I can try to strip down the PR even more, but I don't think it will contribute much.

I'll be responding to the inline comments in related threads.

.pre-commit-config.yaml
ulgens: I remember something similar, and I'd like to clean a duplicate config but I couldn't find anything explaining this behaviour in pyupgrade repo. Would you mind sharing a doc/code reference?

codingjoe:
> @codingjoe Thanks for the review.
> 
> > You might want to update the PR title, though. Adding UV is a little more than adding a config file
> 
> I'm lost here. The PR title doesn't mention "adding UV".

Sorry, I didn't communicate this well. My point was to change the title to `Added UV & pyproject.toml` since you are adding uv.


> > I'd love to use the uv actions in the CI workflow too. And possibly even in Docker.
> 
> The original PR #2321 was handling all related changes, but then I created a smaller one with @pauloxnet 's recommendation. The CI and Docker-related changes will be handled in a follow-up PR.
> 
> > maybe open a new PR for just the pyproject.toml file, without UV.
> 
> Coming from #2321, this is pretty much _that_ PR which adds the pyproject.toml file without the overall uv - pip replacement. I can try to strip down the PR even more, but I don't think it will contribute much.

I see your point. I wasn't aware of the origins of this PR. It does clarify things a little. Thanks!

pyproject.toml
codingjoe: An initial version in semver usually start with 0.1.0. `uv init` does it, so does `npm init`.

pyproject.toml
codingjoe: Hm... now seeing where this PR comes from, I understand your solution. I think this can be changed once everything runs on UV.

I usually have Dependabot configured to only update the lock. It keeps diffs small, and, dependabot won't try to update major versions that I pinned explicitly.

pyproject.toml
codingjoe: I meant that the CI system shouldn't have any dev dependencies installed. 
Let's say you use debug toolbar in development. And you end up importing it somewhere without handling the import error. You won't notice it in dev, and not in CI if all your dev dependencies are installed there too. Similar how test run with the debut setting at false.

ulgens: Ah okay, makes sense. I'll revisit this.

pyproject.toml
ulgens: Oh okay.

ulgens: Fixed.

ulgens:
> My point was to change the title to Added UV & pyproject.toml since you are adding uv.

I updated the title 👍🏻

.pre-commit-config.yaml
codingjoe: Oh NO. I was mistaken! Only django-upgrade supports pyproject parsing. Adding support has been rejected in the past, but maybe he's changed his mind 🤷 see also https://github.com/asottile/pyupgrade/issues/768#issuecomment-3996444899

pyproject.toml
```diff
@@ -0,0 +1,74 @@
...
+description = "Source code to djangoproject.com"
+requires-python = ">=3.12"
+classifiers = [
+    "Framework :: Django",
+    "Framework :: Django :: 6.0",
+    "License :: OSI Approved :: BSD License",
+    "Operating System :: OS Independent",
+    "Programming Language :: Python :: 3 :: Only",
+    "Programming Language :: Python :: 3.12",
```
ronnzw, line 8: If you are going to keep classifiers, l would suggest matching the version of Python with that at line 5.

ulgens: It's already matching, am I missing something?

.pre-commit-config.yaml
ulgens: @codingjoe The referenced issue got closed without any explanation (not a surprise but still sad.)

pyproject.toml
codingjoe: Good point, do you want to keep them though? They are PyPi classifiers and we're not publishing anything on PyPi, are we?

ulgens: I mean... I see no harm, and it gives a summary about the project. If we agree that they are redundant, I can remove them but I also don't think that a little bit of redundancy is a bad thing.

codingjoe: `¯\_(ツ)_/¯`

pyproject.toml
adamzap: I'd say let's remove them to keep things minimal.

pyproject.toml
ulgens: Removed all except the ones pyproject-fmt is looking for: https://pyproject-fmt.readthedocs.io/en/latest/formatting.html#:~:text=Python%20version%20classifiers

pyproject.toml
ulgens: Removed all except the ones pyproject-fmt is looking for: https://pyproject-fmt.readthedocs.io/en/latest/formatting.html#:~:text=Python%20version%20classifiers

pyproject.toml
ulgens: Resolved.

pyproject.toml
ulgens: Resolved.

pyproject.toml
ulgens: Resolved.

ulgens:
Rebased from the latest main, synced deps, fixed change requests. Ready for another round of interview.

pauloxnet (approved):
What's the blocker here?
Can you maybe align back all dependencies from master?

ulgens:
> What's the blocker here?

It was waiting for review.

>  Can you maybe align back all dependencies from master?

Done.

ulgens:
Rebased again, over #2675 and #2679

sabderemane:
Hey @ulgens , sorry for the delay on this. Could you resolve the conflicts please? I will review it so it can be merged 🚀

ulgens:
@sabderemane Done. Would be nice if we can merge without another conflict.

pyproject.toml
ulgens: `.editorconfig` defines

```
[*]
...
indent_size = 4
```

and no other special rules for *.toml.

pauloxnet (approved):
Approved. Again :)

sabderemane (approved):
Thank you @ulgens ! 🎸

PR #2466, opened 2026-01-20
Remove GitHub auth link on registration page

### LINKED ISSUE #835: Github login doesn't work for community pages

vidhiii1711:
GitHub social login currently succeeds at the OAuth step but does not create
a local user account, resulting in users being redirected back to the login
page when attempting to add RSS feeds.

This PR temporarily disables the GitHub login link to avoid exposing users
to a broken authentication flow.

Scope:
-> UI-only change
-> No authentication or backend logic modified

Follow-up:
I’m happy to work on a proper fix for the GitHub/allauth configuration if the
maintainers agree on the expected behavior and approach.

temporily fixes https://github.com/django/djangoproject.com/issues/835

### DIFF
--- a/djangoproject/templates/registration/registration_form.html
+++ b/djangoproject/templates/registration/registration_form.html
@@ -8,7 +8,7 @@
 
   <p>
   {% blocktranslate trimmed %}
-     Github login is currently unavailable.
+  Github login is currently unavailable.
   {% endblocktranslate %}
   </p>
 
--- a/djangoproject/templates/registration/registration_form.html
+++ b/djangoproject/templates/registration/registration_form.html
@@ -7,9 +7,9 @@
   <h1>{% translate "Create an account" %}</h1>
 
   <p>
-  {% blocktranslate trimmed %}
-  Github login is currently unavailable.
-  {% endblocktranslate %}
+    {% blocktranslate trimmed %}
+      Github login is currently unavailable.
+    {% endblocktranslate %}
   </p>
 
   {% if form.errors %}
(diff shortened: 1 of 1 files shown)


adamzap:
I'm going to close this for now. If you want to work on this, could you propose a change in #835 instead? Thanks!

adamzap:
Actually, @vidhiii1711, would you be willing to remove the GitHub login link instead of changing it?

vidhiii1711:
Hi @adamzap, thanks for the clarification!
Yes, I’d be happy to work on removing the GitHub login link instead. I’ll take a look at issue #835 and propose a change there.
Please let me know if there are any specific considerations I should keep in mind while removing it.

vidhiii1711:
Hi, I’ve removed all GitHub login references from the signup/login flow as suggested.

I also updated the related tests to ensure they pass. Please let me know if anything else needs to be adjusted.

adamzap (approved):
👍 Looks great!

@alexgmin Could you also review this to confirm that this is the change we discussed?

alexgmin (approved):
Yeah, this is what we discussed so it's ready to merge.

PR #2467, opened 2026-01-21
Incorporated checklist-generator app into main project.

nessita:
See two commits. The new app `checklists` is a revised port of https://github.com/nessita/checklist-generator. This app is meant to be used by Django Fellows and Releasers via the admin, and has no exposure in the website.


ulgens:
Can you please link if there are any previous discussions/documentation? Not sure how to review it as is.

nessita:
> Can you please link if there are any previous discussions/documentation? Not sure how to review it as is.

Hello Ülgen, thank you for your interest in this PR. As mentioned in the description, this work is primarily intended for Django Fellows (and Releasers), who are the direct users of this new app. This change is not a user-facing feature, so it is not intended for general review in the same way as end-user/public functionality.

The context for this work has been shared over time in Fellows reports (mostly in my reports in the forum), including recent updates on how the release checklist generator is used in practice and how it's was decided to integrate into djangoproject.com for having it deployed in a shared place with the other Fellows. This has been done with the awareness and approval of the Ops Team, our line manager, and our Board liaison.

Review and feedback from Fellows is the main input we are looking for here.

ulgens:
Thanks for the details. I do understand the case, but it feels weird to get a huge PR with no issue and then exclude @django/django-website from a review that includes code changes, even if it's not user-facing. I see that @SaptakS gave a 👍🏻 so I believe this can be discussed internally, no need to clog up the PR.

jefftriplett:
I think @nessita scoped this pull request clearly and set expectations up front. She noted that the feature is intended for Django Fellows and Releasers via the admin, and explicitly asked Jacob to review it.

This work has also been discussed before in other contexts, including Fellows reports and internal conversations, so it is not appearing out of nowhere.

The Foundation and the Fellows have operational needs that are not always part of the website working group’s roadmap or review process. Some of this work exists to support day-to-day operations and to keep things running smoothly, rather than to introduce user-facing changes.

That said, comments framing this as “weird” are not very helpful. If there are concrete concerns about scope, ownership, or maintenance, it would be more useful to call those out directly so they can be addressed.

ulgens:
> This work has also been discussed before in other contexts, including Fellows reports and internal conversations, so it is not appearing out of nowhere.

The PR description doesn't have any references or mentions to what you refer to here. From the perspective of a contributor/maintainer who wasn't part of those discussions, this PR indeed appeared out of nowhere. With the added explanation from the comments above, I do agree that the context is clear now.

>  If there are concrete concerns about scope, ownership, or maintenance, it would be more useful to call those out directly so they can be addressed.

Anything added to the repo comes with a maintenance load, and adding a full-fledged app is not a small change. My concerns were about the feature maintenance of this module, meaning that even if WG members or the online users won't be using the resulting features, the maintainers will need to understand the code and structure. I understand the related discussion/planning was done somewhere else, but I was asking for a link to see if any public docs exist, and I'm not sure what else would be a better approach here. We may want to consider adding a CODEOWNERS entry to clarify these types of cases in the future.

We already talked with @SaptakS about the responsibilities/expectations on website PRs, and I don't want to drag another loaded discussion so I'll be marking my comment as off topic (I'm active on Slack if further communication is needed).  The PR didn't have the context in its first iteration, and it now has, so I consider this an improvement. Thanks.

.pre-commit-config.yaml
```diff
@@ -51,7 +51,7 @@ repos:
       hooks:
           - id: prettier
             exclude_types: [html, json, scss]
-            exclude: '(^djangoproject\/static\/js\/lib\/.*$)'
+            exclude: '(^djangoproject\/static\/js\/lib\/.*$|^checklists\/templates\/.*$)'
```
jacobtylerwalls, line 54: Leaving a note that in my testing this was needed because `md` is not recognized as an excludable type by `identify`/`pre-commit`/`prek`, so some form of regex is needed here.

releases/models.py
```diff
@@ -160,6 +161,7 @@ def upload_to_checksum(release, filename):
     return f"pgp/Django-{version}.checksum.txt"
 
 
+@total_ordering
```
jacobtylerwalls, line 164: Thanks, TIL about this helper. Docs [say](https://docs.python.org/3/library/functools.html#functools.total_ordering):

> In addition, the class should supply an `__eq__()` method.

README.rst
```diff
@@ -256,6 +256,45 @@ To update this file, run::
 
     python -m manage dumpdata dashboard --exclude dashboard.Datum --indent=4 > dashboard_production_metrics.json
 
+Release Checklists
```
jacobtylerwalls, line 259: What about moving this to `checklists/README.md` and then just linking from here?

checklists/apps.py
```diff
@@ -0,0 +1,8 @@
+from django.apps import AppConfig
+from django.utils.translation import gettext_lazy as _
+
+
+class ChecklistsConfig(AppConfig):
+    name = "checklists"
+    verbose_name = _("Release Checklists")
+    default_auto_field = "django.db.models.AutoField"
```
jacobtylerwalls: Other apps in this repo don't appear to override `default_auto_field`.

checklists/models.py
```diff
@@ -0,0 +1,1029 @@
...
+        )
+
+
+class SecurityIssueReleasesThrough(models.Model):
+    securityissue = models.ForeignKey(
+        "SecurityIssue", on_delete=models.CASCADE, verbose_name="Security Issue"
+    )
+    release = models.ForeignKey(Release, on_delete=models.CASCADE)
+    commit_hash = models.CharField(
+        max_length=128, default="", blank=True, db_index=True
```
jacobtylerwalls, line 473: Curious why we need `db_index`? (same on the other one)

releases/models.py
nessita: Thank you for calling this out! I tried to write a test to demonstrate that `__eq__`  was missing/needed, but I couldn't. This is because Django's model define this `__eq__` by default:
```python
> /home/nessita/.venvs/djangoproject.com/lib/python3.14/site-packages/django/db/models/base.py(616)__eq__()
-> def __eq__(self, other):
(Pdb) ll
616  ->     def __eq__(self, other):
617             if not isinstance(other, Model):
618                 return NotImplemented
619             if self._meta.concrete_model != other._meta.concrete_model:
620                 return False
621             my_pk = self.pk
622             if my_pk is None:
623                 return self is other
624             return my_pk == other.pk
```

Considering that `Release` defines the `version` string as PK, even for unsaved release instances, this test pass:
```python
    def test_equality(self):
        r1 = Release(version="5.2")
        r2 = Release(version="5.2")
        self.assertLessEqual(r1, r2)
        self.assertLessEqual(r2, r1)
        # If r1 <= r2 and r2 <= r1, then r1 == r2 should be True.
        self.assertEqual(r1, r2)
```

Because:
```
> /home/nessita/.venvs/djangoproject.com/lib/python3.14/site-packages/django/db/models/base.py(624)__eq__()
-> return my_pk == other.pk
(Pdb) my_pk
'5.2'
(Pdb) other.pk
'5.2'
```

In summary I will leave the test but not add `__eq__` to the `Release` model.

README.rst
nessita: Great idea, will do!

checklists/models.py
nessita: I added it to be able to search for them (they are part of the admin's `search_fields`), happy to remove if you think they are early optimization.

PR #2470, opened 2026-01-21
Update goal amount for fundraising to $500,000

ryancheley:
Update goal based on feedback from the Fundraising WG and approval from the DSF Board

### DIFF
--- a/fundraising/models.py
+++ b/fundraising/models.py
@@ -11,7 +11,7 @@ from sorl.thumbnail import ImageField
 
 from djangoproject.thumbnails import LogoThumbnailMixin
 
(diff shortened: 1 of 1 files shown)


jefftriplett:
> LGTM based on the discussion in the issue. I think at some point should make this be editable from the admin.

Having a goal + amount achieved historically sounds useful.

sarahboyce:
@django/ops-team please can you deploy so the donation amount updates ❤️

tobiasmcnulty:
> @django/ops-team please can you deploy so the donation amount updates ❤️

Done! https://www.djangoproject.com/fundraising/

Issue #1122, opened 2021-11-15
Version indicator on  should be more prominent


rolandcrosby:
I ended up on the [Django 2.2 version](https://docs.djangoproject.com/en/2.2/intro/tutorial01/) of the "Writing your first app" tutorial via web search, and didn't realize I was looking at outdated docs until I noticed the URL a few pages in. I realize that 2.2 [is still getting extended support](https://www.djangoproject.com/download/#supported-versions), but regardless this probably isn't the version of the tutorial that most people want, and the design of the page gives no indication that you're not looking at the latest version. Completely unsupported docs like [those for 1.8](https://docs.djangoproject.com/en/1.8/intro/tutorial01/) have a helpful banner at the top of the page indicating that the version isn't supported. I'm not saying a big scary red banner is necessary for old versions that are still supported, but some indication that there are newer versions would be nice.

This is a problem that a lot of projects face and solve in different ways - some examples below.

<details>
<summary>Examples of how other projects deal with this</summary>

- Postgres has a prominent version switcher at the top of every docs page indicating what versions are supported, plus a warning when you're looking at an unsupported version:

![Postgres docs version switcher](https://user-images.githubusercontent.com/60245/141797050-65652eb3-a861-46e1-8a89-5b8a023191da.png)

- Docs sites that use Read The Docs have a nice little "there is a newer version available" note at the top of outdated pages:
![Requests outdated version indicator](https://user-images.githubusercontent.com/60245/141797183-589d93df-2833-4f5a-8ae0-e749ec0832ee.png)

![standard RTD newer version indicator (csvkit)](https://user-images.githubusercontent.com/60245/141797394-f61901b3-6292-41b8-96a9-cda1b230a799.png)

- CockroachDB provides an explicit warning on docs for old versions to indicate when they will stop being supported:
![CockroachDB support warning](https://user-images.githubusercontent.com/60245/141798057-68c6dc79-d775-4dc6-a9dc-770035fa6352.png)

</details>

sophocodes:
Hi @pauloxnet can I work on this issue? I think it's good to introduce me to the Django project 😊

pauloxnet:
In the documentation I already see banners for unsupported versions and for the development version, I'm not sure if you want to show warnings for supported Django versions as well, for example 3.2 LTS could legitimately be used with the intend to update only when the next LTS release comes out, so it might be annoying for users to see warnings in this case.

I'd like to read other people's comments on this issue as well before implementing anything without a thought.

[...]

PR #2471, opened 2026-01-23
Add info banner for supported but non-latest documentation versions

kehach07:
Refs #1122 — Docs for old supported versions should indicate that a newer version is available

Adds a small informational banner to documentation pages when viewing a supported but non-default Django version, letting users know a newer version exists and linking to the latest docs. The banner is subtle and does not affect unsupported, preview, or development version warnings.

### DIFF
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -3704,3 +3704,14 @@ ul.corporate-members li {
         top: 0;
     }
 }
+.docs-info-banner {
+  background-color: #eef6ff;
+  border-left: 4px solid #3b82f6;
+  padding: 10px 14px;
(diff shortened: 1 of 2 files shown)


adamzap:
@kehach07 Could you provide a screenshot for this change?

kehach07:
> @kehach07 Could you provide a screenshot for this change?

Thanks for the review! @adamzap, 

Locally I only have one supported docs version (6.0), so it is treated as the
default and the banner does not normally appear. I temporarily relaxed the
condition to verify the UI and capture this screenshot.

In production, where multiple supported versions exist, the original condition
(release.is_supported and not release.is_default) works as intended.
image

PR #2485, opened 2026-01-28
Normalize URLs with extra slashes

### LINKED ISSUE #2067: The links and images do not work when extra slashes are in url
On the "How to upgrade version" page (https://docs.djangoproject.com/en/5.2/howto//upgrade-version/), the link to the deprecation timeline (https://docs.djangoproject.com/en/5.2/howto/internals/deprecation/) leads to a 404. 

I was able to find the actual deprecation timeline (https://docs.djangoproject.com/en/dev/internals/deprecation/) via a web search, but it should be accessible from the documentation for stable and old versions.

vidhiii1711:
Fixes #2067

This PR adds middleware to normalize URLs containing extra slashes
(e.g. `//community//` → `/community/`) so links and assets resolve correctly.

Includes tests to ensure correct redirects on www.djangoproject.com.

### DIFF
--- a/djangoproject/middleware.py
+++ b/djangoproject/middleware.py
@@ -3,7 +3,8 @@ from django.http.request import split_domain_port
 from django.middleware.locale import LocaleMiddleware
 from django.utils.functional import cached_property
 from django.utils.http import is_same_domain
-
+from django.http import HttpResponsePermanentRedirect
+from django.urls import resolve, Resolver404
 
 class CORSMiddleware:
     """
@@ -50,3 +51,23 @@ class ExcludeHostsLocaleMiddleware(LocaleMiddleware):
         if self._is_host_included(request.get_host()):
             return super().process_response(request, response)
         return response
+
+class NormalizeSlashesMiddleware:
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        path = request.path
+        if "//" in path[1:]:
+            normalized_path = "/" + "/".join(filter(None, path.split("/"))) + "/"
+
+            if normalized_path != path:
+                try:
+                    resolve(normalized_path)
+                except Resolver404:
+                    pass
+                else:
+                    return HttpResponsePermanentRedirect(normalized_path)
+
+        return self.get_response(request)
--- a/djangoproject/middleware.py
+++ b/djangoproject/middleware.py
@@ -1,10 +1,11 @@
 from django.conf import settings
+from django.http import HttpResponsePermanentRedirect
 from django.http.request import split_domain_port
 from django.middleware.locale import LocaleMiddleware
+from django.urls import Resolver404, resolve
 from django.utils.functional import cached_property
 from django.utils.http import is_same_domain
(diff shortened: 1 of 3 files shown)


djangoproject/middleware.py
```diff
@@ -50,3 +52,24 @@ def process_response(self, request, response):
...
+
+class NormalizeSlashesMiddleware:
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        path = request.path
+        if "//" in path[1:]:
+            normalized_path = "/" + "/".join(filter(None, path.split("/"))) + "/"
```
codingjoe: RegEx sublimation might simplify the implementation and help you to avoid branching (which will slow down all requests).

djangoproject/middleware.py
```diff
@@ -50,3 +52,24 @@ def process_response(self, request, response):
...
+
+            if normalized_path != path:
+                try:
+                    resolve(normalized_path)
+                except Resolver404:
+                    pass
+                else:
+                    return HttpResponsePermanentRedirect(normalized_path)
+
+        return self.get_response(request)
```
codingjoe, line 76: You can wrap this in a try block, and only fall back if a 404 was raised.
The default path should probably be the good path for the CPU to preload the correct instructions for the majority of requests.

djangoproject/tests/test_www_urls.py
```diff
@@ -5,3 +5,15 @@ class IndividualMembershipNominationRedirectTests(TestCase):
     def test_individual_membership_nomination_redirect_exists(self):
         response = self.client.get("/foundation/individual-membership-nomination/")
         self.assertEqual(response.status_code, 302)
+
+
+class NormalizeSlashesMiddlewareTests(TestCase):
+    def test_double_slash_redirects_to_single_slash(self):
```
codingjoe, line 11: Maybe consider adding a few adverse tests to sanity check your solution.

codingjoe (commented):
Hi there 👋,

I had some time and was browsing for reviews. I hope you don't mind me leaving some notes.

Cheerio!
Joe

vidhiii1711:
@codingjoe 
Hi, thanks a lot for taking the time to review this and for the helpful suggestions!

I’ll update the implementation to simplify the slash normalization (likely using regex as you suggested) and reduce branching in the middleware.

I’ll also explore the try/except approach so the normalization only happens when a 404 occurs, keeping the common path fast.

Additionally, I’ll add a few adverse test cases to better cover edge scenarios.

Thanks again for the feedback — I’ll push an update shortly.

vidhiii1711:
I have updated the middleware to normalize multiple slashes in the URL using a regex .

I also added a test case in test_www_urls.py to ensure that URLs like /community// correctly redirect to /community/ with a 301 status.

Please let me know if any further changes are needed. Thanks again for the guidance!

codingjoe (approved):
Slick solution! LGTM

adamzap (approved):
👍 Looks great!

pauloxnet (approved):

PR #2491, opened 2026-02-03
Fix color scheme issues on dark mode theme, and auto theme along with prefers-media-querry: dark

ZougaRida:
For example this how the scrollbar looks like in the search component in dark mode theme *(see it [here](https://docs.djangoproject.com/en/6.0/search/?q=models))* 

image

 the same thing would happen in auto-theme with prefers-media-querry: dark.
 
 The proposed solution not only fixes this issue, but any other issues relating to color-scheme for scrollbars and form's related items that were not targeted by specific style.


sabderemane (approved):
Nice, thank you for your contribution! 🎸

PR #2493, opened 2026-02-03
Fixed Akismet testing.

sarahboyce:
See https://github.com/grundleborg/pykismet3?tab=readme-ov-file#comment-check it should be `is_test`

Note that our Akismet tests are failing: https://github.com/django/djangoproject.com/actions/runs/21634722218/job/62356829891?pr=2492


ulgens:
Do we know how this test was passing before?

sarahboyce:
> Do we know how this test was passing before?

I assume it just wasn't being marked as spam before by Akismet :shrug:

PR #2497, opened 2026-02-04
[checklists] Advise to mention EOL dates

jacobtylerwalls:
Before, our release notes generally only mentioned the anticipated EOL date on LTS releases, and even then only vaguely. Add a reminder to start adjusting our release note skeleton to mention the dates on all releases. This can help with discoverability of this info.

### DIFF
--- a/checklists/templates/checklists/release-skeleton.md
+++ b/checklists/templates/checklists/release-skeleton.md
@@ -47,12 +47,13 @@ At this point, most of the larger features planned for {{ release.feature_versio
 ## Release Day
 
 - [ ] Polish and make cosmetic edits to release notes on `main` and backport
-    {% if not release.is_pre_release %}
+    {% if release.is_pre_release %}
+    - Mention the anticipated end of mainstream and extended support dates
+    {% else %}
     - Remove the `Expected` prefix and update the release date if necessary
     {% endif %}
     {% if release.is_dot_zero %}- Remove the `UNDER DEVELOPMENT` header at the top of the release notes:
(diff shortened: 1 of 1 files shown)


checklists/templates/checklists/release-skeleton.md
```diff
@@ -46,12 +46,13 @@ At this point, most of the larger features planned for {{ release.feature_versio
...
 ## Release Day
 
-- [ ] Polish and  make cosmetic edits to release notes on `main` and backport
+- [ ] Polish and make cosmetic edits to release notes on `main` and backport
     {% if not release.is_pre_release %}
     - Remove the `Expected` prefix and update the release date if necessary
     {% endif %}
     {% if release.is_dot_zero %}- Remove the `UNDER DEVELOPMENT` header at the top of the release notes:
         - e.g. https://github.com/django/django/commit/1994a2643881a9e3f9fa8d3e0794c1a9933a1831{% endif %}
+    - Mention the ancticipated end of mainstream and extended support dates.
```
nessita: This item has a small typo (an*c*ticipated), and also it will show in every patch/bugfix release as well, so we may want to guard this with `{% if release.is_pre_release %}` (so we fix this in alpha, beta, or RC. So maybe move up to the existing `if` and put it the as a `else`?

Or even better, make the guard positive and  have:

```django	
    {% if release.is_pre_release %}
    - Mention the anticipated end of mainstream and extended support dates.
    {% else %}
    - Remove the `Expected` prefix and update the release date if necessary
    {% endif %}
```

(I think we are not adding period at the end of bulleted items.)

Issue #2446, opened 2026-01-10
Migrate to ruff


ulgens:
Coming from https://github.com/django/djangoproject.com/pull/2321#issuecomment-3730504631

How do we feel about using ruff as the formatter & linter, instead of the current black + isort + flake8 + pyupgrade combo? To me, it feels like the natural next step for the evolution of any Python stack, but is there any concern about keeping the tooling in sync with https://github.com/django/django repo? IMO, testing the implementation here would help with the feature discussions about django core repo.

Implementation-wise, because we wrap the tools with git hooks (locally and in CI), this change wouldn't affect how we call them - the interface or processes wouldn't be changing.

Relates to:
- #2433

pauloxnet:
I agree to replace old tools (flake8, sort, py upgrade) with ruff

The Django website is a Django project and not a framework like Django that has to concern about retrocompatibility so it's ok for us to improve the toolset and maybe then propose our improvement back to the framework instead of being stuck with old tools. 

I would propose to set up ruff iteratively: replicate the same configuration as we have now with multiple tools with the minimal changes in the code, and only when merged activate new rules one by one to improve the code with small atomic PR, easy to review, merge and maybe revert.

vidhiii1711:
Hi, thanks for the issue.

Before working on this, could you clarify whether you’re expecting:

-> a concrete implementation,
-> or exploratory work / discussion only?

Happy to help once the intended scope is clear.

Thanks.

pauloxnet:
> Before working on this, could you clarify whether you’re expecting:
> 
> -> a concrete implementation, -> or exploratory work / discussion only?

I think, an exploratory work, before the implementation would be great.

vidhiii1711:
Thanks for confirming. I’ll begin exploratory work and report back with findings.

vidhiii1711:
Exploratory findings:
-> tox.ini is not present in the repository.
-> Lint configuration currently lives in setup.cfg (flake8, isort).
-> .pre-commit-config.yaml does exist and includes hooks for:
       flake8
       isort
       black
I checked CI workflows and flake8/isort aren’t explicitly invoked there—they’re only configured in setup.cfg. 
This suggests CI relies on tool auto-discovery. I think migrating config to pyproject.toml should be safe.
 I can prepare a small PR if that sounds right.

ulgens:
@vidhiii1711 

* The project doesn't use tox.
[...]

PR #2499, opened 2026-02-05
Replaced flake8/isort with ruff configuration

vidhiii1711:
Description:

This PR introduces a ruff.toml configuration to replace the existing flake8 and isort setup.

The configuration aims to be a 1-to-1 match with the current linting behavior:
-> Preserves existing ignores and file-specific exceptions (e.g. Django settings star imports).
-> Keeps line length and import ordering consistent with the current setup.
-> Avoids functional or formatting changes beyond what the existing tools already enforce.

This keeps linting behavior stable while consolidating configuration under Ruff, as discussed in #2446.


aggregator/forms.py
```diff
@@ -20,8 +20,7 @@ class FeedModelForm(forms.ModelForm):
             attrs={
                 "class": "required",
                 "placeholder": _(
-                    "Link to the RSS/Atom feed. Please only use "
-                    "Django-specific feeds."
+                    "Link to the RSS/Atom feed. Please only use " "Django-specific feeds."
```
marksweb:
```suggestion
                    "Link to the RSS/Atom feed. Please only use Django-specific feeds."
```

If this is going to be 1 line, might as well make it 1 string.

vidhiii1711:
@marksweb  
Thanks — applied the suggestion.

.pre-commit-config.yaml
```diff
@@ -40,28 +40,19 @@ repos:
...
-          - id: isort
-
-    - repo: https://github.com/pycqa/flake8
-      rev: 7.3.0
+    - repo: https://github.com/astral-sh/ruff-pre-commit
+      rev: v0.4.10
       hooks:
-          - id: flake8
+          - id: ruff
+            args: ["--fix"]
```
codingjoe, line 56: If you are dropping flake8, you might want to add `ruff-check` too.

ruff.toml
codingjoe, line 1: I'd prefer this to be in a `pyproject.toml` file. As a first step towards consolidating all project configuration there.

ruff.toml
```diff
@@ -0,0 +1,45 @@
...
+[lint]
+select = [
+    "E",
+    "F",
+    "W",
+    "I",
+]
+
+ignore = [
+    "E501",
```
codingjoe, line 12: A comment on both the select and ignore what those letters and codes mean goes a long way.

ruff.toml
```diff
@@ -0,0 +1,45 @@
...
+    "djangodocs",
+    ".direnv",
+    ".venv",
+    "**/migrations/*",
+]
+
+[lint.per-file-ignores]
+"djangoproject/settings/*.py" = ["F405"]
+"docs/tests.py" = ["E501"]
+"**/migrations/*" = ["RUF012"]
```
codingjoe, line 25: This has been added to the exclude list.

ruff.toml
```diff
@@ -0,0 +1,45 @@
...
+
+ignore = [
+    "E501",
+]
+
+exclude = [
+    "djangodocs",
+    ".direnv",
+    ".venv",
+    "**/migrations/*",
```
codingjoe, line 19: It's excluded here, but your patch includes countless migration files.

codingjoe:
After a lot of initial hesitation, I am a big fan of ruff now. However, did you consult the website working group about this?

vidhiii1711:
@codingjoe 
Thank you for the detailed review

All the points are well noted:
- I'll add inline comments explaining the ruff rule codes
- I'll revert the migration files from the diff
- I'll move the config from `ruff.toml` into `pyproject.toml`
- I'll add both `ruff` and `ruff-format` hooks in the pre-commit config

Regarding the website working group consultation — I should have done that before opening this PR. I'll reach out to them first and will update the PR once confirmation comes through.

Thanks again for your patience and guidance!

bhavanasam:
unsubscribe pls

pauloxnet:
@vidhiii1711 can you rebase this branch and solve the conflicts?

ruff.toml
ulgens: I'd prefer having a separate file here. Depending on the complexity of pyproject.toml and ruff.toml, having everything in a single file doesn't always provide the best experience. Also, we don't have a pyproject.toml file yet and I don't think we need to block this task while adding one.

ruff.toml
ulgens: +1. Sharing [the approach I follow](https://github.com/ulgens/django-blasphemy/blob/ad93c1141b546dd2c78a2a5f9cd001ad8c1d0782/ruff.toml#L97) as an example.

ruff.toml
ulgens, line 19: I don't think excluding migrations is the best way to go, they are code too.

ulgens:
+1 overall, added my comments here and there. One recommendation I'd give would be keeping the initial setup PR small, maybe without any rules, then gradually migrating from other tools with smaller PRs. Not that the current approach is bad, the incremental approach feels more managable and easier to review - discuss.

adamzap:
These changes look reasonable to me. However, I'd like to defer to the rest of the working group on tooling changes like this. I don't have a strong opinion and don't use code formatters personally.

@vidhiii1711 Would you be willing to rebase this branch and fix the merge conflicts?

vidhiii1711:
Thank you for the feedback @adamzap. I'll take some time to properly rebase the branch, resolve the merge conflicts, and push a clean update. Will keep you posted once it's ready for re-review.

vidhiii1711:
Hiii @adamzap 
I have resolved the merge conflicts and all CI checks are passing now. Could you please review it?

checklists/tests/test_models.py
```diff
@@ -442,14 +438,10 @@ def test_render_checklist_simple(self):
             )
 
         with self.subTest(task="Stub release notes added"):
-            self.assertStubReleaseNotesAdded(
-                checklist.latest_release, checklist_content
-            )
```
ulgens, line 447: For this type of "It was broken into multiple lines when the line length was enough to keep in one" cases with multiple items, my recommendation is adding a trailing comma so the original intention - keeping the args seperate from the call - will be kept. Nitpick tho.

blog/migrations/0008_entry_is_searchable.py
ulgens, line 1: Why are we generating a new migration as a result of linting?

checklists/models.py
```diff
@@ -629,8 +629,9 @@ def cve_description(self):
 
     @cached_property
     def cve_html_description(self):
-        return "".join(
-            f"<p>{line.strip()}</p>" for line in urlize(self.cve_description).split("\n")
+         text = re.sub(r'`([^`]+)`', r'<code>\1</code>', self.cve_description)
```
ulgens: Which rule was violated here?

ulgens: I checked the related commits too and this looks like a bugfix more than linting. If the case is valid, my recommendation is transferring related changes into a bugfix focused PR.

ulgens (changes requested):
@vidhiii1711 Thanks for the updates. I added a couple of new comments. Also, please check the already existing review comments and change requests.

.pre-commit-config.yaml
```diff
@@ -55,15 +49,12 @@ repos:
...
-      hooks:
-          - id: isort
-
-    - repo: https://github.com/pycqa/flake8
-      rev: 7.3.0
+    - repo: https://github.com/astral-sh/ruff-pre-commit
+      rev: v0.4.10
       hooks:
-          - id: flake8
+          - id: ruff
```
marksweb, line 55: This should be `ruff-check`I think, for clarity, even if they're maintaining the old command.

.pre-commit-config.yaml
marksweb: Technically `ruff` and `ruff-check` both run the linter. But I agree the explicit command should be used.

checklists/tests/test_models.py
vidhiii1711: Thanks for the nitpick! I'll add the trailing comma.

blog/migrations/0008_entry_is_searchable.py
vidhiii1711: You're right, the migration snuck in unintentionally while resolving conflicts. I'll remove it from this PR.

checklists/models.py
vidhiii1711: Good point — looking at it again, the cve_html_description change is indeed a bugfix (backtick conversion was broken) rather than a linting fix. I'll move it to a separate bugfix PR as you suggested."

.pre-commit-config.yaml
vidhiii1711: Makes sense! I'll update it to ruff-check for clarity.

blog/migrations/0008_entry_is_searchable.py
vidhiii1711: > Why are we generating a new migration as a result of linting?

I removed the migration as suggested, but the tests are now failing with column blog_entries.is_searchable does not exist because the field exists in the model but has no migration. It seems this field was introduced during the merge conflict resolution. Should I remove the is_searchable field from the model entirely and keep it out of this PR, or is it okay to keep the migration since it's needed for tests to pass? Happy to follow whatever approach you prefer!"

vidhiii1711:
Reverted ruff-check back to ruff since ruff-check isn't available in v0.4.10. Open to upgrading the version if the team prefers explicit ruff-check.

adamzap:
It looks like this PR was superseded by #2775. We have other PRs open to enable more `ruff` options. Please reopen it if I'm misunderstanding.

PR #2516, opened 2026-02-16
Improve semantic HTML and screen reader support for pagination

mahicodes28:
- Replaced incorrect role="navigation" usage with a semantic <nav> element.

- Ensured <ul> contains only <li> elements.
 
- Moved the current page indicator inside a <li> to fix invalid HTML structure.
 
- Added descriptive aria-label attributes to previous/next links to improve screen reader context.
 
- Preserved Django i18n ({% translate %} / {% blocktranslate %})

### DIFF
--- a/docs/templates/docs/search_results.html
+++ b/docs/templates/docs/search_results.html
@@ -82,45 +82,45 @@
     <div class="pagination">
       <nav aria-label="{% translate 'Pagination' context 'pagination' %}">
         <ul class="nav-pagination">
-    
+
           {% if page.has_previous %}
             <li>
               <a rel="prev"
                  class="previous"
                  href="{% querystring page=page.previous_page_number %}"
                  aria-label="{% blocktranslate with page_number=page.previous_page_number num_pages=page.paginator.num_pages trimmed %}
-                   Previous page, Page {{ page_number }} of {{ num_pages }}
-                 {% endblocktranslate %}">
+                               Previous page, Page {{ page_number }} of {{ num_pages }}
+                             {% endblocktranslate %}">
                 <i class="icon icon-chevron-left"></i>
                 <span class="visuallyhidden">{% translate "Previous" context "pagination" %}</span>
               </a>
             </li>
           {% endif %}
-    
+
           <li class="page-current">
             {% blocktranslate with page_number=page.number num_pages=page.paginator.num_pages trimmed %}
               Page {{ page_number }} of {{ num_pages }}
             {% endblocktranslate %}
           </li>
-    
+
           {% if page.has_next %}
             <li>
               <a rel="next"
                  class="next"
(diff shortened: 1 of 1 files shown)


mahicodes28:
Hi! Just a gentle follow-up on this PR.
Please let me know if any changes or improvements are required. I'd be happy to update it.
@alex

docs/templates/docs/search_results.html
```diff
@@ -79,30 +79,48 @@ <h2 class="result-title">
         {% endfor %}
       </dl>
     </div>
+    <div class="pagination">
+      <nav aria-label="{% translate 'Pagination' context 'pagination' %}">
```
smallsaucepan, line 83: `<nav>` can do layout directly so doesn't need a div wrapper. These could be rolled together.

docs/templates/docs/search_results.html
```diff
@@ -79,30 +79,48 @@ <h2 class="result-title">
...
+        <ul class="nav-pagination">
```
smallsaucepan, line 84: The `<ul>` and `<li>`s can also be removed. Navigation can just be a couple of links.

docs/templates/docs/search_results.html
```diff
@@ -79,30 +79,48 @@ <h2 class="result-title">
...
-            </a></li>
+            <li>
+              <a rel="prev"
+                 class="previous"
+                 href="{% querystring page=page.previous_page_number %}"
+                 aria-label="{% blocktranslate with page_number=page.previous_page_number num_pages=page.paginator.num_pages trimmed %}
+                               Previous page, Page {{ page_number }} of {{ num_pages }}
+                             {% endblocktranslate %}">
+                <i class="icon icon-chevron-left"></i>
+                <span class="visuallyhidden">{% translate "Previous" context "pagination" %}</span>
```
smallsaucepan, line 95: Don't need these hidden spans any more now there is an aria-label on the `<a>`.

docs/templates/docs/search_results.html
```diff
@@ -79,30 +79,48 @@ <h2 class="result-title">
...
+              </a>
+            </li>
           {% endif %}
-          <span class="page-current">
+
+          <li class="page-current">
```
smallsaucepan, line 100: A `span` is fine for this.

smallsaucepan (changes requested):
Hi @mahicodes28. Have added some comments on how to simplify the layout further. It should be possible to get it down to:

```
<nav>
  <a prev>
    <i prev />
  </a>
  <span current />
  <a next>
    <i next />
  </a>
</nav>
```

That makes it very smooth for a screen reader because it doesn't need to read out counts of list items, etc.

Also, as noted on the related issue, this occurs on the search results page and the community feeds pages, in case you wanted to fix it there too 👍

PR #2528, opened 2026-02-27
Replace collapsing list component with `details`

### LINKED ISSUE #2520: Decide on the future of the collapsing list component
Before I refactor [the JavaScript for the collapsing list component](https://github.com/django/djangoproject.com/blob/main/djangoproject/static/js/mod/list-collapsing.js), I want to make sure we that we still want to use it on the site.

This component gives animated collapsing functionality to lists and adds "Expand All" and "Collapse All" buttons.

Image

The collapsing list component is used in four places:

1. The "Intro to Django" section of [Getting Started with Django](https://www.djangoproject.com/start/)
    - A standard use of the component
2. The "Django RSS Feeds" section of [Community](https://www.djangoproject.com/community/)
    - A standard use of the component
3. The "Archives" section of the [the blog](https://www.djangoproject.com/weblog/2026/) sidebar
    - Customized to hide the "Expand All" and "Collapse All" buttons
4. The bottom of [the Style Guide](https://www.djangoproject.com/styleguide/)
    - This use doesn't count

So it's used, but not much.

I think these are our options:

1. Keep the component as-is and refactor the JavaScript
2. Replace it with simpler, more specific JavaScript
    - Remove the animation?
    - Remove the "Collapse All" button?
    - Etc.
3. Remove the JavaScript altogether and use [the `details` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/details)
    - This would probably require a bit of CSS

I'm curious what other WG members think. I will be happy to to implement any of the above or another idea.

adamzap:
The collapsing list component was not used much and required non-trivial JavaScript. This patch replaces all uses of it with the `details` and `summary` elements. It has also been removed from the style guide.

Closes #2520

I think this is a good change, but this PR needs some design help:

## Start page - Intro to Django

Before:

image

After:

image

## Community - Django RSS Feeds

Before:

image

After:

image

## News - Sidebar - Archives

Before:

image

After:

image

### DIFF
--- a/djangoproject/templates/aggregator/index.html
+++ b/djangoproject/templates/aggregator/index.html
@@ -66,34 +66,34 @@
   </div>
 
   <h2>{% translate "Django RSS feeds" %}</h2>
-    {% for feedtype, latest_feeds in feedtype_list %}
-      <details>
-        <summary>
-          <h3 class="bullet-icon"><i class="icon icon-rss blue"></i> {{ feedtype.name }}</h3>
-        </summary>
-        <dl class="list-links">
-          {% for item in latest_feeds %}
-            <dt><a href="{{ item.link }}">{{ item.title }}</a></dt>
-            <dd>
-              {% blocktranslate trimmed with modified_date=item.date_modified|date:"DATETIME_FORMAT" public_url=item.feed.public_url title=item.feed.title %}
-                {{ modified_date }} by <a href="{{ public_url }}">{{ title }}</a>
-              {% endblocktranslate %}
-            </dd>
-          {% endfor %}
-        </dl>
-        <p class="meta">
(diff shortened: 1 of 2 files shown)


SaptakS:
I can take a look at implementing the styling to make it look closer to what we had before. I think I would still prefer to have the "expand all/collapse all" button progressively added through JS, so that if someone is not using JS, the individual expand/collapse still work, but if someone does use JS, they get that extra feature (yay progressive enhancement!).

Thoughts @adamzap ?

adamzap:
@SaptakS I'm good with that. Do you think we need the expand/collapse buttons on the blog sidebar though? Those seem a bit out of place to me.

PR #2421, opened 2025-12-29
Added esupgrade pre-commit hook

codingjoe:
> https://fosstodon.org/@codingjoe/115803674715881666
> @pauloxnet like this?

* Add esupgrade to pre-commit hooks
* pre-commit CI applies safe changes
* small unrelated changes, co-pilot discovered
* set explicit console log level (debug)

PR #2537, opened 2026-03-04
Add vitest browser test framework

codingjoe:
First step towards a robust test suite to enable JavaScript changes with confidence.
I chose the cookie functions as PoC.

@django/django-website was this what you had regarding 
["We need to test the changed javascript for regressions"](https://github.com/django/djangoproject.com/pull/2421#issuecomment-3997747430)?

I didn't add Edge as a test browser since Edge and Chromium are engine-wise almost identical.


codingjoe:
Ref #2325 @ulgens I added codecov here, mainly to test it and to make sure I add tags. I can drop the changes, but someone needs to remember to tag both languages.

sarahboyce:
We recently added Playwright tests. I just opened a pr to show how we could add some tests here: https://github.com/django/djangoproject.com/pull/2623
@codingjoe Do you still feel this is needed?

codingjoe:
> We recently added Playwright tests. I just opened a pr to show how we could add some tests here: #2623 @codingjoe Do you still feel this is needed?

Yes, I do believe it's needed. I want to add unit tests, not integration tests. The latter is done via Playwright but comes at a performance and reliability cost. Besides, writing unit tests is a more rewarding experience in most cases.

We recently added the same framework to django-debug-toolbar too, and it serves us well.

adamzap:
There's not much JavaScript in this project currently, and there should be less as time goes on. My preference would be to invest in Playwright tests now and finish the in-progress JavaScript refactor before evaluating the need for JavaScript unit tests.

codingjoe:
> There's not much JavaScript in this project currently, and there should be less as time goes on. My preference would be to invest in Playwright tests now and finish the in-progress JavaScript refactor before evaluating the need for JavaScript unit tests.

While I don't believe there will be less JavaScript in the future, I only added this to provide tests for the fix of the `setCookie` utility, link in the pr description.

If you say I don't have to, I won't. However, a refactor may benefit from unit test. I would use Playwright with the same caution, as Selenium. I love Python, but not to test JavaScript.

marksweb:
I'm not against more things to help us test. And I'm sure there will always be some JS that could be tested.

adamzap:
The working group discussed this pull request, and we'd like to close it for now. We'll see how far we can get with just Playwright. Thanks for your effort here @codingjoe!

PR #2538, opened 2026-03-04
Add yamlfmt pre-commit hook

codingjoe:
I noticed that the YAML indentation is all over the place. There's a handy linter for that.


codingjoe:
cc @ulgens since we previously taked about it

Issue #2535, opened 2026-03-04
Inconsistent focus indicators in website footer


erosselli:
The website footer has two different issues related to focus indicators:
1. The Django logo on the very left of the footer has no focus indicator
2. The "Hosting by In-kind donors" link has a focus indicator of a different color than the surrounding ones 

<img width="1438" height="154" alt="django logo focused but no visible focus indicator" src="https://github.com/user-attachments/assets/8e89d71b-0909-4003-88ab-2dd1f64c4ecd" />
<img width="1440" height="146" alt="focus indicator for 'hosting by in-kind donors' link, in a light green color" src="https://github.com/user-attachments/assets/38d506b6-19c8-4c0e-bf19-ac94aaade8e4" />
<img width="1436" height="149" alt="focus indicator for all other surrounding elements, in a white color" src="https://github.com/user-attachments/assets/dc82a393-596c-447f-a3a2-8bde89f4ae8c" />

Team Mars will work on this as part of Djangonaut Space Session 6.

Mr-Sunglasses:
Hey, Eli could you please assign it to me, I would like to work on it.

thearyamann:
Hi, I noticed this issue was assigned yesterday.
If it becomes available or needs additional help, I would be happy to contribute.

PR #2540, opened 2026-03-05
fix: keyboard focus visibility for footer logo and in-kind-donors links

Mr-Sunglasses:
fix: #2535 

After changes:

keyboard focus visibility around django logo in footer

consistent keyboard focus visibility for in-kind-donors with respect to the others

### DIFF
--- a/requirements/common.txt
+++ b/requirements/common.txt
@@ -8,7 +8,7 @@ django-push @ git+https://github.com/brutasse/django-push.git@22fda99641cfbd2f30
 django-read-only==1.21.0
 django-recaptcha==4.1.0
 django-registration-redux==2.13
-Django==6.0.2
+Django==6.0.3
 docutils==0.21.2
 feedparser==6.0.12
 Jinja2==3.1.6


djangoproject/scss/_style.scss
```diff
@@ -1010,6 +1010,10 @@ footer {
             background-position: left center;
             margin-right: 0;
             margin-top: 20px;
+
+            &:focus {
+                outline: 2px solid #fff;
```
Skyiesac: It might be better to use `var(--white-color)` instead of hardcoding `#fff`, so it stays consistent with the existing color variables and is easier to maintain.

Mr-Sunglasses: great catch @Skyiesac 👏, I've used `var(--menu)` as keyboard focus visibility remains in same color (white) when switching dark and light mode, while in `var(--white-color)` it changes to black color when in dark mode.

Thanks for the suggestion, it is really helpful 💜

Mr-Sunglasses:
> This looks great ⭐ Just a suggestion - according to Django contributor guidelines , the commit message should be in past tense and end with a full stop. :)

Thanks @Skyiesac for the review, I've made the suggested changes.

PR #2542, opened 2026-03-06
Add Django template linguist instruction

codingjoe:
Same as on the django repo, this patch highlight template tags and filters.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,3 @@
 # shell scripts need LF endings to work
 *.sh text eol=lf
+*.html linguist-language=django


JohananOppongAmoateng (commented):
Can we scope this to the templates folders

codingjoe:
> Can we scope this to the templates folders

We don't do so here. I also don't see the harm, since I wouldn't know what HTML files would include Django template syntax without being a template: https://github.com/django/django/blob/476e5def5fcbcf637945985a23675db0e1f59354/.gitattributes#L8

sarahboyce (approved):
Thank you!
I scanned through the `.html` files and I think they'd all benefit so I don't think there's a need to scope

ulgens (approved):

Issue #2554, opened 2026-03-10
Accessibility issue with Django community images .


Skyiesac:
In the [community](https://www.djangoproject.com/community/) endpoint , 
Images in community page lacked proper labeling for screen readers, due to which screen reader reads it as : 

Screen reader: "Unlabeled image, Forum - Post a question" (confusing)

Screenshots:

<img width="786" height="540" alt="Forum icon error" src="https://github.com/user-attachments/assets/4bd581a8-ad85-49a6-b95d-d040b7d69c7b" />

<img width="724" height="542" alt="Another icon error" src="https://github.com/user-attachments/assets/f5e926bc-a7b2-4549-b541-8bca15d7d0e4" />

Darshankg18:
Hi, I'm interested in working on this issue. I'm a Python/Django developer and would like to contribute this as part of my GSoC 2026 application. I'll start working on a fix shortly. Please assign this to me if possible.

PR #2555, opened 2026-03-10
Fixed the missing labels for community images.

Skyiesac:
Fixes: #2554 

Added missing labels for Community page images to make it clear for screenreaders .

Forum icon

Alternate way:

1. if these images are completely decorative ,  `<div aria-hidden="true">`  can be added to hide them completely from screen readers .

### DIFF
--- a/djangoproject/templates/aggregator/index.html
+++ b/djangoproject/templates/aggregator/index.html
@@ -9,7 +9,7 @@
   <div class="community-cta-wrapper">
     <a href="https://forum.djangoproject.com/" class="community-cta-a">
       <div class="community-cta">
-        <div>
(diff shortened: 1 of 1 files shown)


erosselli:
Hello! I think we might be able to consider these images decorative since there's already links with text in them that clearly explain what the link is. I'm tagging @SaptakS how knows more about this than I do to confirm if this would be a better approach :)

SaptakS:
I think I agree with @erosselli that these images (which are technically just icons whose purpose is defined in the links below) can be made decorative by adding `aria-hidden="true"`

Skyiesac:
@erosselli @SaptakS  Thankyou for your review. I've pushed the new changes.

PR #2586, opened 2026-04-08
Allowed creation of site-wide banners via the admin interface.

### LINKED ISSUE #1550: Make it easier to display a site-wide banner

nessita:
Fixes #1550.

A banner requires a title, and optionally a plaintext body and CTA label/URL. A banner can be active or inactive, and only one can be active at a time.

This initial implementation of the feature is intentionally scoped to a minimum to provide something robust but flexible to start using, and to gather feedback for future improvements.

Future improvements could include:

- Active since and until dates.
- Flexible CTA URL handling, such as URL names or local URLs.
- Rich text for the body.


foundation/models.py
```diff
@@ -186,6 +186,29 @@ def __str__(self):
...
+    Only one banner can be active at a time; activating a banner automatically
+    deactivates all others.
+
+    """
+
+    title = models.CharField(max_length=512)
+    body = models.TextField(blank=True)
+    cta_label = models.CharField("Call-to-action label", max_length=100, blank=True)
+    cta_url = models.URLField("Call-to-action URL", blank=True)
+    is_active = models.BooleanField(default=False)
```
sarahboyce:
```suggestion
    is_active = models.BooleanField(default=False, db_index=True)
```
Maybe worth adding as we would query against this a bit

djangoproject/templates/foundation/active_banner.html
```diff
@@ -0,0 +1,9 @@
+{% if banner %}
+  <div class="banner">
+    <p>{{ banner.title }}</p>
+    {% if banner.body %}<small>{{ banner.body|linebreaksbr }}</small>{% endif %}
+    {% if banner.cta_label and banner.cta_url %}
+      <a id="banner-cta" class="cta" href="{{ banner.cta_url }}">{{ banner.cta_label }}</a>
+    {% endif %}
+  </div>
```
sarahboyce: If we want to support full flexibility. We may need to be able to render html.

The previous PyCharm banner was like this:
```html
<div class="banner">
  <p>
    <a href="https://www.jetbrains.com/pycharm/promo/support-django/?utm_campaign=pycharm&utm_content=django25&utm_medium=referral&utm_source=dsf-banner">
      Extended until November 19, 2025, <em class="underlined">get PyCharm for 30% off</em>.
      All money goes to the <em>Django Software Foundation</em>!
    </a>
  </p>
</div>
```
Catherine specifically asked we updated it to make the whole banner a clickable link (previously was just part of it). She also asked to make the get PyCharm underlined (which previously was the link).

I later made a banner like:
```html
<div class="banner">
  <p>Help us reach our 2025 fundraising goal!</p>
  <small>
    Donate to the <em>Django Software Foundation</em> to fund Django's development, security, and community events!<br/>
    If you got value out of using Django this year, please consider making a donation &#9829;
  </small>
  <a class="cta" href="{% url 'fundraising:index' %}">Donate today</a>
</div>
```
The only thing I wouldn't be able to do is emphasize "Django Software Foundation"

This is maybe what the PyCharm banner could look like with the CTA button 

Image

What do you think?

djangoproject/templates/base.html
```diff
@@ -59,7 +59,7 @@
     </section>
 
     <div id="billboard">
-      {% block billboard %}{% endblock %}
+      {% block billboard %}{% active_banner %}{% endblock %}
```
sarahboyce, line 62: Not for this PR, and unsure on our caching strategy anyway, but we could consider add caching to this template fragment

sarahboyce (commented):
This looks great! Thank you for pulling this together ❤️ 

I mainly have a question as to whether we need the flexibility to render html. It's full flexibility but hard to "test" if it looks nice from the admin (folks should test locally before updating). But we cannot 100% replicate our previous banners without it 🤷

djangoproject/templates/foundation/active_banner.html
ryancheley: @sarahboyce  @nessita this looks great! do you happen to have a dark mode version of the screenshot as well?

edit: added Sarah

djangoproject/templates/base.html
nessita: I agree this is a nice to have and we can add it in a second iteration. Will create a ticket once this is merged.

priyapahwa:
Thanks for putting this together. Really handy utility. Appreciate it!

foundation/models.py
priyapahwa: Should we include a `help_text` on `is_active` field to make it explicit that only one banner can be active at a time?

nessita: Thanks for the suggestion! The only query against this field is in the template tag, and given the `Banner` table is expected to stay very small (less than a dozen of rows?), the index overhead would likely outweigh any benefit. I think we could revisit if that assumption ever changes.

nessita: Good idea! Will add a help text.

djangoproject/templates/foundation/active_banner.html
nessita: @sarahboyce Thank you for all those details, they are extremely helpful. I'll push a new commit with:

* a way for users with view perm on Banner to see the most recent banners rendered, even if inactive
* metadata in banners about who created it and when, also last updated
* for full flexibility, I'll follow what other parts of the projects are doing by allowing "free form html" via text field + `|safe` in templates. Given this is configured from the admin, the risk of exposing an XSS attack or similar are minimal IMHO.

What's left to decide is whether to support only free form HTML -- leaving it to the author to use the correct CSS classes (example `<a class="cta" href="...">`), or something slightly smarter: if `cta_url` is set and `cta_label` is also set, render a CTA button. If `cta_url` is set but `cta_label` is blank, make the whole banner a clickable link.

sarahboyce: Both are great options, so I would vote for whichever you prefer to impliment 👍

I will add a dark mode screenshot on Monday but I would class that as global banner styling which can be tweaked in furture PRs. This is scoped around the admin interface to activate banners without code changes and deploments being needed. However, we need to decide the flexibility of the banner specific dynamic content and the styling of such (if that makes sense)

sarahboyce: This is the banner on dark mode, it doesn't change its formatting:
Screenshot from 2026-04-12 12-14-39

ryancheley: @sarahboyce looks amazing. Great work to you and @nessita on this!!!

nessita:
I have applied feedback from all comments. Key updates in this revision:
* Raw HTML is allowed in banner body.
* Preview for authorized users under `/foundation/banners/<pk>/preview/`
* Call-to-action button: After testing the "whole banner as link" option, I decided to not include it. Without an explicit button there is no visual indication that the banner is clickable. The supported structure is now:
  - A mandatory title
  - An optional body (HTML supported)
  - An optional CTA button, which requires both a label and a URL (enforced via model validation and surfaced in the admin with help text)
  
EDIT: on a second thought, I implemented a simpler banner preview option by providing an authenticated endpoint. See recorded gif:

banner-djangoproject

foundation/admin.py
```diff
@@ -52,6 +52,35 @@ class ActionItemInline(admin.StackedInline):
...
+    list_display = ["title", "is_active", "created_at", "updated_at"]
+    list_editable = ["is_active"]
+    readonly_fields = ["created_by", "created_at", "updated_by", "updated_at"]
+    fieldsets = [
+        (None, {"fields": ["title", "body", "is_active"]}),
+        (
+            "Call to action",
+            {
+                "description": (
+                    "Both fields should be define for the CTA button to be displayed."
```
sarahboyce:
```suggestion
                    "Both fields should be defined for the CTA button to be displayed."
```

foundation/models.py
```diff
@@ -186,6 +188,84 @@ def __str__(self):
...
+        editable=False,
+        on_delete=models.SET_NULL,
+        related_name="+",
+    )
+
+    def __str__(self):
+        return self.title
+
+    def get_absolute_url(self):
+        return reverse("foundation_banner_preview", kwargs={"pk": self.pk})
```
sarahboyce, line 255:
```suggestion
        from django_hosts.resolvers import reverse as reverse_with_host
        return reverse_with_host("foundation_banner_preview", kwargs={"pk": self.pk}, host="www")
```

Testing locally with docker, I need the host for the "View on site" link to work in the admin

sarahboyce (commented):
Thank you @nessita ! I love love love it ❤️
A couple of minor tweaks and then I think it's ready ⭐

foundation/admin.py
nessita: Great catch! fixed

foundation/models.py
nessita: I have mixed feelings about this because it breaks the non-docker use case, but I'll incorporate it since it's consistent with the rest of the code. Thanks for pointing this out!

jefftriplett (approved):
Looks great. Can we ship it?

jacobian:
Looks fantastic to me too.

djangoproject/templates/foundation/banner.html
```diff
@@ -0,0 +1,9 @@
+{% if banner %}
+  <div class="banner">
+    <p>{{ banner.title }}</p>
```
sabderemane: This should be a heading (h2) not a paragraph, so it show in the accessibility tree with headings. Style should be modified accordingly.

CuriousLearner (approved):
LGTM 🌮

sabderemane (approved):
Thank you @nessita for this work! 💟 

I still have a comment related to the semantic of the title of the banner that I would highly recommend to change but I don't want to block the PR due to its urgency.

This could be changed in a second time if it's not in this PR, up to you to decide.

djangoproject/templates/foundation/banner.html
nessita: Thank you! this makes total sense and sorry I missed it -- I think I was copying from the previous banner examples we had. I'm fixing and pushing!

nessita: I also made the tests more strict for this so we don't get a regression.

PR #2599, opened 2026-04-17
[checklists] Updated SecurityIssue model to track CVSS 3.1 and 4.0 scores.

### LINKED ISSUE #2592: [checklists] Compile CVSS score & vector into CVE JSON
We have model fields to capture a CVSS score & vector already, but we should compile it to the CVE JSON so that it's included with our next submissions.

nessita:
This work also drops all unused fields specifics to the CVSS score.

Fixes #2592.


jacobtylerwalls (approved):
LGTM, thanks!

PR #2129, opened 2025-07-16
style(docs): improve table readability

WajahatKanju:
### Summary

This PR improves the readability and visual appeal of tables in Django documentation. Specifically, it addresses the following  issue( #2128):

- Table headers now have clearer separation via borders and background.
- Added padding for cells to reduce visual cramping.
- Introduced alternating row colors and a hover effect for better scanning.
- Applied a subtle shadow and bottom margin to give tables space and structure.

https://github.com/user-attachments/assets/fa182638-56b6-46e9-9b8c-d668c220cbc6

https://github.com/user-attachments/assets/4af8e212-c5e1-4734-a970-7da49a689e3b

Issue #2128, opened 2025-07-11
Improve styling of tables in docs


bmispelon:
I noticed today that [some pages](https://docs.djangoproject.com/en/dev/topics/http/shortcuts/#django.shortcuts.redirect) in the documentation use tables:

<img width="376" height="273" alt="Image" src="https://github.com/user-attachments/assets/3c13472c-cad0-489e-ba04-940bf8933db8" />

I find the styling a bit hard to read:

1) The table headers are not visually separated (no border, no horizontal margin)
2) The table has no bottom margin

blighj:
I'd agree. In a lot of cases, I find the tables easier to read in the original rst format, that's a bad sign.

### List of tables in docs 
- <details><summary>faq/install/#what-python-version-can-i-use-with-django</summary><br><a href="https://docs.djangoproject.com/en/dev/faq/install/#what-python-version-can-i-use-with-django">faq/install/#what-python-version-can-i-use-with-django</a><img width="480" height="400" alt="Image" src="https://github.com/user-attachments/assets/3429aed5-6f9b-48b0-b081-a3dc19cdb95d" /></details>
- <details><summary>ref/contrib/admin/#django.contrib.admin.ModelAdmin.search_fields</summary><a href="https://docs.djangoproject.com/en/dev/ref/contrib/admin/#django.contrib.admin.ModelAdmin.search_fields">ref/contrib/admin/#django.contrib.admin.ModelAdmin.search_fields</a><br><img width="200" height="335" alt="Image" src="https://github.com/user-attachments/assets/cc55c1b0-8fc3-42cf-b0ee-b1f81da21f74" /></details>
- <details><summary>ref/contrib/admin/#reversing-admin-urls (3)</summary><a href="https://docs.djangoproject.com/en/dev/ref/contrib/admin/#reversing-admin-urls">ref/contrib/admin/#reversing-admin-urls</a><br><img width="774" height="627" alt="Image" src="https://github.com/user-attachments/assets/2a73434c-912b-47ef-bdb4-59768d134a4e" /><hr><img width="697" height="419" alt="Image" src="https://github.com/user-attachments/assets/f742ff40-3b46-4178-b67f-d3112ac33602" /><hr><img width="556" height="159" alt="Image" src="https://github.com/user-attachments/assets/17f24c1a-c039-488c-ab0e-6e7131eecd7c" /></details>
[...]

PR #2603, opened 2026-04-25
Improved styling of tables in docs

blighj:
Addresses #2128, PR is a follow on from #2129. The various comments have been addressed.

For the row highlight I went with a color-mix approach, making it 5% darker/lighter for even rows and 10% for hover.
That is less colorful than what was there from the previous PR and my original prototyping. I'm still happy with it.

Color-mix is supported since May 2023 according to caniuse and mdn. I put in a fallback for the even row highlight, but the hover is more additive so I left it out.

image

image

It's subtle in darkmode, but its a tradeoff with color contrast.
Currently the dark-mode hover highlight is already slightly outside AAA. If we aim for that, I'd recommend droping the hover highlight in darkmode, like I said above, additive.

If you are reviewing this, worth checking out how it looks in places with lots of tables near each other like 
/en/dev/ref/templates/builtins/#floatformat

and somewhere with an example box (which has a similar background) right above a table
/en/dev/ref/templates/builtins/#yesno

### DIFF
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -2633,6 +2633,30 @@ table.foundation td {
     padding: 0 5px;
 }
 
+table.docutils {
+    margin-bottom: 1.5em;
+
+    td,
+    th {
+        border-bottom: 1px solid var(--hairline-color);
+        padding: 8px 12px;
+    }
+
+    thead th {
+        border-bottom: 2px solid $gray-medium;
+        font-weight: bold;
+    }
+
+    tr:nth-child(even) {
+        background-color: var(--sidebar-bg);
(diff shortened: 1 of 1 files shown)


ronnzw:
Thank you very much for your PR, I'm not sure if you are aware that we have plans to design the [website](https://www.djangoproject.com/weblog/2026/apr/27/its-time-to-redesign-djangoprojectcom/). I wouldn't want to see such a fantastic work not being utilised. @sabderemane might be in a better position to advice.

sabderemane:
Late reply, I'm sorry but I think that make sense to get this on before the redesign phase, at least it will be used until the redesign and it's definitely an improvement ⭐

PR #2604, opened 2026-04-27
Update initial header level in BLOG_DOCUTILS_SETTINGS to 1

sabderemane:
This will allow to get correct headings levels in the website and review content headings to match the change.


SaptakS (approved):
Makes sense to me with the current structure of blog articles in both rst and markdown.

PR #2607, opened 2026-04-29
fixing formatting on Code of Conduct translate tags

dryan:
I noticed that the `{% translate %}` tags on a few of the metadata blocks we're wrapping across lines so the tag isn't processed. This leaves our `<title>` as `
<title>{% translate
--
"Django Code of Conduct" %} \| Django</title>` instead of `<title>Django Code of Conduct</title>`

### DIFF
--- a/djangoproject/templates/conduct/index.html
+++ b/djangoproject/templates/conduct/index.html
@@ -8,554 +8,554 @@
 
 {% block content %}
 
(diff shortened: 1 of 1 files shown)


SaptakS (approved):
Looks good to me. Thanks for fixing.

enaden711 (approved):

PR #2609, opened 2026-05-01
Update initial header level in BLOG_DOCUTILS_SETTINGS to 2

SaptakS:
Sets initial header level for rst to 2 since it always starts with heading level 1 automatically but keeps markdown baselevel to 1 since markdown has actual heading level applied.


sabderemane (approved):
LGTM! 🔥

PR #2612, opened 2026-05-06
[checklists] Minor improvements following recent security release

sarahboyce:
Code block formatting was improved with the following:
- Adding the `markdown.extensions.fenced_code` extension
- Ensuring we don't render the code block indented a bullet point, this meant that the immediately following bullet must not be indented
- Adding some styling as a nice-to-have


checklists/admin.py
```diff
@@ -143,7 +143,6 @@ def get_ordering(self, request):
         return [
             "-updated_at",
             "-created_at",
-            *cve_sort_key(desc=True),
```
sarahboyce, line 146: As the `cve-year-number` default is the following:
```python
def get_cve_default():
    return f"CVE-{datetime.date.today().year}-{get_random_string(5)}"
```

The random string is not numbers. This means saving with this dummy number breaks the admin

sarahboyce (commented):
.

checklists/admin.py
jacobtylerwalls: Can we keep the sorting and fix the default e.g. `randint(1000, 100_000)`?

sarahboyce: Is there a risk that might be a real CVE number?

jacobtylerwalls: Ohhh, got it. So we want it to be broken. So why this change to the sorting, then?

sarahboyce: Because if you save anything with the dummy number, let's say by accident, then the admin changelist breaks because this ordering does not handle this case

jacobtylerwalls: Right, okay. What about just adding a RegexValidator on this field, then? The ordering is nice because if we remove it we get alphabetic ordering, e.g. 11 before 2.

sarahboyce: Updated :+1:

checklists/templates/checklists/_make_release_public.md
```diff
@@ -8,11 +8,8 @@
     - Save
     - Check at: https://www.djangoproject.com/admin/releases/release/{{ release }}/change/
 
-- [ ] Test the release locally with script from `scripts` folder:
-    - `VERSION={{ release }} scripts/test_new_version.sh`
-
-- [ ] Confirm the release signature with script from `scripts` folder:
-    - `VERSION={{ release }} scripts/confirm_release.sh`
+- [ ] Test the release locally and confirm the release signature with script from `scripts` folder:
+    - `VERSION={{ release }} verify_release.sh`
```
jacobtylerwalls: Optional: if you leave scripts/... here, then the following `cd ..` makes sense.

jacobtylerwalls (approved):
Thank you!

checklists/templates/checklists/_make_release_public.md
sarahboyce: I didn't like that it said "from scripts folder" and then the next command says "use commands printed by release script" which half means, ignore these commands to me :thinking: 
Do you have a suggestion of what makes the most sense?

jacobtylerwalls: All I had in mind was:
```suggestion
    - `VERSION={{ release }} scripts/verify_release.sh`
```

Yes, I've been ignoring the commands, so we could spruce up the wording here, but I wouldn't block on it.

sarahboyce: Ok I will change for it to not say "from scripts folder"

sarahboyce: Ahhhhhhhh I think I am finally reading it as I am meant to. I will have a think

sarahboyce: Pushed a tweak, I was reading ```from `scripts` folder``` to mean, be inside the `scripts` folder then run the script, rather than this script is in the `scripts` folder

jacobtylerwalls: Perfect.

PR #2622, opened 2026-05-12
[checklists] Add discovery field

jacobtylerwalls:
This allows producing CVE JSON with "source: discovery: internal" instead of hardcoding "external".

Choices taken from vulnogram.org, ~minus "unknown".~

The vulnogram choices are not part of the spec yet, but looks like they will be:
https://github.com/CVEProject/cve-schema/issues/339


checklists/models.py
```diff
@@ -498,6 +498,11 @@ class SecurityIssue(models.Model):
     )
     reporter = models.CharField(max_length=1024, blank=True)
     remediator = models.CharField(max_length=1024, blank=True)
+    discovery = models.CharField(
+        max_length=128,
+        choices=[(i, i.title()) for i in ("EXTERNAL", "INTERNAL", "USER", "UPSTREAM")],
+        default="EXTERNAL",
+    )
```
sarahboyce, line 511: Image

Optional: we could match the description of `USER` as `During use`. (this would need a migration change) - what do you think?

```suggestion
    discovery = models.CharField(
        max_length=128,
        choices={
            "EXTERNAL": "External",
            "INTERNAL": "Internal",
            "USER": "During use",
            "UPSTREAM": "Upstream",
        },
        default="EXTERNAL",
    )
```

sarahboyce (approved):
Thank you! Just a query on whether we want to match the choice descriptions

checklists/models.py
jacobtylerwalls: Good idea, and in that case I'll throw undefined back in 👍

PR #2623, opened 2026-05-14
Added tests for theme toggle button.

sarahboyce:


djangoproject/templates/includes/toggle_theme.html
```diff
@@ -1,6 +1,6 @@
 {% load i18n %}
 
-<button class="theme-toggle">
+<button class="theme-toggle" data-testid="theme-toggle">
```
adamzap: The `theme-toggle` class is only used on this element, and the JavaScript uses the class name. Could we drop this attribute and use the class name in the tests?

djangoproject/tests.py
```diff
@@ -282,3 +282,39 @@ def test_search_placeholder_mac_mode(self):
...
+        page.get_by_test_id("theme-toggle").click()
+        theme = page.evaluate("document.documentElement.dataset.theme")
+        self.assertEqual(theme, "dark")
+
+    def test_cycle_theme_when_prefers_dark(self):
+        page = self.browser.new_page(user_agent=self.mac_user_agent)
+        page.emulate_media(color_scheme="dark")
+        page.goto(self.live_server_url)
+
+        for expected_theme in ["light", "dark", "auto"]:
```
adamzap, line 312: Would it be worth using `subTest` here and in the test below?

djangoproject/tests.py
```diff
@@ -282,3 +282,39 @@ def test_search_placeholder_mac_mode(self):
...
+        )
+        page.goto(self.live_server_url)
+        theme = page.evaluate("document.documentElement.dataset.theme")
+        self.assertEqual(theme, "dark")
+
+    def test_set_theme_updates_data_attribute(self):
+        page = self.browser.new_page(user_agent=self.mac_user_agent)
+        page.goto(self.live_server_url)
+        page.get_by_test_id("theme-toggle").click()
+        theme = page.evaluate("document.documentElement.dataset.theme")
```
adamzap, line 301: Would it be worth asserting that the theme is `auto` before the button is clicked since that is added by JavaScript?

adamzap (commented):
Thank you for this! A few questions.

sarahboyce:
Thank you for the review! Pushed up changes :+1:

djangoproject/tests.py
adamzap: @sarahboyce Did you make a change related to this? I'm not able to see a diff of your latest changes.

sarahboyce: I added it to a different test to the one you commented on but added to this one now too

adamzap: Thanks!

adamzap (approved):
👍 Looks great!

PR #2628, opened 2026-05-22
Fix dead FAQ link and broken anchors on the fundraising page

ericholscher:
The fundraising page has a "Frequently Asked Questions" link that points back to the page itself, plus two links to in-page anchors that don't exist.

### DIFF
--- a/djangoproject/templates/fundraising/index.html
+++ b/djangoproject/templates/fundraising/index.html
@@ -46,11 +46,9 @@
     </li>
     <li><a href="https://github.com/sponsors/django" target="_blank">{% translate "Sponsor Django via GitHub Sponsors" %}</a>.</li>
     <li>
-      {% url 'fundraising:index' as fundraising_url %}
       {% blocktranslate trimmed %}
-        <a href="{{ fundraising_url }}#benevity-giving">Benevity Workplace Giving
-          Program</a> - If your employer participates, you can make donations to the
-        DSF via payroll deduction.
+        Benevity Workplace Giving Program - If your employer participates, you can
+        make donations to the DSF via payroll deduction.
       {% endblocktranslate %}
     </li>
   </ul>
@@ -91,13 +89,6 @@
       {% endblocktranslate %}
     </li>
   </ul>
-
-  <p>
-    {% blocktranslate trimmed %}
-      Still curious? See our <a href="{{ fundraising_url }}">Frequently Asked Questions</a>
-      about donations.
-    {% endblocktranslate %}
-  </p>
 
   <h2 id="fellowship-program">{% translate "Django Fellowship Program" %}</h2>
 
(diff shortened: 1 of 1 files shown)


djangoproject/templates/fundraising/index.html
```diff
@@ -46,11 +46,9 @@ <h2>{% translate "Other ways to give" %}</h2>
     </li>
     <li><a href="https://github.com/sponsors/django" target="_blank">{% translate "Sponsor Django via GitHub Sponsors" %}</a>.</li>
     <li>
-      {% url 'fundraising:index' as fundraising_url %}
       {% blocktranslate trimmed %}
-        <a href="{{ fundraising_url }}#benevity-giving">Benevity Workplace Giving
-          Program</a> - If your employer participates, you can make donations to the
-        DSF via payroll deduction.
+        Benevity Workplace Giving Program - If your employer participates, you can
```
adamzap: I wonder if this should point to somewhere like [here](https://benevity.com/products/donate)?

There's also a bad link in the footer. However, this may be out of scope.

adamzap (approved):
👍 Looks great! I'd also like to hear from whoever owns the content of the page.

djangoproject/templates/fundraising/index.html
ericholscher: Perhaps. I was mostly trying to clean up obviously broken things. There's a bunch that could be improved, but didn't want to expand the scope so it would be easy to merge.

djangoproject/templates/fundraising/index.html
```diff
@@ -150,7 +141,7 @@ <h2 id="fellowship-program">{% translate "Django Fellowship Program" %}</h2>
   <p>
     {% blocktranslate trimmed %}
       <strong>If you use Django on a daily basis and care about the development of Django
-        itself, you should donate today</strong> (may be <a href="{{ fundraising_url }}#tax-deductible">tax deductible</a>).
+        itself, you should donate today</strong> (may be tax deductible).
```
medmunds: If possible, it would be good to link this to https://www.djangoproject.com/foundation/donate/#tax-deductible (which provides additional information), rather than removing the link altogether.

djangoproject/templates/fundraising/index.html
medmunds: Looking at the history, I'm pretty sure the intended target of this link is https://www.djangoproject.com/foundation/donate/#benevity-giving (in the flatpages, see https://github.com/django/djangoproject.com/issues/2654#issuecomment-4665149623). And that has useful additional info for anyone interested, so might be better than removing the link.

I don't think a link to Benevity's site would be helpful here: it's all targeted to *employers* who would engage Benevity to manage a charitable payroll deduction program. I couldn't find any useful info there for *employees* who want to contribute to a charitable organization through such a program.

Incidentally, the broken footer link was apparently meant to target this `<li>` (though maybe should be repointed directly to the foundation/donate content instead).

ericholscher: Yea, I'm guessing the content was moved at some point, and the links weren't updated? I've updated the PR to point at the right pages -- thanks for the note!

jefftriplett (approved):
LGTM. Nice change. I updated the s/IRC/IRS/ typo in the flat page directly too.

PR #2629, opened 2026-05-22
Link to the corporate membership page, not straight to the signup form

ericholscher:
The main fundraising page links straight to the corporate membership signup form, and the signup form has no link back to the page that explains membership.

People likely want to know what they're getting before signing up. Also `apply` isn't mentioned anywhere other than the primary CTA :)

### DIFF
--- a/djangoproject/templates/members/corporatemember_form.html
+++ b/djangoproject/templates/members/corporatemember_form.html
@@ -9,6 +9,12 @@
       organization!
     {% endblocktranslate %}
   </p>
+  <p>
+    {% blocktranslate trimmed %}
+      See the various sponsorship benefits on our
+      <a href="https://www.djangoproject.com/foundation/corporate-membership/">corporate membership page</a>.
+    {% endblocktranslate %}
+  </p>
   <form class="form-input corporate-membership-join-form" method="post" action="."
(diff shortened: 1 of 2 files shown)


djangoproject/templates/members/corporatemember_form.html
```diff
@@ -9,6 +9,12 @@ <h1>{% translate "Become a DSF corporate member" %}</h1>
       organization!
     {% endblocktranslate %}
   </p>
+  <p>
+    {% blocktranslate trimmed %}
+      See the various sponsorship benefits on our
+      <a href="https://www.djangoproject.com/foundation/corporate-membership/">corporate membership page</a>.
```
ericholscher, line 15: Note that Benefits on the flatpage doesn't have an `id` to link to. 😨 

I have a bunch of copy edits for the flatpage as well, but not sure how to submit those?

jefftriplett: Sorry about that, if this is helpful, I made a gist in case that's a helpful stop gap https://gist.github.com/jefftriplett/ebe87f093d0ad609a52c01a13547ca86

nessita:
This was covered and included in #2833.

PR #2630, opened 2026-05-22
Hide the site banner on foundation pages

ericholscher:
The site-wide banner is a distraction on the foundation pages (corporate membership and the rest). 

I'm less than 100% sure on this PR -- but in general if people want to give you money, the "giving you money" thing should be the primary CTA on the page.


jefftriplett (approved):
Thank you for the contribution!

PR #2632, opened 2026-05-25
[checklists] Fixed CVE description to include product name and render inline code in HTML.

nessita:
`SecurityIssue.cve_description` was missing "Django" before the affected version range (e.g. "An issue was discovered in 6.0 before 6.0.5" instead of "An issue was discovered in Django 6.0 before 6.0.5).

`SecurityIssue.cve_html_description` was passing backtick-delimited inline code through as-is into HTML output. Backticks are now converted to `<code>` tags.

`SecurityIssue.blogdescription` had a tuple help text instead of a str.


sarahboyce (approved):
Looks great ⭐

PR #2648, opened 2026-06-05
[checklists] Added PyPI token revocation to checklist.

jacobtylerwalls:
Happy to move this to whatever template include is best!


nessita (approved):
This looks good. An option to remove duplication would be to move it inside `checklists/_make_release_public.md` saying something like:

- [ ] Generate and store new PyPI token if you haven't already today
- [ ] ... existing stuff ...
- [ ] Revoke PyPI token if no other artifacts are to be pushed today

PR #2652, opened 2026-06-09
Changed dark mode theme switcher to FontAwesome icons.

### LINKED ISSUE #2650: Symbols svg renders as large empty space
#### Summary

The sprite svg that defines symbols used by the dark-mode theme switcher is being rendered toward the bottom of every page, as a large empty block.

#### To reproduce

Go to any page on djangoproject.com, scroll to the bottom, and look for unnecessary empty space above the footer. Or go into devtools and add this rule to highlight the svg:

```css
svg {
  outline: 2px dashed darkorange;
}
```

Image

#### Suggested fix

That svg is rendered in [base.html](https://github.com/django/djangoproject.com/blob/5475f6c0f1baecbfd8e950b2b71b2780c6149dab/djangoproject/templates/base.html#L97-L103) and used only in [includes/toggle_theme.html](https://github.com/django/djangoproject.com/blob/5475f6c0f1baecbfd8e950b2b71b2780c6149dab/djangoproject/templates/includes/toggle_theme.html#L9-L17). There are a few options for fixing it:

1. (Minimal change) Add `style="display:none;"` to the sprite svg. I'd suggest also moving the svg from base.html to toggle_theme.html for better locality of reference. (I ended up [doing both of these](https://github.com/medmunds/django/blob/6e976d325d96835759356057c7258f42ebc9b17a/docs/_theme/djangoproject/toggle_theme.html#L20) in adapting the djangoproject template and themes for django/django#21416.)

2. Inline the icon svg paths directly in the theme picker and remove the separate symbol defs. (These symbols are not reused elsewhere on the site, so a sprite svg is an unnecessary complication.)

3. Switch the theme picker to use FontAwesome icons and remove the symbol svg. This is more in keeping with the [style guide](https://www.djangoproject.com/styleguide/#icons) and could probably simplify the theme picker DOM and JS a bit. The site currently uses [FontAwesome v4.2](https://fontawesome.com/v4/icons/), which has [adjust](https://fontawesome.com/v4/icon/adjust), [moon-o](https://fontawesome.com/v4/icon/moon-o) and [sun-o](https://fontawesome.com/v4/icon/sun-o) icons that are good matches for the symbols defined in the svg sprite.

(I'm happy to submit a PR if one of these options seems reasonable.)

medmunds:
Changed toggle_theme.html button to use FontAwesome icons and removed svg icon sprite from base template.

Fixes #2650. (This is option 3 from that report.)

### DIFF
--- a/djangoproject/scss/_dark-mode.scss
+++ b/djangoproject/scss/_dark-mode.scss
@@ -296,29 +296,21 @@ html[data-theme="dark"] .img-release {
     justify-content: center;
 }
 
-.theme-toggle svg {
-    max-width: 16px;
-    max-height: 16px;
-    display: none;
-}
-
 /* ICONS */
-.theme-toggle svg.theme-icon-when-auto,
-.theme-toggle svg.theme-icon-when-dark,
-.theme-toggle svg.theme-icon-when-light {
-    fill: var(--menu);
-    color: $green-dark;
+.theme-toggle .icon {
+    display: none;
+    color: var(--menu);
 }
 
-html[data-theme="auto"] .theme-toggle svg.theme-icon-when-auto {
+html[data-theme="auto"] .theme-toggle .theme-icon-when-auto {
     display: block;
 }
 
-html[data-theme="dark"] .theme-toggle svg.theme-icon-when-dark {
+html[data-theme="dark"] .theme-toggle .theme-icon-when-dark {
     display: block;
 }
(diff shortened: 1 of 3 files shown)


SaptakS (approved):
Looks good to me.

adamzap (approved):
👍 Looks great! We'll go with this solution.

PR #2656, opened 2026-06-10
Updated link to release roadmap.

jacobtylerwalls:
Follow-up to 8ef4db8

Instead of linking to the Trac wiki, which just links to the new download page, link there directly.

See this link: https://www.djangoproject.com/download/#:~:text=6.1%20development%20process


nessita (approved):
Looks great, thank you! (I would have suggested adding a small test but your call)

PR #2659, opened 2026-06-10
Replaced deprecated name attribute with id in py-modindex.html.

### LINKED ISSUE #2658: py-moduleindex.html uses deprecated name attribute
The [Python module index](https://docs.djangoproject.com/en/6.0/py-modindex/) in the documentation uses `<a name=…>` for internal link targets. The [`name` attribute is deprecated](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/a#name) on the `<a>` tag and should be replaced with `id`.

```html
<h3>
  <a name="d">D</a>
</h3>
```

This is in [docs/templates/docs/py-modindex.html](https://github.com/django/djangoproject.com/blob/cd2964e3a45e02c64c54538fb2a0bc8e184755ae/docs/templates/docs/py-modindex.html#L20). (The similar genindex.html correctly uses `id`, so this was likely an oversight during Sphinx→Django template conversion.)

medmunds:
Fixes #2658.

### DIFF
--- a/docs/templates/docs/py-modindex.html
+++ b/docs/templates/docs/py-modindex.html
@@ -17,7 +17,7 @@
     </ul>
 
     {% for letter, objects in doc.content %}
-        <h3><a name="{{ letter }}">{{ letter|upper }}</a></h3>
+        <h3><a id="{{ letter }}">{{ letter|upper }}</a></h3>
         <ul>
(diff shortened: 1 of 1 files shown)


SaptakS (approved):
Looks good to me. Thanks for the fix!

PR #2665, opened 2026-06-17
Handle corporate member SVG logos

### LINKED ISSUE #2664: Fundraising page returned 500 when it encountered .svg files for corporate logos
`/fundraising` went down today with 500s because of template code calling `obj.thumbnail.width`, which apparently is not safe if the underlying logo is an SVG.

SVGs don't need thumbnails anyway; they're already vectorized.

Catherine deleted the offending SVGs to get the page restored quickly, but we should make the page more fault-tolerant, since I expect we want to accept SVGs (the admin form doesn't allow them due to the quirk of our use of sorl-thumbnail's ImageField -- the corporate membership form uses a vanilla ImageField and does allow them).

jacobtylerwalls:
The `DjangoHeroForm`/`CorporateMemberSignUpForm` accepted SVG logos, but SVG aren't processable via Pillow, so the thumbnail code threw errors when trying to run them through Pillow.

sorl-thumbnail logged this Pillow error:

```py
cannot identify image file <_io.BytesIO object at 0x7965b761e7a0>
```

And the /fundraising page crashed on trying to access the `.width` property on a thumbnail that didn't exist. (The dimension properties in `sorl_thumbnail` are unsafely implemented, accessing `._size[0]` even though `._size` is `None` in a case like this.)

In the template, that resulted in:
```py
'NoneType' object is not subscriptable
```

Closes #2664

### DIFF
--- a/djangoproject/templates/fundraising/includes/top_corporate_members.html
+++ b/djangoproject/templates/fundraising/includes/top_corporate_members.html
@@ -5,7 +5,15 @@
       <div class="clearfix">
         <div class="member-logo">
           <a href="{{ obj.url }}" title="{{ obj.display_name }}">
-            <img src="{{ obj.thumbnail.url }}" alt="{{ obj.display_name }}" />
+            {% if obj.logo %}
+              <img
+                {% if obj.thumbnail %}
+                  src="{{ obj.thumbnail.url }}"
+                {% else %}
+                  src="{{ obj.logo.url }}"
+                {% endif %}
+                alt="{{ obj.display_name }}" />
+            {% endif %}
           </a>
         </div>
         <div class="small-cta">
--- a/djangoproject/templates/members/includes/_member_with_logo.html
+++ b/djangoproject/templates/members/includes/_member_with_logo.html
@@ -1,9 +1,14 @@
 {% load static thumbnail %}
 <li>
-    {% if obj.logo %}
-        <img src="{{ obj.thumbnail.url }}"
-             alt="{{ obj.display_name }}" class="corporate-member-logo" />
-    {% endif %}
-    <h3><a href="{{ obj.url }}">{{ obj.display_name }}</a></h3>
-    <p>{{ obj.description|linebreaksbr }}</p>
+  {% if obj.logo %}
+    <img
+      {% if obj.thumbnail %}
+        src="{{ obj.thumbnail.url }}"
+      {% else %}
+        src="{{ obj.logo.url }}"
+      {% endif %}
+      alt="{{ obj.display_name }}" class="corporate-member-logo" />
+  {% endif %}
+  <h3><a href="{{ obj.url }}">{{ obj.display_name }}</a></h3>
+  <p>{{ obj.description|linebreaksbr }}</p>
 </li>
--- a/djangoproject/templates/fundraising/includes/_hero_with_logo.html
+++ b/djangoproject/templates/fundraising/includes/_hero_with_logo.html
@@ -4,13 +4,13 @@
     {% if obj.url %}<a href="{{ obj.url }}" rel="nofollow">{% endif %}
     {% if obj.logo %}
       <img
-        {% if obj.logo_is_svg %}
-          src="{{ obj.logo.url }}"
-        {% else %}
+        {% if obj.thumbnail %}
           src="{{ obj.thumbnail.url }}"
           srcset="{{ obj.thumbnail.url|resolution:'2x' }} 2x"
           width="{{ obj.thumbnail.width }}"
           height="{{ obj.thumbnail.height }}"
+        {% else %}
+          src="{{ obj.logo.url }}"
         {% endif %}
         loading="lazy"
         alt="{% blocktranslate trimmed with company_name=obj.display_name %}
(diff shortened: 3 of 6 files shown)


nessita (commented):
Thank you for this work! 🌟  Looks good, though there is one thing I'd like to discuss before merging.

Right now `logo_is_svg` is the only thing keeping the page up, and it's an naive extension check, so it has to be "right" every time or we 500 again. An SVG uploaded as `logo.png`, or any file that `get_thumbnail` can not handle (truncated upload, odd format, corrupted file, etc.), still reaches the code and raises. So we've gone from "SVGs crash the page" to "SVGs we happen to identify by filename don't crash the page", which is definitely an improvement but perhaps we could be a bit more robust?

I'd rather the SVG check not be the only thing preventing the crash. I was thikning we could make `thumbnail` handle failures and having the template fallback to the original file:

```python
@property
def thumbnail(self):
    if not self.logo:
        return None
    if self.logo_is_svg:
        return None
    try:
        geometry = f"{self.THUMBNAIL_SIZE}x{self.THUMBNAIL_SIZE}"
        return get_thumbnail(self.logo, geometry, quality=100)
    except Exception:
        logger.warning("Could not thumbnail logo %r", self.logo.name)
        return None
```

Then maybe:

```django
{% if obj.logo %}
  {% if obj.thumbnail %}
    <img src="{{ obj.thumbnail.url }}"
         srcset="{{ obj.thumbnail.url|resolution:'2x' }} 2x"
         width="{{ obj.thumbnail.width }}"
         height="{{ obj.thumbnail.height }}"
         loading="lazy" alt="...">
  {% else %}
    <img src="{{ obj.logo.url }}" loading="lazy" alt="...">
  {% endif %}
{% else %}
  ...
{% endif %}
```

One other thought: this only touches `_hero_with_logo.html`. I think `members/includes/_member_with_logo.html` and `fundraising/includes/top_corporate_members.html` render the same logos and still do a bare `{{ obj.thumbnail.url }}`, so they won't 500 anymore but they'll show a broken image for SVGs. Is it worth applying the same fallback there (while also adding tests so it doesn't regress)?

Image

jacobtylerwalls:
Thanks, great feedback. I applied your suggestions and added one test for the broken thumbnail case. I think I'd like to stop here before investing more time.

djangoproject/templates/fundraising/includes/top_corporate_members.html
```diff
@@ -5,7 +5,14 @@ <h3>{{ header }}</h3>
       <div class="clearfix">
         <div class="member-logo">
           <a href="{{ obj.url }}" title="{{ obj.display_name }}">
-            <img src="{{ obj.thumbnail.url }}" alt="{{ obj.display_name }}" />
+            {% if obj.logo %}
+              {% if obj.thumbnail %}
+                <img src="{{ obj.thumbnail.url }}"
+                   {% else %}
+                     <img src="{{ obj.logo.url }}"
+                   {% endif %}
```
nessita: I know you said no more changes but this indentation is unexpected and arguably confusing, would you consider:

```suggestion
              {% if obj.thumbnail %}
                <img src="{{ obj.thumbnail.url }}"
              {% else %}
                <img src="{{ obj.logo.url }}"
              {% endif %}
```

djangoproject/templates/members/includes/_member_with_logo.html
```diff
@@ -1,9 +1,13 @@
...
-             alt="{{ obj.display_name }}" class="corporate-member-logo" />
-    {% endif %}
-    <h3><a href="{{ obj.url }}">{{ obj.display_name }}</a></h3>
-    <p>{{ obj.description|linebreaksbr }}</p>
+  {% if obj.logo %}
+    {% if obj.thumbnail %}
+      <img src="{{ obj.thumbnail.url }}"
+         {% else %}
+           <img src="{{ obj.logo.url }}"
+         {% endif %}
```
nessita: Same here:

```suggestion
    {% if obj.thumbnail %}
      <img src="{{ obj.thumbnail.url }}"
    {% else %}
      <img src="{{ obj.logo.url }}"
    {% endif %}
```

nessita (approved):
Looks great! Added small optional comment.

djangoproject/templates/fundraising/includes/top_corporate_members.html
jacobtylerwalls: Yeah I wasn't happy about that either, but the formatter chewed it up into this. I shouldn't have tried to change this file from 4 spaces to 2 to match the others, I'll figure something out 😄

jacobtylerwalls: Actually, it was the other file that I changed from 4 to 2, so that's not related.

Fun fact, the formatter is only installed for pre-commit, and not enforced on CI, so I can just cheat! 🙈

jacobtylerwalls: Done! I was able to match the indentation by breaking the if/else branch after opening the `<img` tag like in the other file. Thanks for the reminder.

jacobtylerwalls:
Thanks for the helpful review! (I also added the logger.warning(... back from your suggestion, which I left out the first time on the idea sorl would take care of it, but the whole point is that maybe they're not taking care of everything!)

SaptakS (approved):
Looks good to me.

jacobtylerwalls:
Thanks for checking!

Re:

> The DjangoHeroForm/CorporateMemberSignUpForm accepted SVG logos

I should clarify that this was a _hypothesis_ -- we know for sure that there were .svg's in the system, but I don't exactly know when when or how they were added.

PR #2666, opened 2026-06-19
Fixed #2647 -- Reduced billboard banner height on small screens.

### LINKED ISSUE #2647: 🖼️ Banner height on mobile is too tall
Something that came up in our Fellow's check-in was that the banner app might be taller than we want it to be on mobile. 

Image

While this isn't too bad, when you have more than one paragraph, it can easily take up 60% to 75% of the screen.

nessita:
On small screens the billboard banner took up an excessive share of the viewport, leaving little room for page content.

Reworked the banner into a flex column where the container padding defines the outer edges and `gap` owns the spacing between the title, body, and call-to-action, clearing inherited margins so the rhythm stays uniform. The body now renders at a single consistent size and the mobile height is trimmed mainly by widening the text column (narrower horizontal padding) rather than by cramping the vertical spacing, which stays comfortable.

The body is wrapped in a block-level element so that, in the flex container, inline content such as links flows inline instead of each inline node being promoted to its own flex item and stacking onto separate lines.

The call-to-action button is capped and centered at every breakpoint instead of stretching full width on mid-size screens.

Screenshots:
image
image
image
image

### DIFF
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -694,20 +694,41 @@ header {
     }
 
     .banner {
+        display: flex;
+        flex-direction: column;
+        gap: 1em;
         text-align: center;
-        padding: 1em 3em;
+        // Narrower side padding than desktop widens the text column on mobile
+        // (fewer wrapped lines), while vertical padding stays comfortable.
+        padding: 1em 1.25em;
         max-width: 60em;
         margin: 0 auto;
-        p {
-            @include sans-serif;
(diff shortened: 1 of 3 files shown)


SaptakS (approved):
I am fine with these changes.

nessita:
> I am fine with these changes.

Thank you! I agree with your @SaptakS that the banner text should be reduced, I'll discuss this with Jeff in our 1-1 sync. :100:

PR #2667, opened 2026-06-19
Fix fundraising page crash when corporate logos are SVGs

### LINKED ISSUE #2664: Fundraising page returned 500 when it encountered .svg files for corporate logos

hardikkaurani:
Fixes django/djangoproject.com#2664. Avoid generating thumbnails for SVG logos as they are vector images and do not require rasterized thumbnails. Instead, return a lightweight SVG wrapper when the logo ends with .svg, preventing PIL from raising UnidentifiedImageError and crashing the page.

### DIFF
--- a/djangoproject/thumbnails.py
+++ b/djangoproject/thumbnails.py
@@ -1,6 +1,37 @@
 from sorl.thumbnail import get_thumbnail
 
 
+class SVGLogoWrapper:
+    def __init__(self, logo):
+        self.logo = logo
+
+    @property
+    def url(self):
+        return self.logo.url
+
+    @property
+    def width(self):
+        return None
+
+    @property
+    def height(self):
+        return None
+
+    @property
+    def x(self):
+        return None
+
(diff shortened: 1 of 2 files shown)


SaptakS:
@hardikkaurani thank you for the PR but https://github.com/django/djangoproject.com/pull/2665/ fixes the issue. Closing this PR, but I would encourage to take a look at other open issues and feel free to open PRs. Thanks!

PR #2668, opened 2026-06-19
Remove obsolete CSS targeting the Opera browser

adamzap:
This CSS was added in 2014 when the site was redesigned (88bcfc32), and Opera switched from its own rendering engine to Chromium in 2013. The selector and associated bug is no longer relevant.

- https://en.wikipedia.org/wiki/Opera_(web_browser)#:~:text=In%202013,%20it%20switched%20from%20the%20Presto%20engine%20to%20Chromium.
- https://web.archive.org/web/20130306150536/http://www.opera.com/docs/specs/presto2.12/css/o-vendor/

### DIFF
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -1882,19 +1882,6 @@ h2+.list-link-soup {
                     background-position: -334px -12px;
                 }
             }
-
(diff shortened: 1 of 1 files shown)


sarahboyce (approved):
Thank you!

PR #2686, opened 2026-07-01
Fixed #2685 -- Switched update_docs management command to use html builder

### LINKED ISSUE #2685: Docs build is erroring due to "Builder name djangohtml not registered or available through entry point"
A cleanup was merged for the building of Django docs in https://github.com/django/django/commit/9b1db9cb2c38c75b8fb0fc64ba5f73a6012e5b2d cc/@medmunds

We need to handle the cleanup in this repo as well:
https://github.com/django/djangoproject.com/blob/84b23d95b9f140e4b8f03530911249ea06026966/docs/management/commands/update_docs.py#L98

From logs:
```
$ grep "Builder name djangohtml not registered or available through entry point" cron_update_docs.log|head -n 5
[2026-06-10 01:15:43] sphinx-build returned an error (release en/dev, builder djangohtml): Builder name djangohtml not registered or available through entry point
[2026-06-10 01:16:21] sphinx-build returned an error (release en/6.1, builder djangohtml): Builder name djangohtml not registered or available through entry point
[2026-06-10 15:15:42] sphinx-build returned an error (release en/dev, builder djangohtml): Builder name djangohtml not registered or available through entry point
[2026-06-10 21:15:42] sphinx-build returned an error (release en/dev, builder djangohtml): Builder name djangohtml not registered or available through entry point
[2026-06-10 21:16:22] sphinx-build returned an error (release en/6.1, builder djangohtml): Builder name djangohtml not registered or available through entry point
````

### PR DESCRIPTION
**Pull request by nessita**
Following Django ticket #37150, whit commit
https://github.com/django/django/commit/9b1db9cb2c38c75b8fb0fc64ba5f73a6012e5b2d, docs building has been failing with:

sphinx-build returned an error (release en/6.1, builder djangohtml): Builder name djangohtml not registered or available through entry point

### DIFF
--- a/docs/management/commands/update_docs.py
+++ b/docs/management/commands/update_docs.py
@@ -95,8 +95,6 @@ class Command(BaseCommand):
         self.verbosity = kwargs["verbosity"]
         self.purge_cache = kwargs["purge_cache"]
 
-        self.default_builders = ["json", "djangohtml"]
-
         # Keep track of which Git sources have been updated, e.g.,
         # {'1.8': True} if the 1.8 docs updated.
         self.release_docs_changed = {}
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by tobiasmcnulty — APPROVED** [review]
Looks good and works for me! 🚢 

Image

**Comment by nessita** [comment]
@medmunds I can't request your review via GH but let me know what you think!

**Review by tobiasmcnulty — APPROVED** [review]
Good catch @medmunds and thanks for the changes @nessita ! I see the correct builders for `dev` and `6.0` now:

```
$ ls ~/.djangoproject/djangodocs/en/*/_build/

~/.djangoproject/djangodocs/en/6.0/_build/:
djangohtml	gettext		json

~/.djangoproject/djangodocs/en/dev/_build/:
html	json
```

medmunds (approved):
> @medmunds I can't request your review via GH but let me know what you think!

This all looks correct to me. (And it makes the logic around builders easier to follow. I haven't tested it locally.)

Can the `_html_builder_name()` function eventually be replaced with `"html"`, or does the site always keep the ability to build *all* old versions of the docs?

tobiasmcnulty:
> Can the `_html_builder_name()` function eventually be replaced with `"html"`, or does the site always keep the ability to build all old versions of the docs?

I looked into this briefly, and I believe it would be safe to simplify `_html_builder_name()` once 5.2 is no longer receiving security updates, provided we keep a copy of the built HTML for the older versions (which we will, barring disaster).

If we make such a change, the command could fail gracefully if the `html` builder can't be found. (I believe part of the issue now is that the database isn't updated until after the HTML is built, otherwise we may not have noticed this until someone checked the `.zip` download.)

PR #2688, opened 2026-07-02
Refs #2685 -- Isolated per-release docs via a separate process to avoid Sphinx module caching.

nessita:
`update_docs` builds every release (across all supported Django versions and languages) in a single long-running process. Each checkout's `docs/_ext/djangodocs.py` is loaded by dynamically appending that checkout's own `docs/_ext` directory to `sys.path` and importing it by bare module name ("djangodocs"). Once Python caches that name in `sys.modules`, a later checkout's import of the same name silently reused whichever checkout's module got imported first in the process, regardless of which checkout was actually being built.

In practice this meant a newer branch's djangodocs.py (which no longer registers the "djangohtml" builder) got reused for older branches that still require it, so those older branches failed outright with "Builder name djangohtml not registered", and worse, that failure aborted publishing for that release entirely, including versions whose json build had already succeeded.

Split the actual Sphinx build out into a new `build_doc_release` management command, invoked by `update_docs` as a subprocess, one process per release. `update_docs` now only handles orchestration: which releases need building, syncing their git checkouts (shared across a version's languages) and translations, and dispatching the build. `build_doc_release` does nothing but build an already-synced checkout, in its own fresh interpreter, so no Python state can leak between checkouts regardless of build order.

### DIFF
--- a/docs/management/commands/build_doc_release.py
+++ b/docs/management/commands/build_doc_release.py
@@ -35,6 +35,7 @@ from sphinx.testing.util import _clean_up_global_state
 from sphinx.util.docutils import docutils_namespace, patch_docutils
 
 from ...models import DocumentRelease
+from ...utils import capture_sentry_exception
 
 
 class Command(BaseCommand):
@@ -136,11 +137,11 @@ class Command(BaseCommand):
                 # Clean up global state after building each language.
                 _clean_up_global_state()
             except SphinxError as e:
-                self.stderr.write(
+                capture_sentry_exception(e, flush=True)
+                raise CommandError(
                     "sphinx-build returned an error (release %s, builder %s): %s"
                     % (release, builder, str(e))
-                )
-                return
+                ) from e
 
         #
         # Create a zip file of the HTML build for offline reading.
--- a/docs/management/commands/build_doc_release.py
+++ b/docs/management/commands/build_doc_release.py
@@ -1,0 +1,216 @@
+"""Build the documentation for a single release (version + language).
+
+This command is meant to be invoked as a subprocess by `update_docs`, one
+process per release. `update_docs` builds many releases across Django versions
+in a single run, and in some checkouts, `docs/_ext/djangodocs.py` dynamically
+sys.path-append their own `_ext` directory and import extension modules (e.g.
+"djangodocs", "github_links") by bare name. Building all releases in one Python
+process let a newer checkout's already-imported module get silently reused for
+an older checkout's build later in the same run. A fresh process per release
+avoids that class of bug entirely.
+
+`update_docs` is responsible for syncing the release's git checkout (shared
+across a version's languages) before invoking this command; this command
+assumes the checkout is already up to date and only builds it.
+
+"""
+
+import json
+import multiprocessing
+import os
+import shutil
+import subprocess
+import sys
+import zipfile
+from contextlib import closing
+from pathlib import Path
+
+from django.conf import settings
+from django.core.management import BaseCommand, CommandError
+from django.utils.translation import to_locale
+from sphinx.application import Sphinx
+from sphinx.config import Config
+from sphinx.errors import SphinxError
+from sphinx.testing.util import _clean_up_global_state
+from sphinx.util.docutils import docutils_namespace, patch_docutils
+
+from ...models import DocumentRelease
+
+
+class Command(BaseCommand):
+    help = (
+        "Build the docs for a single release (version + language). Intended "
+        "to be invoked as a subprocess by update_docs, one process per "
+        "release, once its git checkout is already up to date."
+    )
+
+    def add_arguments(self, parser):
+        parser.add_argument("version", help="Which version to build (e.g. 6.0, dev)")
+        parser.add_argument(
+            "--language", required=True, help="Language code of the release to build"
+        )
+
+    def handle(self, version, **options):
+        self.verbosity = options["verbosity"]
+        try:
+            release = DocumentRelease.objects.get_by_version_and_lang(
+                version, options["language"]
+            )
+        except DocumentRelease.DoesNotExist:
+            raise CommandError(
+                "No DocumentRelease found for version=%r lang=%r"
+                % (version, options["language"])
+            )
+
+        self.build_doc_release(release)
+
+    def _html_builder_name(self, source_dir):
+        """Return the Sphinx builder name to use for HTML output for a given
+        checkout.
+
+        Older Django versions' docs/_ext/djangodocs.py registers a custom
+        "djangohtml" builder. Django's #37150 cleanup removed it in favor of
+        hooking into any HTML-format builder, including the standard "html"
+        builder, but that cleanup isn't backported to branches that predate
+        it (e.g. 6.0, 5.2), so "djangohtml" must still be requested there.
+        """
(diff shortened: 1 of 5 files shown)


nessita:
cc/ @medmunds for awareness and perhaps you have ideas about how to make this whole django/website docs relationship more healthy :-)

tobiasmcnulty (approved):
This looks good and works for me, with and without versions passed to `update_docs`! 🚢 

We might want to `raise CommandError()` from the subprocess so the status code can be checked from the parent. We can also optionally send the original exception to Sentry.

```patch
diff --git a/docs/management/commands/build_doc_release.py b/docs/management/commands/build_doc_release.py
index e89f589d..67b345c3 100644
--- a/docs/management/commands/build_doc_release.py
+++ b/docs/management/commands/build_doc_release.py
@@ -35,6 +35,7 @@ from sphinx.testing.util import _clean_up_global_state
 from sphinx.util.docutils import docutils_namespace, patch_docutils
 
 from ...models import DocumentRelease
+from ...utils import capture_sentry_exception
 
 
 class Command(BaseCommand):
@@ -136,11 +137,12 @@ class Command(BaseCommand):
                 # Clean up global state after building each language.
                 _clean_up_global_state()
             except SphinxError as e:
-                self.stderr.write(
+                error_msg = (
                     "sphinx-build returned an error (release %s, builder %s): %s"
                     % (release, builder, str(e))
                 )
-                return
+                capture_sentry_exception(e, flush=True)
+                raise CommandError(error_msg) from e
 
         #
         # Create a zip file of the HTML build for offline reading.
diff --git a/docs/management/commands/update_docs.py b/docs/management/commands/update_docs.py
index 671002ad..0021a96f 100644
--- a/docs/management/commands/update_docs.py
+++ b/docs/management/commands/update_docs.py
@@ -188,11 +188,12 @@ class Command(BaseCommand):
             str(self.verbosity),
         ]
 
-        result = subprocess.run(command)
+        result = subprocess.run(command, capture_output=True, text=True)
         if result.returncode != 0:
+            error_detail = result.stderr.strip() or "(no stderr output)"
             self.stderr.write(
-                "build_doc_release subprocess failed for %s (exit code %s)"
-                % (release, result.returncode)
+                "build_doc_release subprocess failed for %s (exit code %s): %s"
+                % (release, result.returncode, error_detail)
             )
 
     def update_git(self, url, destdir, changed_dir="."):
diff --git a/docs/tests/test_utils.py b/docs/tests/test_utils.py
index f676540b..86ca9430 100644
--- a/docs/tests/test_utils.py
+++ b/docs/tests/test_utils.py
@@ -1,9 +1,39 @@
 import os
 from pathlib import Path
+from unittest.mock import MagicMock, patch
 
 from django.test import SimpleTestCase
 
-from ..utils import extract_inner_html, get_doc_path, sanitize_for_trigram
+from ..utils import (
+    capture_sentry_exception,
+    extract_inner_html,
+    get_doc_path,
+    sanitize_for_trigram,
+)
+
+
+class TestCaptureSentryException(SimpleTestCase):
+    def test_calls_capture_exception_when_sentry_available(self):
+        mock_sentry = MagicMock()
+        with patch("docs.utils.sentry_sdk", mock_sentry):
+            error = ValueError("something broke")
+            result = capture_sentry_exception(error)
+        mock_sentry.capture_exception.assert_called_once_with(error)
+        self.assertTrue(result)
+
+    def test_no_op_when_sentry_not_available(self):
+        with patch("docs.utils.sentry_sdk", None):
+            error = ValueError("something broke")
+            result = capture_sentry_exception(error)
+        self.assertFalse(result)
+
+    def test_flush_blocks_until_sent(self):
+        mock_sentry = MagicMock()
+        with patch("docs.utils.sentry_sdk", mock_sentry):
+            error = ValueError("something broke")
+            capture_sentry_exception(error, flush=True)
+        mock_sentry.capture_exception.assert_called_once_with(error)
+        mock_sentry.flush.assert_called_once()
 
 
 class TestUtils(SimpleTestCase):
diff --git a/docs/utils.py b/docs/utils.py
index 973f684b..3bf67ecd 100644
--- a/docs/utils.py
+++ b/docs/utils.py
@@ -6,6 +6,25 @@ from pathlib import Path
 from django.http import Http404
 
+try:
+    import sentry_sdk
+except ImportError:
+    sentry_sdk = None
+
+
+def capture_sentry_exception(error, flush=False):
+    """
+    Report an exception to Sentry if sentry_sdk is available.
+    Pass flush=True to block until the event is sent (use before process exit).
+    Returns True if sent to Sentry, False if sentry_sdk is not installed.
+    """
+    if sentry_sdk is not None:
+        sentry_sdk.capture_exception(error)
+        if flush:
+            sentry_sdk.flush()
+        return True
+    return False
+
 
 def get_doc_root(lang, version, builder="json"):
     return settings.DOCS_BUILD_ROOT / lang / version / "_built" / builder

```

medmunds:
> cc/ @medmunds for awareness and perhaps you have ideas about how to make this whole django/website docs relationship more healthy :-)

👍 There's a chance this resolves #2416. (Or we might need to isolate per language as well as per release for that, but this gives us the mechanism.)

I'm not sure how to make the relationship more healthy yet, but I'm starting to understand some of the conflicts. (I should maybe take this to a forum discussion, but am dropping it here for now.)
1. djangoproject.com docs builds depend on *code* (not just docs source) that's bundled with Django itself, and can vary release-to-release.
   * django:docs/_ext/djangodocs and github_links Sphinx extensions; also to some extent django:docs/conf.py which mixes config and code
   * An alternative solution to #2685 would have been to vendor the latest django:docs/_ext/djangodocs.py into djangoproject.com and use it for all builds. (That would have problems of its own, and I'm not arguing in favor of it.)
3. We have two different Sphinx themes for Django docs: one in django:docs/_theme/djangodocs and one in djangoproject\.com.
   * The djangodocs theme is used for local and Django PR preview builds, as well as for the downloadable version of the docs from djangoproject\.com. (The downloadable docs style doesn't match the online style, which is maybe an issue to report.)
   * The djangodocs theme is less-actively maintained and has some bugs and accessibility issues. (See [ticket-37138](https://code.djangoproject.com/ticket/37138) and forum discussions linked from there.)
   * The djangoproject\.com docs theme is intertwined with other djangoproject\.com styles. https://github.com/django/django/pull/21416 is an attempt to extract that for bundling with Django, but it's a bit ugly. It also raises maintenance concerns that are effectively the opposite of item 1.
   * (Related maintenance concern: code\.djangoproject\.com has duplicate, vendored copies of some djangoproject\.com SCSS and JS.)

Edit: I guess one solution might be to move the djangoproject Sphinx theme and the djangodocs Sphinx extension into separate packages, and have both djangoproject\.com and Django depend on them.

nessita:
> Edit: I guess one solution might be to move the djangoproject Sphinx theme and the djangodocs Sphinx extension into separate packages, and have both djangoproject.com and Django depend on them.

Another solution which we discussed already but leaving here for completeness, is to fully drop the  djangoproject Sphinx theme in the Django source and use a well-known theme. This way we could try to clean up conf.py and djangodocs.py to separate concerns. In an ideal split:
* django source curates and is responsible for the docs content and semantic structure (notes, warning, versionadded, versionchanged, etc etc)
* django source builds docs using a standard high quality theme (but not custom)
* the website curates and is responsible for the styling of the site which includes the docs

PR #2690, opened 2026-07-03
Refs #2685 -- Fixed build_doc_release subprocess call failing under "-m django".

nessita:
When `update_docs` run was launched via `python -m django update_docs` (as cron does in production), sys.argv[0] resolved to the path of Django's own __main__.py inside site-packages. Running that path directly as a script, rather than via -m, meant Python prepended that script's own directory to sys.path instead of the cwd, so the child process couldn't import the project's `djangoproject` settings package, failing with "ModuleNotFoundError: No module named 'djangoproject'".

Local testing never caught this because `python manage.py update_docs` is used, where sys.argv[0] is a project-local script that works fine when re-invoked directly.

The fix is to invoke the child via `-m django` explicitly instead of reusing sys.argv[0], matching how update_docs itself is actually invoked in production and restoring the sys.path behavior the child needs.

### DIFF
--- a/docs/management/commands/update_docs.py
+++ b/docs/management/commands/update_docs.py
@@ -186,7 +186,8 @@ class Command(BaseCommand):
         """
         command = [
             sys.executable,
-            sys.argv[0],
+            "-m",
+            "django",
             "build_doc_release",
             release.version,
             "--language",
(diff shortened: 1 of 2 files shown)


tobiasmcnulty (approved):
This looks right to me! Tested with both of the following:

1. `python manage.py update_docs`
2. `python -m django update_docs`

PR #2713, opened 2026-07-13
Added GitHub action to test against Django main.

### LINKED ISSUE #2391: [Proposal] Scheduled daily github workflow to run tests using Django's main branch

sarahboyce:
Fixes https://github.com/django/djangoproject.com/issues/2391

### DIFF
--- a/.github/workflows/django-main.yml
+++ b/.github/workflows/django-main.yml
@@ -1,0 +1,71 @@
+---
+name: Tests on Django main
+
+on:
+    schedule:
+        - cron: "0 0 * * *"
+    workflow_dispatch:
+
+env:
+    PYTHON_VERSION: 3.12
+
+jobs:
+    tests:
+        runs-on: ubuntu-24.04
+
+        services:
+            postgres:
+                image: postgres:17-trixie
+                env:
+                    POSTGRES_PASSWORD: postgres
+                options: >-
+                    --health-cmd pg_isready
+                    --health-interval 10s
+                    --health-timeout 5s
+                    --health-retries 5
+                ports:
+                    - 5432:5432
+
+        steps:
+            - uses: actions/checkout@v7.0.0
+            - name: Set up Python
+              uses: actions/setup-python@v6
+              with:
+                  python-version: ${{ env.PYTHON_VERSION }}
+            - name: Install gettext
+              run: sudo apt-get update && sudo apt-get install -y gettext
+            - name: Install dependencies
+              run: |
+                  python -m pip install --upgrade pip setuptools
+                  python -m pip install -r requirements/tests.txt
+                  python -m pip install --upgrade --force-reinstall \
+                            "Django @ git+https://github.com/django/django.git@main"
+            - name: Install playwright browsers
+              run: |
+                  python -m playwright install --with-deps chromium
+            - name: Set up databases
+              run: |
+                  PGPASSWORD="postgres" createuser -U postgres -d djangoproject --superuser -h localhost
+                  PGPASSWORD="postgres" createdb -U postgres -O djangoproject djangoproject -h localhost
+                  PGPASSWORD="postgres" createuser -U postgres -d code.djangoproject --superuser -h localhost
+                  PGPASSWORD="postgres" createdb -U postgres -O code.djangoproject code.djangoproject -h localhost
+                  PGPASSWORD="postgres" psql -U postgres -h localhost -c "ALTER USER djangoproject WITH PASSWORD 'secret';"
+                  PGPASSWORD="postgres" psql -U postgres -h localhost -c "ALTER USER \"code.djangoproject\" WITH PASSWORD 'secret';"
+            - name: Load Trac data
+              run: |
+                  PGPASSWORD="postgres" psql -U postgres -d code.djangoproject -h localhost < tracdb/trac.sql
+            - name: Create secrets.json
+              working-directory: ..
+              run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/django-main.yml
```diff
@@ -0,0 +1,72 @@
+---
+name: Tests on Django main
+
+on:
+    schedule:
+        - cron: "0 0 * * *"
+    workflow_dispatch:
+    pull_request:
```
sarahboyce:
```suggestion
```
I will remove this once the action is approved. Added for the review process only

tracdb/migrations/0003_attachment__time_attachment_author_and_more.py
```diff
@@ -0,0 +1,425 @@
+# Generated by Django 6.2 on 2026-07-13 09:35
```
sarahboyce: This is required after the change of https://github.com/django/django/commit/7b8504eaabea6c70195baa3c02155c24a6e270c6
I see no issue in adding this migration now. It is essentially a no-op.

foundation/admin.py
```diff
@@ -19,7 +19,7 @@ class TermAdmin(admin.ModelAdmin):
 class BoardMemberAdmin(admin.ModelAdmin):
     list_display = ("full_name", "office", "term")
     list_filter = ("office", "term")
-    list_select_related = True
+    list_select_related = ("account", "office", "term")
```
adamzap, line 22: Is this change intentional?

sarahboyce: I can pull out to a different PR but it was intentional driven by a deprecation warning

adamzap: Ah, ok! I see it [here](https://docs.djangoproject.com/en/6.1/releases/6.1/#minor-features). I think a separate PR would be great since this should be fixed eventually anyway. Thanks!

adamzap:
I rebased this since #2716 was merged.

adamzap (commented):
The action looks good to me!

tracdb/migrations/0003_attachment__time_attachment_author_and_more.py
adamzap: I seems a bit unfortunate for something from Django `main` to leak back into the codebase, and it's kind of confusing that the comment says "Generated by Django 6.2...". Is there no way around committing the migration?

I wonder what others think!

CuriousLearner (approved):
Looks good to me. 🌮 The only thing is removing `pull_request` trigger before merging.

sarahboyce:
I want to wait till https://code.djangoproject.com/ticket/37224 is merged as then I can regenerate the migration without the temporary defaults :+1:

tracdb/migrations/0003_attachment__time_attachment_author_and_more.py
adamzap: Would it be too weird to have this workflow run `makemigrations` before the test suite runs?

sarahboyce: I have pulled out the migration to https://github.com/django/djangoproject.com/pull/2755, once merged I can remove this from here
It is a really unusual case and normally a Django upgrade wouldn't cause a migration but our migration state is out of sync with the current models.

sarahboyce:
> I want to wait till https://code.djangoproject.com/ticket/37224 is merged as then I can regenerate the migration without the temporary defaults 👍

This ticket is fixed, I pulled the newly generated migration to this PR https://github.com/django/djangoproject.com/pull/2755
If that gets merged, I can remove the migration from here and rebase

SaptakS:
Merged the PR. You can rebase this PR.

tracdb/migrations/0003_attachment__time_attachment_author_and_more.py
jacobtylerwalls: > Would it be too weird to have this workflow run makemigrations before the test suite runs?

There's already a check in the test suite for pending migrations, see `PendingMigrationsTests`.

.github/workflows/django-main.yml
sarahboyce: The latest run passed so dropping: https://github.com/django/djangoproject.com/actions/runs/31110200783/job/92645676563?pr=2713

tracdb/migrations/0003_attachment__time_attachment_author_and_more.py
adamzap: Ah, right. My idea wasn't to catch them and fail the test suite but to generate and apply any new migrations from `django@main` without needing to be commit them. I guess we'd have to skip that test if we went that route.

jacobtylerwalls: Gotcha. I think we'd rather have a canary test failure, since it's at least as likely there's a regression on Django main than another legitimate migration change if this were to fail again.

adamzap: Ok cool. That seems reasonable to me!

adamzap (approved):
👍 Looks great!

PR #2715, opened 2026-07-13
Fixed #2714 -- Made the Django Core Developers page discoverable.

### LINKED ISSUE #2714: The Django Core Developers page is not discoverable
The page listing Django Core Developer title recipients at https://www.djangoproject.com/foundation/django_core/ is an orphan, and has been since it was added in 2021 (#1099, which superseded #1043): its route has no `name=`, it's absent from `sitemap.xml`, and as far as I can tell nothing on the site links to it (no template references it; I also checked the foundation pages and the governance blog posts). The only way to reach it is by knowing the URL.

Since the site's search is sitemap-driven, the page is also invisible to the site's own search — as #2284 put it, "if anything is missing from the search, it is missing from a sitemap. Going forward we should make sure pages we want to be searched are included in our sitemaps."

For context, the page exists because [DEP 10](https://github.com/django/deps/blob/main/final/0010-new-governance.rst#repurposed-role-django-core-developer) has the DSF "maintain and publish" the list of title recipients — making the page reachable seems in line with that intent. I couldn't find any recorded decision to keep it unlinked (nothing in #1043/#1099, no `noindex`), but if that was deliberate, happy to close this.

Two parts:

- Mechanical: name the route and add the page to the sitemap, next to the other people-listing pages (`members:individual-members`, `members:teams`), which are already there. I'll open a PR for this.
- Navigation: decide where to link the page from. The natural spot seems to be the /foundation/ page itself (next to the membership and prizes mentions; that's a flatpage, so an admin edit rather than a template change), or a "see also" link from the individual members page. Precedent: the teams page used to have no navigation route either (#2226) and was fixed by pointing a footer link at it (#2225).

For transparency: I'm one of the people listed on that page, which is how I noticed it isn't linked from anywhere. This issue and the accompanying PR were prepared with help from Claude Code; I verified every claim above by hand.

sir-sigurd:
Fixes #2714.

The `/foundation/django_core/` page has no route name, no `sitemap.xml` entry, and no inbound links that I could find — see the issue for the full background.

Changes:

- Add `name="foundation_core_developers"` to the route
- Add the page to `TemplateViewSitemap`, next to the other people-listing pages (`members:individual-members`, `members:teams`), which are already there

The sitemap's `URLObject` inventory is name-based, which is probably how the unnamed `django_core` route got skipped when the sitemap was introduced in #2284.

Verified locally: the full test suite passes (`make ci`), and pre-commit is clean on the changed files; after the change, `sitemap.xml` contains the page URL and the page renders with a single `<h1>`. Naming the route also brings the page under the existing `test_single_h1_per_page` coverage, which iterates over the named top-level URLs.

This PR deliberately does **not** add a navigation link — where to put one is a call I'd rather leave to the maintainers (options in the issue).

In line with the [AI-assisted contributions policy](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/submitting-patches/#ai-assisted-contributions), I used Claude Code while preparing this patch; I reviewed and tested the change myself and take full responsibility for it.

### DIFF
--- a/djangoproject/tests.py
+++ b/djangoproject/tests.py
@@ -233,6 +233,10 @@ class SiteMapTests(TestCase):
         response = self.client.get(reverse("sitemap"))
         self.assertEqual(response.status_code, 200)
 
+    def test_sitemap_contains_core_developers_page(self):
+        response = self.client.get(reverse("sitemap"))
+        self.assertContains(response, "/foundation/django_core/")
+
 
 class EndToEndTests(ReleaseMixin, StaticLiveServerTestCase):
     @classmethod
--- a/djangoproject/tests.py
+++ b/djangoproject/tests.py
@@ -233,10 +233,6 @@ class SiteMapTests(TestCase):
         response = self.client.get(reverse("sitemap"))
         self.assertEqual(response.status_code, 200)
 
-    def test_sitemap_contains_core_developers_page(self):
-        response = self.client.get(reverse("sitemap"))
-        self.assertContains(response, "/foundation/django_core/")
-
 
 class EndToEndTests(ReleaseMixin, StaticLiveServerTestCase):
     @classmethod
--- a/djangoproject/sitemaps.py
+++ b/djangoproject/sitemaps.py
@@ -60,6 +60,7 @@ class TemplateViewSitemap(LocationAbsoluteUrlMixin, sitemaps.Sitemap):
             URLObject("diversity_changes"),
             # foundation
             URLObject("foundation_meeting_archive_index"),
+            URLObject("foundation_core_developers"),
             # fundraising
             URLObject("fundraising:index"),
             # members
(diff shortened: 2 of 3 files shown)


djangoproject/tests.py
```diff
@@ -233,6 +233,10 @@ def test_sitemap_renders(self):
         response = self.client.get(reverse("sitemap"))
         self.assertEqual(response.status_code, 200)
 
+    def test_sitemap_contains_core_developers_page(self):
+        response = self.client.get(reverse("sitemap"))
+        self.assertContains(response, "/foundation/django_core/")
+
```
sarahboyce: Personally I wouldn't add a new test asserting only this URL as we do not assert for all the others

sarahboyce (approved):
Thank you!

djangoproject/tests.py
adamzap: I agree. I think we can drop the test and be good to merge.

sir-sigurd: Dropped, thanks!

PR #2717, opened 2026-07-17
[checklists] A couple of fixes.

nessita:
There are some CVEs that only affect supported versions, so we should not include the line "_Earlier, unsupported Django series (such as 4.1.x) were not evaluated and may also be affected._" in the CVE description when we are sure it does not.


checklists/models.py
```diff
@@ -69,6 +69,15 @@ def get_cve_default():
     return f"CVE-{datetime.date.today().year}-{get_random_string(5)}"
 
 
+def get_unsupported_series_default():
+    """Comma separated list of the 3 most recently EOL'd feature versions.
+
+    Return an empty string when no unsupported releases are known.
+    """
+    series = [f"{r.feature_version}.x" for r in Release.objects.unsupported()[:3]]
```
jacobtylerwalls, line 77: We should be okay here, since it looks like we tend to not immediately EOL the last patch release on an EOL version, even if the notes say we do.

e.g on April 7, 2026 we said with this patch release 4.2 was EOL, but in the admin we bumped the EOL date out to the end of the month.

jacobtylerwalls (approved):
Looks good!

checklists/models.py
nessita: We can always override manually the list per CVE so I think we are OK.

PR #2728, opened 2026-07-21
Added a reCAPTCHA to the account registration form.

nessita:
Self-serve registration was the only public form on the site that sent mail without any protection, so automated signups turned directly into activation emails to nonexistent addresses. The resulting hard multiple re-occurring bounces lowering email sending reputation scores.

`RegistrationForm` now carries the same ReCaptchaV3 field already used by the contact and fundraising forms.

PR #2729, opened 2026-07-21
Added `delete_unactivated_users` management command.

tobiasmcnulty:
Follow-up work to #2728 to delete spam users.

### DIFF
--- a/accounts/management/commands/delete_unactivated_users.py
+++ b/accounts/management/commands/delete_unactivated_users.py
@@ -1,0 +1,86 @@
+"""Delete unactivated user accounts.
+
+An alternative to django-registration-redux's ``cleanupregistration`` with:
+
+* **Bulk deletion** - single ``QuerySet.delete()`` instead of iterating one by
+  one. Django still batches the SQL statements via ``GET_ITERATOR_CHUNK_SIZE``,
+  but the number of queries does not grow per user.
+* **Hard safety** - users with protected related objects (BoardMember,
+  Releaser) are skipped, so a single protected user cannot block the cleanup.
+* **``--dry-run``** for a safe preview.
+"""
+
+from datetime import timedelta
+
+from django.conf import settings
+from django.contrib.auth.models import User
+from django.core.management.base import BaseCommand
+from django.db import transaction
+from django.utils import timezone
+
+
+class Command(BaseCommand):
+    help = (
+        "Delete users and profiles that never activated their account "
+        "(older than ACCOUNT_ACTIVATION_DAYS and never clicked the activation link). "
+        "Only User, Profile, and RegistrationProfile rows are removed. Users with "
+        "protected related objects (e.g. BoardMember or Releaser) are skipped."
+    )
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--dry-run",
+            action="store_true",
+            help="Show what would be deleted without actually deleting.",
+        )
+
+    def handle(self, *args, **options):
+        verbosity = options["verbosity"]
+        dry_run = options["dry_run"]
+
+        expiry_cutoff = timezone.now() - timedelta(
+            days=settings.ACCOUNT_ACTIVATION_DAYS
+        )
+
+        qs = User.objects.filter(
+            is_active=False,
+            date_joined__lt=expiry_cutoff,
+            registrationprofile__activated=False,
+            last_login__isnull=True,
+            # Skip users holding data that must outlive them. Deleting in bulk
+            # is all or nothing, so a single protected user would otherwise
+            # abort the whole run.
+            boardmember__isnull=True,
+            releaser__isnull=True,
+        )
+        total = qs.count()
+
+        if verbosity >= 1:
+            self.stdout.write(f"Found {total} unactivated user(s) to process.")
+
+        if total == 0:
+            if verbosity >= 1:
+                self.stdout.write("Nothing to do.")
+            return
+
+        if dry_run:
+            if verbosity >= 1:
+                self.stdout.write(
+                    self.style.NOTICE(
+                        f"DRY RUN - would delete {total} user(s) "
+                        f"(cascading to Profile and RegistrationProfile)."
+                    )
+                )
+            return
+
+        # CASCADE chain: User -> Profile, User -> RegistrationProfile.
+        with transaction.atomic():
+            _deleted_count, deleted_by_model = qs.delete()
+
+        if verbosity >= 1:
+            self.stdout.write(
+                self.style.SUCCESS(
+                    f"Deleted {deleted_by_model.get('auth.User', 0)} user(s) "
+                    f"(plus associated profiles)."
+                )
+            )
--- a/accounts/test_delete_unactivated_users.py
+++ b/accounts/test_delete_unactivated_users.py
@@ -1,0 +1,206 @@
+from datetime import timedelta
+from io import StringIO
+
+from django.contrib.auth.models import User
+from django.core.management import call_command
+from django.db import connection
+from django.test import TestCase, override_settings
+from django.test.utils import CaptureQueriesContext
+from django.utils import timezone
+from registration.models import RegistrationProfile
+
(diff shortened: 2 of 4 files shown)


accounts/management/commands/delete_unactivated_users.py
```diff
@@ -0,0 +1,92 @@
...
+    def handle(self, *args, **options):
+        verbosity = options["verbosity"]
+        dry_run = options["dry_run"]
+
+        expiry_cutoff = timezone.now() - timezone.timedelta(
+            days=settings.ACCOUNT_ACTIVATION_DAYS
+        )
+
+        qs = User.objects.filter(
+            is_active=False,
```
nessita, line 46: Did you consider the thing that Markus said about is_active potentially being False?

On a second look, the field is not `null=True` so I think `None` is not an option in this field.

accounts/management/commands/delete_unactivated_users.py
```diff
@@ -0,0 +1,92 @@
...
+                    f"Deletion aborted: {len(models_involved)} related "
+                    f"object(s) would be affected: {', '.join(sorted(models_involved))}"
+                )
+            )
+            raise
+
+        if verbosity >= 1:
+            self.stdout.write(
+                self.style.SUCCESS(
+                    f"Deleted {deleted_count} user(s) " f"(plus associated profiles)."
```
nessita: A nitpick, `deleted_count` is not the amount of users, but the amount of all deleted objects:
```pycon
>>> User.objects.filter(username="qqqq").delete()
(2, {'registration.RegistrationProfile': 1, 'auth.User': 1})
```

accounts/management/commands/delete_unactivated_users.py
```diff
@@ -0,0 +1,92 @@
...
+                deleted_count, _ = qs.delete()
+        except ProtectedError as e:
+            models_involved = {model._meta.label for model in e.protected_objects}
+            self.stderr.write(
+                self.style.ERROR(
+                    f"Deletion aborted: {len(models_involved)} related "
+                    f"object(s) would be affected: {', '.join(sorted(models_involved))}"
+                )
+            )
+            raise
```
nessita: This `raise` has the whole command invocation failing even if a single user has protected objects. This means the bulk delete() is all-or-nothing, and per the docstring this may not be the intention?

accounts/test_delete_unactivated_users.py
```diff
@@ -0,0 +1,258 @@
...
+    #  16. DELETE registration_registrationprofile - CASCADE
+    #  17. DELETE auth_user - main deletion
+    #  18. RELEASE SAVEPOINT - commit transaction
+    # NOTE: Update this count if new ForeignKey to User is added to the schema.
+
+    def test_deletes_expired_unactivated_user(self):
+        """Single expired, unactivated user is deleted."""
+        user = self._create_unactivated_user(days_ago=4)
+
+        with self.assertNumQueries(18):
```
nessita: This feels a bit like a maintenance trap. I would rather we test deleting 1 user, record the query count, then delete 3 and assert the query count is the same.

nessita: Something like:
```python
from django.db import connection
from django.test.utils import CaptureQueriesContext


    def test_query_count_does_not_scale_with_number_of_users(self):
        self._create_unactivated_user(days_ago=4)
        with CaptureQueriesContext(connection) as one_user:
            call_command("delete_unactivated_users", verbosity=0)
        self.assertEqual(User.objects.count(), 0)

        for days_ago in (5, 6, 7):
            self._create_unactivated_user(days_ago=days_ago)
        with CaptureQueriesContext(connection) as three_users:
            call_command("delete_unactivated_users", verbosity=0)
        self.assertEqual(User.objects.count(), 0)

        self.assertEqual(len(three_users), len(one_user))
```

accounts/test_delete_unactivated_users.py
```diff
@@ -0,0 +1,258 @@
...
+
+        self.assertTrue(User.objects.filter(pk=user.pk).exists())
+
+    # -- Failure cases (database-level constraints) --
+
+    def test_fails_on_board_member_protect(self):
+        """BoardMember (PROTECT) fails at the database level."""
+        user = self._create_unactivated_user(days_ago=4)
+        office = Office.objects.create(name="Chair")
+        term = Term.objects.create(year="2024")
```
nessita: nitpick: I think term should be an int

```suggestion
        term = Term.objects.create(year=2024)
```

accounts/management/commands/delete_unactivated_users.py
nessita: Also, if we do want to abort execution, I think we should raise CommandError for better output for when running the command.

accounts/management/commands/delete_unactivated_users.py
```diff
@@ -0,0 +1,92 @@
...
+
+        # Single deletion query wrapped in a transaction.
+        # CASCADE chain: User → Profile, User → RegistrationProfile
+        # Database-level constraints will abort if:
+        #   - BoardMember (PROTECT) exists for any user
+        #   - Releaser (PROTECT) exists for any user
+        try:
+            with transaction.atomic():
+                deleted_count, _ = qs.delete()
+        except ProtectedError as e:
```
jacobtylerwalls: I don't know that I'd bother catching and massaging the error message: the error message is already pretty supple.

nessita (commented):
Thanks, added some comments!

accounts/management/commands/delete_unactivated_users.py
tobiasmcnulty: Good catch, I think that was `last_login__isnull=True`. I will add a test + `filter()` kwarg.

accounts/management/commands/delete_unactivated_users.py
tobiasmcnulty: @jacobtylerwalls in his review also suggested removing the try/except altogether. I'm good with doing that or raising a CommandError here instead. Thoughts either way?

accounts/test_delete_unactivated_users.py
tobiasmcnulty: I don't mind changing it, but my intent here was to have the test break if a new foreign key is added, so that we have to think intentionally about what to do with that relationship during user cleanup.

Are there are situations that make this brittle in less helpful ways? Let me know what you think. Also happy to remove the assertion altogether if that's simpler.

accounts/management/commands/delete_unactivated_users.py
tobiasmcnulty: Good catch, thanks! Will fix.

accounts/management/commands/delete_unactivated_users.py
```diff
@@ -0,0 +1,83 @@
...
+        # Database-level constraints will abort if:
+        #   - BoardMember (PROTECT) exists for any user
+        #   - Releaser (PROTECT) exists for any user
+        with transaction.atomic():
+            _deleted_count, deleted_by_model = qs.delete()
+
+        if verbosity >= 1:
+            self.stdout.write(
+                self.style.SUCCESS(
+                    f"Deleted {deleted_by_model['auth.User']} user(s) "
```
jacobtylerwalls: Seems OK, we're never going to swap the user model.

jacobtylerwalls (approved):
👍

accounts/test_delete_unactivated_users.py
tobiasmcnulty: Updated in 16a0d0193618cf4af1a24405765cc590e932fa08

accounts/management/commands/delete_unactivated_users.py
tobiasmcnulty: I removed the assertion in f65aa7ec93f6aca4701906a96bd1b08503ae5890 and propose we leave it like that for now.

nessita (commented):
Thank you @tobiasmcnulty! I'll push minimal cleanups, squash commits, and merge.

Issue #2373, opened 2025-12-06
Set Django 6 as the new django-docs-translations sources


claudep:
It looks like the https://docs.djangoproject.com/pots/*.pot files (source for Transifex translations) is still on Django 5.2.
Could you please update the `DocumentRelease.is_default` flag so as the source will be Django 6.0?

@nessita, is it something you can do?

nessita:
> It looks like the https://docs.djangoproject.com/pots/*.pot files (source for Transifex translations) is still on Django 5.2. Could you please update the `DocumentRelease.is_default` flag so as the source will be Django 6.0?
> 
> [@nessita](https://github.com/nessita), is it something you can do?

Yes, I have the permissions, but also I already did this last week, as part of the 6.0 final release:

<img width="1446" height="922" alt="Image" src="https://github.com/user-attachments/assets/00ad0995-3cab-46ed-8c76-917bb35300da" />

claudep:
Yes, looks good. However, I wonder why we still have "Django 5.2" in the file http://docs.djangoproject.com/pots/index.pot (and oll other pot files).

sarahboyce:
Does the 5.2 branch need to be set as stale or something like this? https://github.com/django/django-docs-translations/branches

claudep:
No, I don't think so. The `is_default` flag should be enough. See https://github.com/django/djangoproject.com/blob/39467c3ae39d5708ee9486840643a3e1448f2f35/docs/management/commands/update_docs.py#L182

claudep:
Please, could someone having access to the server debug what's happening and why the pot build is not using the 6.0 sources?

kehach07:
@claudep ,
I can’t debug server-side, but a few quick checks might help:
1) Log which `DocumentRelease` (version + is_default) is selected during POT generation.
2) Verify the POT build path doesn’t bypass `is_default`.
3) Try a clean POT rebuild to rule out stale 5.2 artifacts.

jacobtylerwalls:
Hey @claudep, thanks for raising this on the forum. You linked to an `update_docs` management command. Is there an automated process that runs that? I don't see it mentioned anywhere in [How to release Django](https://docs.djangoproject.com/en/6.0/internals/howto-release-django/).

bmispelon:
I've just looked into this briefly and it appears the cause of this is that the `stable` symlink that's used by our nginx configuration to serve the `*.pot` was still pointing at 5.2 and not 6.0.

[...]

PR #2733, opened 2026-07-23
Updated download page as per DEP-20.

sarahboyce:
See newly approved DEP-20: https://github.com/django/deps/pull/109

New download page:

Screenshot from 2026-08-07 12-48-32

Sample roadmap page:

Screenshot from 2026-08-07 12-52-13

As we may announce with a blog post, it would be nice if this page is updated accordingly

We perhaps want to time the updates alongside the following docs that need updating:
- https://docs.djangoproject.com/en/6.1/faq/install/#faq-python-version-support
- https://docs.djangoproject.com/en/6.1/internals/release-process/

### DIFF
--- a/djangoproject/templates/releases/download.html
+++ b/djangoproject/templates/releases/download.html
@@ -44,14 +44,14 @@
   <pre class="literal-block"><code>py -m pip install Django=={{ current.version }}</code></pre>
 
   {% if preview %}
-    {% with major_version=preview.version|slice:":3" %}
-      <h2>Option {% cycle options %}: Get the {{ preview.get_status_display }} for {{ major_version }}</h2>
+    {% with feature_version=preview.feature_version %}
+      <h2>Option {% cycle options %}: Get the {{ preview.get_status_display }} for {{ feature_version }}</h2>
       <p>As part of the
-        <a href="{% url 'roadmap' major_version %}">
-          Django {{ major_version }} development process</a>, Django
+        <a href="{% url 'roadmap' feature_version %}">
+          Django {{ feature_version }} development process</a>, Django
         {{ preview.version }} is available. This release is only for users who
         want to try the new version and help identify remaining bugs before the
-        {{ major_version }} release. Please read the
+        {{ feature_version }} release. Please read the
         {% release_notes preview.version show_version=True %} before using this package.
         <p>Install the {{ preview.get_status_display }} with <a href="https://pip.pypa.io/">pip</a>:</p>
       <pre class="literal-block"><code>pip install --pre django</code></pre>
@@ -70,16 +70,20 @@
   <p>And be sure to sign up to the <a href="https://forum.djangoproject.com">Django Forum</a>, where other Django users and the Django developers themselves all hang out to help each other.</p>
 
   <h2 id="supported-versions">Supported Versions</h2>
-  <p><strong>Feature releases</strong> (A.B, A.B+1, etc.) will happen roughly every eight months.
-    These releases will contain new features, improvements to existing features, and such.</p>
-  <p><strong>Patch releases</strong> (A.B.C, etc.) will be issued as needed, to
-    fix bugs and/or security issues. These releases will be 100% compatible with
+  <p><strong>Feature releases</strong> contain new features, improvements to existing
+    features, and other enhancements. Beginning with Django 2028, feature releases will
+    use the version format YYYY and will be released every January. Earlier feature
+    releases used the version format A.B and were released roughly every eight months.</p>
+  <p><strong>Patch releases</strong> (for example, 6.1.2 or 2028.3) are issued as
+    needed, to fix bugs and/or security issues. These releases are 100% compatible with
     the associated feature release, unless this is impossible for security
     reasons or to prevent data loss. So the answer to "should I upgrade to the
     latest patch release?” will always be "yes."</p>
-  <p>Certain feature releases will be designated as <strong>long-term support
-    (LTS) releases</strong>. These releases will get security and data loss
-    fixes applied for a guaranteed period of time, typically three years.</p>
+  <p>Django 6.2 is the final release under the previous versioning and support policy.
+    Under that policy, only designated long-term support (LTS) releases received the
+    full three-year support period. Starting with Django 2028, every feature release
+    receives the same three-year support period. Refer to the tables below for the
+    support periods.</p>
   <p>See the <a href="{% url 'document-detail' lang='en' version='dev' url='internals/release-process' host 'docs' %}#supported-versions">
     supported versions policy</a> for detailed guidelines about what fixes will be backported.</p>
 
@@ -95,19 +99,19 @@
     </tr>
     <tr>
       <td>5.2 LTS</td>
-      <td>{% get_latest_micro_release '5.2' %}</td>
+      <td>{% get_latest_release_version '5.2' %}</td>
       <td>December 3, 2025</td>
       <td>April 2028</td>
     </tr>
     <tr>
       <td>6.0</td>
-      <td>{% get_latest_micro_release '6.0' %}</td>
+      <td>{% get_latest_release_version '6.0' %}</td>
       <td>August 4, 2026</td>
       <td>April 2027</td>
     </tr>
     <tr>
       <td>6.1</td>
-      <td>{% get_latest_micro_release '6.1' %}</td>
+      <td>{% get_latest_release_version '6.1' %}</td>
       <td>April 2027</td>
       <td>December 2027</td>
     </tr>
@@ -129,22 +133,16 @@
       <td>April 2030</td>
     </tr>
     <tr>
-      <td><a href="{% url 'roadmap' '7.0' %}">7.0</a></td>
-      <td>December 2027</td>
-      <td>August 2028</td>
-      <td>April 2029</td>
+      <td><a href="{% url 'roadmap' '2028' %}">2028</a></td>
+      <td>January 2028</td>
+      <td>January 2029</td>
+      <td>January 2031</td>
     </tr>
     <tr>
-      <td><a href="{% url 'roadmap' '7.1' %}">7.1</a></td>
-      <td>August 2028</td>
-      <td>April 2029</td>
-      <td>December 2029</td>
-    </tr>
-    <tr>
-      <td><a href="{% url 'roadmap' '7.2' %}">7.2 LTS</a></td>
-      <td>April 2029</td>
-      <td>December 2029</td>
-      <td>April 2032</td>
+      <td><a href="{% url 'roadmap' '2029' %}">2029</a></td>
+      <td>January 2029</td>
+      <td>January 2030</td>
+      <td>January 2032</td>
     </tr>
   </table>
 
@@ -160,121 +158,121 @@
     </tr>
     <tr>
       <td>5.1</td>
-      <td>{% get_latest_micro_release '5.1' %}</td>
+      <td>{% get_latest_release_version '5.1' %}</td>
       <td>April 2, 2025</td>
       <td>December 3, 2025</td>
     </tr>
     <tr>
       <td>5.0</td>
-      <td>{% get_latest_micro_release '5.0' %}</td>
+      <td>{% get_latest_release_version '5.0' %}</td>
       <td>August 7, 2024</td>
       <td>April 2, 2025</td>
     </tr>
     <tr>
       <td>4.2 LTS</td>
-      <td>{% get_latest_micro_release '4.2' %}</td>
+      <td>{% get_latest_release_version '4.2' %}</td>
       <td>December 4, 2023</td>
       <td>April 7, 2026</td>
     </tr>
     <tr>
       <td>4.1</td>
-      <td>{% get_latest_micro_release '4.1' %}</td>
+      <td>{% get_latest_release_version '4.1' %}</td>
       <td>April 5, 2023</td>
       <td>December 1, 2023</td>
     </tr>
     <tr>
       <td>4.0</td>
-      <td>{% get_latest_micro_release '4.0' %}</td>
+      <td>{% get_latest_release_version '4.0' %}</td>
       <td>August 3, 2022</td>
       <td>April 1, 2023</td>
     </tr>
     <tr>
       <td>3.2 LTS</td>
-      <td>{% get_latest_micro_release '3.2' %}</td>
+      <td>{% get_latest_release_version '3.2' %}</td>
       <td>December 7, 2021</td>
       <td>April 1, 2024</td>
     </tr>
     <tr>
       <td>3.1</td>
-      <td>{% get_latest_micro_release '3.1' %}</td>
+      <td>{% get_latest_release_version '3.1' %}</td>
       <td>April 6, 2021</td>
       <td>December 7, 2021</td>
     </tr>
     <tr>
       <td>3.0</td>
-      <td>{% get_latest_micro_release '3.0' %}</td>
+      <td>{% get_latest_release_version '3.0' %}</td>
       <td>August 3, 2020</td>
       <td>April 6, 2021</td>
     </tr>
     <tr>
       <td>2.2 LTS</td>
-      <td>{% get_latest_micro_release '2.2' %}</td>
+      <td>{% get_latest_release_version '2.2' %}</td>
       <td>December 2, 2019</td>
       <td>April 11, 2022</td>
     </tr>
     <tr>
       <td>2.1</td>
-      <td>{% get_latest_micro_release '2.1' %}</td>
+      <td>{% get_latest_release_version '2.1' %}</td>
       <td>April 1, 2019</td>
       <td>December 2, 2019</td>
     </tr>
     <tr>
       <td>2.0</td>
-      <td>{% get_latest_micro_release '2.0' %}</td>
+      <td>{% get_latest_release_version '2.0' %}</td>
       <td>August 1, 2018</td>
       <td>April 1, 2019</td>
     </tr>
     <tr>
       <td>1.11 LTS <sup><a href="#ft3">3</a></sup></td>
-      <td>{% get_latest_micro_release '1.11' %}</td>
+      <td>{% get_latest_release_version '1.11' %}</td>
       <td>December 2, 2017</td>
       <td>April 1, 2020</td>
     </tr>
     <tr>
       <td>1.10</td>
-      <td>{% get_latest_micro_release '1.10' %}</td>
+      <td>{% get_latest_release_version '1.10' %}</td>
       <td>April 4, 2017</td>
       <td>December 2, 2017</td>
     </tr>
     <tr>
       <td>1.9</td>
-      <td>{% get_latest_micro_release '1.9' %}</td>
+      <td>{% get_latest_release_version '1.9' %}</td>
       <td>August 1, 2016</td>
       <td>April 4, 2017</td>
     </tr>
     <tr>
       <td>1.8 LTS</td>
-      <td>{% get_latest_micro_release '1.8' %}</td>
+      <td>{% get_latest_release_version '1.8' %}</td>
       <td>December 1, 2015</td>
       <td>April 1, 2018</td>
     </tr>
     <tr>
       <td>1.7</td>
-      <td>{% get_latest_micro_release '1.7' %}</td>
+      <td>{% get_latest_release_version '1.7' %}</td>
       <td>April 1, 2015</td>
       <td>December 1, 2015</td>
     </tr>
     <tr>
       <td>1.6</td>
-      <td>{% get_latest_micro_release '1.6' %}</td>
+      <td>{% get_latest_release_version '1.6' %}</td>
       <td>September 2, 2014</td>
       <td>April 1, 2015</td>
     </tr>
     <tr>
       <td>1.5</td>
-      <td>{% get_latest_micro_release '1.5' %}</td>
+      <td>{% get_latest_release_version '1.5' %}</td>
       <td>November 6, 2013</td>
       <td>September 2, 2014</td>
     </tr>
     <tr>
       <td>1.4 LTS</td>
-      <td>{% get_latest_micro_release '1.4' %}</td>
+      <td>{% get_latest_release_version '1.4' %}</td>
       <td>February 26, 2013</td>
       <td>October 1, 2015</td>
     </tr>
     <tr>
       <td>1.3</td>
-      <td>{% get_latest_micro_release '1.3' %}</td>
+      <td>{% get_latest_release_version '1.3' %}</td>
       <td>March 23, 2012</td>
       <td>February 26, 2013</td>
     </tr>
--- a/releases/management/commands/generate_release_roadmap.py
+++ b/releases/management/commands/generate_release_roadmap.py
@@ -1,28 +1,16 @@
 """
-Generates an SVG roadmap of Django releases,
-showing mainstream and extended support periods.
+Generates an SVG roadmap of Django releases, showing mainstream and extended
+support periods based on the hardcoded release schedule in RELEASES.
 
 Usage:
-  python -m manage generate_release_roadmap.py --first-release <VERSION> --date <YYYY-MM>
+  python -m manage generate_release_roadmap
 
-Arguments:
-    --first-release   First release number in Django versioning style, e.g.,"4.2"
-    --date            Release date of first release in YYYY-MM format, e.g.,"2023-04"
-
-Behavior:
-    - Automatically generates 8 consecutive Django releases:
-        X.0, X.1, X.2 (LTS), X+1.0, X+1.1, X+1.2 (LTS), X+2.0, X+2.1
-    - Mainstream support: 8 months per release
-    - Extended support:
-        - LTS releases (*.2) have 28 months of extended support beyond mainstream
-        - Non-LTS releases have 8 months of extended support beyond mainstream
-    - Produces an SVG at: ../djangoproject/static/img/release-roadmap.svg
+Produces an SVG at: ../djangoproject/static/img/release-roadmap.svg.
 """
 
 import datetime as dtime
 from pathlib import Path
 
-from dateutil.relativedelta import relativedelta
 from django.conf import settings
 from django.core.management.base import BaseCommand
 from jinja2 import Environment, FileSystemLoader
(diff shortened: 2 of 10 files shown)


djangoproject/templates/releases/download.html
```diff
@@ -95,19 +100,19 @@ <h2 id="supported-versions">Supported Versions</h2>
...
+      <td>{% get_latest_release_version '6.0' %}</td>
       <td>August 4, 2026</td>
       <td>April 2027</td>
     </tr>
     <tr>
       <td>6.1</td>
-      <td>{% get_latest_micro_release '6.1' %}</td>
+      <td>{% get_latest_release_version '6.1' %}</td>
       <td>April 2027</td>
       <td>December 2027</td>
```
jacobtylerwalls, line 116: We might decide to bump this to January 2028, as otherwise for a single month period there will be no version in mainstream support.

releases/management/commands/generate_release_roadmap.py
```diff
@@ -64,24 +53,65 @@
...
+    },
+    {
+        "name": "6.2",
+        "is_lts": True,
+        "release_date": dtime.date(2027, 4, 1),
+        "mainstream_end": dtime.date(2027, 12, 1),
+        "extended_end": dtime.date(2030, 4, 1),
+    },
+    {
+        "name": "2028.0",
```
jacobtylerwalls: I would generate the SVG with just "2028", since 2028.0 is only good for about 30 days.

djangoproject/templates/releases/download.html
```diff
@@ -70,16 +70,21 @@ <h2>After you get it</h2>
...
   <h2 id="supported-versions">Supported Versions</h2>
-  <p><strong>Feature releases</strong> (A.B, A.B+1, etc.) will happen roughly every eight months.
-    These releases will contain new features, improvements to existing features, and such.</p>
-  <p><strong>Patch releases</strong> (A.B.C, etc.) will be issued as needed, to
-    fix bugs and/or security issues. These releases will be 100% compatible with
+  <p><strong>Feature releases</strong> contain new features, improvements to existing
+    features, and other enhancements. Beginning with Django 2028.0, feature releases
+    will use the version format YYYY.0 and will be released every January. Earlier
+    feature releases used the version format A.B and were released roughly every eight
+    months.</p>
```
jacobtylerwalls: A.B is analogous to YYYY, I think, not YYYY.N

djangoproject/templates/releases/download.html
```diff
@@ -70,16 +70,21 @@ <h2>After you get it</h2>
...
+    needed, to fix bugs and/or security issues. These releases are 100% compatible with
     the associated feature release, unless this is impossible for security
     reasons or to prevent data loss. So the answer to "should I upgrade to the
     latest patch release?” will always be "yes."</p>
-  <p>Certain feature releases will be designated as <strong>long-term support
-    (LTS) releases</strong>. These releases will get security and data loss
-    fixes applied for a guaranteed period of time, typically three years.</p>
+  <p>Django 6.2 is the final release under the previous versioning and support policy.
+    Under that policy, only designated long-term support (LTS) releases received the
+    full three-year support period. Starting with Django 2028.0, every feature release
```
jacobtylerwalls: 2028 here as well?

djangoproject/templates/releases/download.html
```diff
@@ -70,16 +70,21 @@ <h2>After you get it</h2>
...
+    receives the same three-year support period. Refer to the tables below for the
+    support periods of individual release series.</p>
```
jacobtylerwalls: optional: just "support periods." ("release series" being its own plural is potentially distracting?)

releases/tests.py
```diff
@@ -817,6 +863,12 @@ def setUpTestData(cls):
                 ("rc1", datetime.date(2026, 7, 22)),
                 ("", datetime.date(2026, 8, 5)),  # final
             ],
+            "2028.0": [
+                ("a1", datetime.date(2027, 10, 1)),
+                ("b1", datetime.date(2027, 11, 1)),
+                ("rc1", datetime.date(2027, 12, 20)),
```
jacobtylerwalls: Nothing needs to change for the test, but just mentioning my understanding is that we'd do the RC much earlier in December, and grow the typical 2 week interval to final to 4.

releases/urls.py
```diff
@@ -7,5 +7,5 @@
     re_path(
         "^([0-9a-z_.-]+)/(tarball|wheel|checksum)/$", redirect, name="download-redirect"
     ),
-    re_path(r"^(?P<series>\d{1,2}\.[0-2])/roadmap/$", roadmap, name="roadmap"),
+    re_path(r"^(?P<series>\d+\.[0-2])/roadmap/$", roadmap, name="roadmap"),
```
jacobtylerwalls: I see we will update the roadmap template in another MR.

releases/tests.py
```diff
@@ -281,6 +319,8 @@ def test_version_tuple(self):
             ("1.8rc1", (1, 8, 0, "rc", 1)),
             ("5.2", (5, 2, 0, "final", 0)),
             ("5.2a1", (5, 2, 0, "alpha", 1)),
+            ("2028.0a1", (2028, 0, 0, "alpha", 1)),
+            ("2028.0", (2028, 0, 0, "final", 0)),
```
jacobtylerwalls: This fails
```suggestion
            ("2028", (2028, 0, 0, "final", 0)),
            ("2028.0", (2028, 0, 0, "final", 0)),
```

releases/management/commands/generate_release_roadmap.py
```diff
@@ -111,109 +141,47 @@ def calculate_dimensions(config: dict, num_releases: int) -> int:
...
-
-    # Month labels only for the VERY FIRST set of lines
-    FIRST_YEAR_MONTH_LABELS = {
-        0: None,
-        1: "April",
-        2: "August",
-        3: "December",
-    }
+
+    month_lines = (
```
jacobtylerwalls, line 159: optional: We could stop drawing lines on the chart after 6.2 goes EOL?

jacobtylerwalls (commented):
Marvelous, you found the right amount of changes to tackle first IMO!

djangoproject/templates/releases/download.html
jacobtylerwalls: In other words, you can keep "Beginning with Django 2028.0" if you wish, but "feature releases ... YYYY.0" and "feature releases ... A.B" don't seem analogous.

sarahboyce: I think where I am slightly struggling is we have a 6.2 release being something you can pip install but not 2028, we release 2028.0 but we released 6.2 (if that makes sense)

releases/tests.py
sarahboyce: Does that one need to be handled? I think we won't release `"2028"`

djangoproject/templates/releases/download.html
jacobtylerwalls: I see. I didn't think about pip. I still think a reader might have my confusion, so what are your thoughts about:
```
Beginning with Django 2028.0, feature releases will use YYYY[.N]
Earlier feature releases used A.B[.C]
```

releases/tests.py
jacobtylerwalls: Right, good point.

djangoproject/templates/releases/download.html
carltongibson: Not sure I follow the pip point: 

```
❯ pip install 'Django==6.*'
Collecting Django==6.*
  Using cached django-6.1-py3-none-any.whl.metadata (3.9 kB)
Requirement already satisfied: asgiref>=3.9.1 in ./lib/python3.12/site-packages (from Django==6.*) (3.12.1)
Requirement already satisfied: sqlparse>=0.5.0 in ./lib/python3.12/site-packages (from Django==6.*) (0.5.5)
Using cached django-6.1-py3-none-any.whl (8.4 MB)
Installing collected packages: Django
Successfully installed Django-6.1
```

So I'm imagining the instruction to be: 

```
pip install Django==2028.*
```

Which will get you the latest patch release. 

But the feature version should be YYYY, rather than YYYY.N (where N is the patch number) I think 🤔

carltongibson (commented):
This is lovely. 👍 

I think the only quibble is about not using the .0 to refer to the feature release: YYYY being equivalent to X.Y now, with YYYY.N equivalent to X.Y.Z. 

(Yesterday's release was 6.1.0, even though we only ever say "6.1")

djangoproject/templates/releases/download.html
sarahboyce: I was confused by how we usually omit the .0 but have discussed the .0 in the DEP
Screenshot from 2026-08-07 09-59-47


Do we think we will have the same logic that we omit the .0 when displaying the feature release?
I think this logic is in `django.utils.version.get_main_version()` for example

carltongibson: I think that's the goal. We want to be talking about "Django 2028" or "Django 2030", where "2023.5" is for talking about the patch release.

Like, strictly "6.1" **is** "6.1.0" right? We just don't speak that way.

djangoproject/templates/releases/download.html
carltongibson: I think you mean the [6.2 row below](https://github.com/django/djangoproject.com/pull/2733/changes/BASE..126def378ee1b833b58ad4060b7331112a37cea4#diff-9ece7fd6d5728200f2be6ed04905593d65933c02c891648158db11232c915f31R133) 🤔

I'd likely say this is OK as it is: 6.2 **will** get its final _mainstream release_ at the beginning of December. At that point it's still covered for security releases, and the new version is just about to arrive. Nobody needs anything from a beginning of Jan release beyond that. 🤷

releases/urls.py
sarahboyce: I updated this a little bit

releases/management/commands/generate_release_roadmap.py
sarahboyce: I think it looks a little nicer with consistent lines but we can simplify the graph quite significantly once 5.2 is EOL in April 2028

releases/models.py
```diff
@@ -308,10 +310,18 @@ def is_pre_release(self):
         """Return True if this is an alpha, beta, or rc release."""
         return self.status != "f"
 
+    @cached_property
+    def is_year_version(self):
```
jacobtylerwalls: In a future MR we could add a CheckConstraint that micro == 0 if major > 999, to prevent data entry/test fixture mistakes.

releases/views.py
```diff
@@ -32,7 +32,9 @@ def index(request):
 
 
 def roadmap(request, series):
-    major, minor = series.split(".")
+    major, separator, minor = series.partition(".")
```
jacobtylerwalls: Nice. If the roadmap template is in scope now, you might adjust the RC section to not say "two weeks", maybe like:

```
if is_year_version
    will be released in January
```
or sim.

jacobtylerwalls (approved):
👍

carltongibson (approved):
I think this is great. (No doubt we'll make more tweaks, but this is the most of it for sure!) 👍

releases/management/commands/generate_release_roadmap.py
```diff
@@ -64,24 +53,65 @@
     "month_line_width": 1,
 }
 
+RELEASES = [
```
pauloxnet, line 58: Worth adding a note somewhere obvious that this needs a manual update each
January, otherwise it will silently go stale.

releases/models.py
```diff
@@ -308,10 +310,18 @@ def is_pre_release(self):
...
+        return len(str(self.major)) == 4
```
pauloxnet:
```suggestion
        return self.major >= 2028
```
Would `major >= 2028` be safer than just checking for four digits? Avoids any
weird edge cases with test fixtures using arbitrary version numbers.

pauloxnet (approved):
Great work @sarahboyce ! Left a few small and non blocking comments below.

---

P.S. @carltongibson one thing I noticed looking at the SVG preview: between January and April 2030 we'll briefly have 4 versions supported at the same time: 6.2, 2028, 2029 and 2030. That's could means a few extra months of overhead for the fellows, the security team and third-party maintainers.

Since 6.2 hasn't been released yet, would it make sense to shorten its support window to December 2029 instead of April 2030? No one is relying on that date yet so nothing would be broken, and we'd stay at three concurrent versions throughout the transition, which feels cleaner.

carltongibson:
@pauloxnet The DEP's plan is to leave 6.2 support commitment unchanged. I think that's right. (The correct thing to do.) 

I'll raise it with the Fellows though to discuss!

releases/models.py
sarahboyce: Good idea :+1: 
I also like how this documents which release was the first release of this format

releases/management/commands/generate_release_roadmap.py
sarahboyce: I've updated the file's docstring and I've now added a couple of TODOs :+1: 
We will be able to simplify the logic and make this management command generate the SVG dynamically again once we're out of the transition period and we have a regular rhythm again

djangoproject/templates/releases/download.html
nessita: Isn't what Jacob is saying that even when we fully switch to YYYY versions (i.e. no 6.2 at play), there will be a month without any version with mainstream support? i.e. 2028 is EOM in December 2028 and 2029 is released ~45 days later (mid Jan 2029)? Feels like perhaps YYYY maintenance support should be dropped on Jan next year (ie cover the whole December too)?

jacobtylerwalls: Yes. However, I understand Carlton's view to be that it's not an issue if there's a month with no regression fixes. 8 months of regression fixes being enough, no reason to stretch the window to 9 months over the holidays. I can dig it. I just thought the diagram looked weird without it, and if we _close the gap_ it won't _look like an oversight_.

We're in what would have been the ninth month of mainstream support for 6.0, and nothing has come up so far 🤷

jacobtylerwalls: > (i.e. no 6.2 at play)

Sorry, now that I re-read your question, no, I did mean the one-time-only gap in from December 2027 to January 2028 (the gap on the graph).

> Feels like perhaps YYYY maintenance support should be dropped on Jan next year (ie cover the whole December too)?

I think Sarah's changes already show that.

pauloxnet:
BTW reviewing this PR I had some idea to improve the release graph that I wrote down in the issue #2758

djangoproject/templates/releases/download.html
nessita: Well, it depends on how you read it. The download image does show the yearly releases "ending" at the end of the year, right(ish) when the next one starts. Given our current semantics, I think we should say YYYY ends mainstream support in "Jan YYYY+1". For me that is clearer to reason about (and that speech would match the graph).

djangoproject/templates/releases/download.html
nessita: This component will need update too, `slice:":3"` gives "202" for a version like `2028a1`, so this heading reads "the alpha for 202" and the `{% url 'roadmap' %}` below raises `NoReverseMatch`, taking down the whole download page during the pre-release cycle. I will push a fix using `preview.feature_version` instead.

djangoproject/templates/releases/download.html
jacobtylerwalls: I think we should at least update the blog post to show EOLs in January, not December.

nessita: Yeap agreed!

releases/models.py
nessita: I think we should name this after the scheme, not the cadence. "Calendar version" is the term the DEP and the blog post use, and it stays accurate if we ever move to two feature releases a year (the DEP leaves `YYYY.M.N` open), whereas "year version" would quietly become wrong while the check stayed correct:

```suggestion
    def is_calendar_version(self):
```

(I'll push this and other changes in a new commit for your evaluation.)

releases/models.py
```diff
@@ -263,7 +263,7 @@ def version_tuple(self):
         """Return a tuple in the format of django.VERSION."""
         version = self.version.replace("-", "").replace("_", "")
         version = list(get_loose_version_tuple(version))
-        if len(version) == 2:
+        while len(version) < 3:
```
nessita: I think this method needs more re-work since as is it doesn't survive the calendar pre-releases. `version_tuple` identifies the components by position, which assumes two numbers before the status letter, so `2028a1` parses as `(2028, 'a', 1, 'final', 0)` and saving it fails with "Field 'minor' expected a number but got 'a'".

In my optional commit with suggestions, I've moved the parsing to `get_version_tuple()` in `releases/utils.py`, which splits the leading numbers from the status and its iteration instead of patching
the list by index, and pads the numbers to three. Same results for everything in the A.B[.C] form, and `2028a1` now gives `(2028, 0, 0, 'alpha', 1)`. Tests cover the round trip between every version string we publish and its tuple, in both schemes.

releases/templatetags/release_notes.py
nessita: The following may be a little of scope creep but as this stands, it does not work for the new calendar versions: it takes the first two numeric components, so `2028.3` links to `/en/2028.3/releases/2028.3/` and `2028a1` to `/en/2028.0/releases/2028.0/`, neither of which exists. What it actually wants is
the feature version, which is the same thing `Release.feature_version` needs, so I've pulled that into `get_feature_version()` in `releases/utils.py` and used it here. Patch releases now point at `/en/2028/releases/2028.3/` and pre-releases at `/en/2028/releases/2028/`, matching what we do for 5.2/6.0/6.1 etc.

releases/urls.py
```diff
@@ -7,5 +7,5 @@
     re_path(
         "^([0-9a-z_.-]+)/(tarball|wheel|checksum)/$", redirect, name="download-redirect"
     ),
-    re_path(r"^(?P<series>\d{1,2}\.[0-2])/roadmap/$", roadmap, name="roadmap"),
+    re_path(r"^(?P<series>(?:\d+\.[0-2]|\d{4}))/roadmap/$", roadmap, name="roadmap"),
```
nessita: Making the left part of the regex `\d+` also matches `2028.0`, so `/download/2028.0/roadmap/` becomes a second live URL for the same page as `/download/2028/roadmap/`. Since the calendar series are covered by the `\d{4}` branch, I've put the A.B branch back to `\d{1,2}`.

(And we dont want to support YYYY.N links for roadmaps)

nessita (commented):
I've pushed some changes as a separated commit, and left inline comments on most of
them. Two things drove the set as a whole.

First, the download page keeps asking the same question in different places: given a version string, which feature release does it belong to? Every place answered it on its own, by slicing the string or indexing the parsed components, and each of those breaks differently on `YYYY[.N]`. I've pulled the shared pieces into `releases/utils.py` (`get_version_tuple()`, `get_feature_version()`, `get_main_version()`, `is_calendar_version()`), so the model, the template tag and the views now give the same answer. That's what makes the individual fixes below one-liners.

Second, once the site can hold a 2028 release, the code around the page has to cope with one. I created a 2028 release locally to check the tables and hit the rest. The one I haven't commented on inline is artifact naming: `get_version()` drops a trailing zero from the third component, so `(2028, 0, 0, 'final', 0)` renders as "2028.0", resulting in the admin rejecting `Django-2028.tar.gz` and the checksum lands at `pgp/Django-2028.0.checksum.txt`. `get_main_version()` now picks the component to drop based on the scheme, and tests cover the round trip between every version string we publish and its tuple.

I've also respelled the pre-releases in the test data from `2028.0a1` to `2028a1`: `YYYY` is to `YYYY.N` what `A.B` is to `A.B.C`, and we ship 6.1a1 rather than 6.1.0a1. The parser needed fixing for that to be storable at all. (I also discussed this in slack with the SC.)

None of this is live until a calendar-versioned release exists, but it seemed better to land it with the change that introduces them than to leave the site in a state where creating one breaks pages.

releases/tests.py
```diff
@@ -325,14 +336,28 @@ def test_version_tuple(self):
...
                 release = Release.objects.create(version=version)
                 self.assertEqual(release.version_tuple, expected)
 
+    def test_invalid_version(self):
+        # Malformed versions are rejected when saving rather than when parsing,
+        # as get_version_tuple() passes unknown statuses and extra numeric
+        # components through.
+        cases = [
+            ("5.2dev1", KeyError, "'dev'"),
+            ("1.2.3.4", ValueError, "too many values to unpack"),
```
jacobtylerwalls: I think we should relax the specificity of these assertions.

releases/tests.py
```diff
@@ -512,6 +626,19 @@ def test_upload_to_artifact(self):
                 "django-5.2b2-py3-none.whl",
                 "releases/5.2/django-5.2b2-py3-none.whl",
             ),
+            # All the releases in a year series share a directory.
+            ("2028", "django-2028.tar.gz", "releases/2028/django-2028.tar.gz"),
+            ("2028.1", "django-2028.1.tar.gz", "releases/2028/django-2028.1.tar.gz"),
```
jacobtylerwalls: My inclination would have been to explore the release/artifact/documentrelease changes in a separate stream of MRs, and possibly after debugging https://github.com/django/djangoproject.com/issues/2373 first.

If we take this now, I'll want to do some manual testing myself before the next security release, whichever month that turns out to be.

releases/models.py
sarahboyce: This is a better name :+1:

djangoproject/templates/releases/download.html
sarahboyce: > I think we should at least update the blog post to show EOLs in January, not December.

I agree

I feel the graph is ok. It shows that 2028 mainstream support ends when 2029 is released (which is true) that the year line means mid January rather than the first of January I feel we can live with given we have a table below with specific dates and we already labeled the line before with "December" and it is clearly after that?

releases/tests.py
```diff
@@ -871,3 +1120,31 @@ def test_links_to_contributing_and_release_process_present(self):
         response = self.client.get(url)
         self.assertContains(response, 'en/dev/internals/contributing/"')
         self.assertContains(response, 'en/dev/internals/release-process/"')
+
+    def test_release_candidate_description_year_version(self):
```
sarahboyce:
```suggestion
    def test_release_candidate_description_calendar_version(self):
```
I think as `is_year_version` got renamed to `is_calendar_version` we should update the test names where it says "year" to "calendar"

djangoproject/templates/releases/download.html
carltongibson: >  Carlton's view to be that it's not an issue if there's a month with no regression fixes. 8 months of regression fixes being enough, no reason to stretch the window to 9 months over the holidays.

>  I think we should say YYYY ends mainstream support in "Jan YYYY+1"

OK, so the DEP explicitly went for _strictly 35 months_ in order to say that mainstream support ends at the end of the third year, not rolling into the January. 

I don't think the extra month matters. 

Except nobody needs that **13th** (Jan to Dec inclusive is 12 already) maintenance release in the first week of January, and having been in the role of making that January release, it's a much more human policy to just say, you know what folks, we're not making that commitment over the new year holiday period. Security (and major bugs) is still covered, and the brand new version is two weeks away: you can cope. 

When I was Fellow, I would have much preferred a stance like that, and I assume that's true for you all, and all future Fellows too. No? 🙂

I think we should go with the DEP as written, both for this, and for the 6.2 timeline: Let it be out of mainstream support in the December: the new version is coming January. (And again, it's still in extended support, so security releases etc still happen)

WDTY?

carltongibson:
@sarahboyce @nessita @jacobtylerwalls I'm trying to get clear on what's outstanding, dividing between the narrow topic of the PR "Update the download page to show the new roadmap" and the other bits which are "Make sure the site is 100% 2028 ready"? (Some of which are https://code.djangoproject.com/ticket/37271 &co related)

releases/tests.py
nessita: Excellent point, I addressed this and pushed fixes in my commit.

releases/tests.py
nessita: I was feeling a bit lost by not being able to test the whole cycle (release artifact -> download page showing it -> roadmap page) so I needed these fixes and in my view these all are "download page components". This is how I found the "download preview" issue.

I can try to separate them in a second isolated commit if you prefer?

releases/tests.py
nessita: Done.

sarahboyce:
> That's could means a few extra months of overhead for the fellows, the security team and third-party maintainers.
>
> 
> Since 6.2 hasn't been released yet, would it make sense to shorten its support window to December 2029 instead of April 2030? No one is relying on that date yet so nothing would be broken, and we'd stay at three concurrent versions throughout the transition, which feels cleaner.

@pauloxnet thank you for pointing this out :star: 
We discussed this among the Fellows, and we're comfortable supporting more releases than usual during this period. We feel that keeping the full three-year support period for the 6.2 LTS provides a smoother transition and migration experience for our user base, while also giving us one less deviation from the norm to communicate :+1:

djangoproject/templates/releases/download.html
nessita: We discussed this among the Fellows, and our preference is to move the support transitions to January, tied to the new feature release.

The main reason is consistency and clarity: keeping the transitions tied to feature releases preserves the existing release continuum. So the release of Django YYYY marks the end of mainstream support for YYYY-1 and the end of extended support (EOL) for YYYY-3.

We understand the motivation of reducing January releaser work, but from our perspective the practical saving is limited, since December/early January may still involve security releases and final-release preparation. On balance, we prefer the cleaner January boundary.

releases/tests.py
```diff
@@ -456,6 +626,19 @@ def test_upload_to_artifact(self):
                 "django-5.2b2-py3-none.whl",
                 "releases/5.2/django-5.2b2-py3-none.whl",
             ),
+            # All the releases in a calendar series share a directory.
+            ("2028", "django-2028.tar.gz", "releases/2028/django-2028.tar.gz"),
```
jacobtylerwalls, line 630: Since we're tackling the artifact logic in this PR now, I think we need negative test cases to catch cases like this:
```py
        # Invalid cases.
        for version, filename in [
            ("2028", "django-2028.1.tar.gz"),
        ]:
            with self.assertRaises(ValidationError):
                upload_to_artifact(Release(version=version), filename=filename),
```
```py
FAIL: test_upload_to_artifact (releases.tests.ReleaseUploadToTestCase.test_upload_to_artifact)
  File "/Users/jwalls/djangoproject.com/releases/tests.py", line 653, in test_upload_to_artifact
    with self.assertRaises(ValidationError):
         ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
AssertionError: ValidationError not raised
```

releases/models.py
```diff
@@ -308,10 +299,19 @@ def is_pre_release(self):
...
         return self.status != "f"
 
+    @cached_property
+    def is_calendar_version(self):
+        """Return True if this release is calendar versioned (YYYY[.N])."""
+        return self.major >= FIRST_CALENDAR_VERSION_YEAR
+
     @cached_property
     def is_dot_zero(self):
         """Return True if this is a final X.Y.0 release."""
```
jacobtylerwalls: May as well update the docstring.

jacobtylerwalls (commented):
I'm comfortable with the increased scope: the changes in releases/models.py have a smaller blast radius than I expected. One requested test & fix and I think this is g2g.

releases/tests.py
jacobtylerwalls: Thanks for offering, no need. After taking a another look today, the changes have a smaller impact than I expected, so my comment was a little premature.

nessita (approved):
This looks great and it's very exciting!

I will merge and deploy tomorrow if no one objects.

PR #2742, opened 2026-07-27
Add test to detect production staticfiles issues

### LINKED ISSUE #1831: Add a test (or a github action) to catch static files issues early

adamzap:
Since we use `ManifestStaticFilesStorage` in production, `collectstatic` can fail on a nonexistent file reference that it attempts to rewrite.

This patch adds a test that runs `collectstatic` with `ManifestStaticFilesStorage` to ensure that it doesn't fail before a production deploy.

Fixes #1831.

A few notes:

1. We get some extra output when the test fails: "Post-processing 'js/lib/jquery.min.js' failed!". I think this is ok since it only happens in the case of a failure and it seems to me that suppressing this output would increase complexity more than it's worth. I'm open to it if others want to go that route though.
2. I like this approach more than #1836, #1837, and #1855 because this test can fail in a local development environment, not just in CI.


djangoproject/tests.py
```diff
@@ -253,6 +254,23 @@ def test_sitemap_renders(self):
         self.assertEqual(response.status_code, 200)
 
 
+class ProductionSettingsTests(TestCase):
```
adamzap, line 257: I'm open to naming suggestions for the test class and the test method.

djangoproject/tests.py
```diff
@@ -253,6 +254,23 @@ def test_sitemap_renders(self):
...
+            }
+        }
+    )
+    def test_collectstatic_with_manifeststaticfilesstorage(self):
+        with tempfile.TemporaryDirectory() as temp_dir_path:
+            with override_settings(STATIC_ROOT=temp_dir_path):
+                try:
+                    call_command("collectstatic", "--no-input", "--verbosity=0")
+                except ValueError as e:
+                    self.fail(e)
```
adamzap, line 271: I caught the `ValueError` for two reasons:

1. A failure seems better than an error
2. We can be sure the command failed in, at least to some extent, the way we expected

djangoproject/static/js/lib/jquery.min.js
```diff
@@ -2,3 +2,4 @@
...
+//# sourceMappingURL=jquery.min.map
```
adamzap, line 5: This change will be removed before merge.

marksweb: This file is coming up in a test failure 🤔 

`test_collectstatic_with_manifeststaticfilesstorage`

adamzap: The second commit is there to show that the test fails on a bad reference like this.

djangoproject/tests.py
sarahboyce:
```suggestion
class StaticFilesTests(TestCase):
```
Perhaps? I tried to experiment testing the prod settings but importing these raises a `ValueError`
So maybe we can rename this if we add further tests to cover production settings

djangoproject/tests.py
```diff
@@ -253,6 +254,23 @@ def test_sitemap_renders(self):
...
+            "staticfiles": {
+                "BACKEND": "django.contrib.staticfiles.storage.ManifestStaticFilesStorage"
+            }
+        }
+    )
+    def test_collectstatic_with_manifeststaticfilesstorage(self):
+        with tempfile.TemporaryDirectory() as temp_dir_path:
+            with override_settings(STATIC_ROOT=temp_dir_path):
+                try:
+                    call_command("collectstatic", "--no-input", "--verbosity=0")
```
sarahboyce:
```suggestion
                    call_command("collectstatic", interactive=False, verbosity=0)
```

sarahboyce (approved):
Thank you for this! I like the approach ⭐

djangoproject/tests.py
adamzap: Ah yes, this is better. Thanks!

adamzap: Changed in 69321f93.

djangoproject/tests.py
adamzap: Thanks! Changed in 69321f93.

adamzap:
@sarahboyce Thanks for the review! This should be ready.

PR #2767, opened 2026-08-11
Removed uncessary returning of stripe object in Fundraising tests.

p-r-a-v-i-n:


fundraising/tests/test_views.py
```diff
@@ -369,8 +369,7 @@ def setUp(self):
     def stripe_data(self, filename):
         file_path = settings.BASE_DIR / f"fundraising/test_data/{filename}.json"
         with file_path.open() as f:
-            data = json.load(f)
-            return stripe.convert_to_stripe_object(data, stripe.api_key, None)
```
p-r-a-v-i-n, line 373: I didn't understand the reason why we were previously coverting it back to stripe object.

ulgens:
@p-r-a-v-i-n Does the fix/change in the code works with the current version? If so, I'd recommend making a PR with only the fix and let the automated systems handle the rest.

p-r-a-v-i-n:
> @p-r-a-v-i-n Does the fix/change in the code works with the current version? If so, I'd recommend making a PR with only the fix and let the automated systems handle the rest.

Yes. that make sense , @ulgens thanks :)

ulgens:
Thanks for the update.

For updating the branch history, please always prefer `rebase`. `merge main into ...` approach can cause unexpected issues with the git history.

p-r-a-v-i-n:
I agree. Even I also prefer `rebase`.  I will keep in mind `not to click on update` .

marksweb:
Yes it's quite annoying you can't change the update options like you can change the merge options.

jacobtylerwalls (approved):
It looks like all feedback has been addressed. @nessita and I are testing some new ops scripts and would like to have a new hash on main, so I'm taking the liberty to merge this one, hope that's ok!

PR #2771, opened 2026-08-12
✏️ Update fundraising page

ryancheley:
Update to the fundraising page to:

1. Add link to Sponsor page
2. Reorder 'Other ways to give' to prioritize most helpful at top
3. Update the Why give to the DSF section with more modern example

### DIFF
--- a/djangoproject/templates/fundraising/includes/donation_form_with_heart.html
+++ b/djangoproject/templates/fundraising/includes/donation_form_with_heart.html
@@ -21,7 +21,7 @@
 
     <div class="description">
       <ul>
-        <li><strong>{% blocktranslate trimmed %}{{ goal_percent }}% funded{% endblocktranslate %}</strong></li>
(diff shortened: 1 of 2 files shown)


djangoproject/templates/fundraising/includes/donation_form_with_heart.html
```diff
@@ -21,7 +21,7 @@
 
     <div class="description">
       <ul>
-        <li><strong>{% blocktranslate trimmed %}{{ goal_percent }}% funded{% endblocktranslate %}</strong></li>
+        <li><strong>{% blocktranslate trimmed %}{{ goal_percent }}% funded{% endblocktranslate %}</strong> (<a href="/sponsor/">Why Sponsor?</a>)</li>
```
jefftriplett, line 24: Just noting, we used a flat page to get this page up quickly. It might be worthwhile, in a follow-up ticket, to move this page back to a template. Then we can link to page references directly. 

Nothing for you to do here. Just commenting so we can turn this into a follow-up issue.

jefftriplett (approved):
LGTM. I think we will want to do a follow-up, much deeper content cut.

PR #2773, opened 2026-08-15
Update `black` from v25.12.0 to v26.5.1

ulgens:


adamzap (approved):
👍 Looks great!

This is the same version `django/django` uses: https://github.com/django/django/blob/3436cf9bce84bb1f6877ad96819637366b27b719/.pre-commit-config.yaml#L3.

PR #2775, opened 2026-08-15
Enable ruff

ulgens:
The goal is to introduce ruff to the lint setup with minimal config & changes

---

Softer version of #2499

PR #2776, opened 2026-08-15
Switch to ruff's isort

ulgens:
Following up #2775

---

Depends on:
- #2775

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -51,11 +51,6 @@ repos:
             exclude_types: [html, json, scss]
             exclude: '(^djangoproject\/static\/js\/lib\/.*$|^checklists\/templates\/.*$)'
 
-    - repo: https://github.com/pycqa/isort
-      rev: 7.0.0
-      hooks:
-          - id: isort
-
     - repo: https://github.com/pycqa/flake8
       rev: 7.3.0
       hooks:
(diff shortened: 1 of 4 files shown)


pauloxnet (approved):

SaptakS:
@ulgens can you resolve the conflicts here?

ulgens:
@SaptakS Done.

.isort.cfg
adamzap, line 1: @ulgens Could you help me understand why this file is needed?

adamzap:
@ulgens There is a reference to `isort` in the project's README. Do you think it should be updated?

.isort.cfg
ulgens: To keep isort config.

https://isort.readthedocs.io/en/latest/configuration/config_files.html

The existings setup.cfg is an outdated format from setuptools times. `.isort.cfg` is the current top recommendation by the isort docs and I don't prefer using pyproject.toml for tool settings because it gets bloated very quickly.

ulgens:
> @ulgens There is a reference to `isort` in the project's README. Do you think it should be updated?

Thanks for the pointer. Updated.

.isort.cfg
adamzap: Oh, I assumed we were dropping `isort` for `ruff` here. Is that not the case? Are you thinking that `ruff` is using `.isort.cfg`? If so, why are there changes to imports in this PR?

ulgens: > Are you thinking that ruff is using .isort.cfg? 

I was 😬 Turned out it doesn't read either setup.cfg or .isort.cfg. Moved the settings to ruff.toml

marksweb (approved):
Looks good, keeps it simple 👍🏼

adamzap (approved):
👍 Looks great!

PR #2777, opened 2026-08-16
Fix general index hierarchy styling on docs pages.

### LINKED ISSUE #2769: CSS for general index page flattens hierarchy
A local docs build correctly uses indentation to show general index entries:

Image

However, on the site, everything is flush, left, leading to things like 'Suckiness of XML (Spatial Reference)', which is a mash-up of a design philosophy opinion with a GeoDjango attribute:

Image

https://docs.djangoproject.com/en/6.1/genindex/#X

anujharne:
Add dl.index to the definition list styles so sub-entries are indented and main entries are bold, matching other Sphinx doc lists.

Fixes #2769.

### DIFF
--- a/djangoproject/scss/_style.scss
+++ b/djangoproject/scss/_style.scss
@@ -2606,7 +2606,8 @@ dl.classmethod,
 dl.staticmethod,
 dl.attribute,
 dl.exception,
-dl.data {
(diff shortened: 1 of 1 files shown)


anujharne:
Fix it succesfully

PR #2778, opened 2026-08-16
Convert /sponsor/ flat page to a template view

### LINKED ISSUE #2772: [sponsor efforts] :tractor: Convert /sponsor/ Flat Page into a template url + view
Just noting, we used a flat page to get this page up quickly. It might be worthwhile, in a follow-up ticket, to move this page back to a template. Then we can link to page references directly. 

Nothing for you to do here. Just commenting so we can turn this into a follow-up issue.

_Originally posted by @jefftriplett in https://github.com/django/djangoproject.com/pull/2771#discussion_r3770093385_

anujharne:
## Summary

- Move the sponsor prospectus from a flat page into `fundraising/sponsor.html` with a dedicated view and URL at `/sponsor/`.
- Update fundraising page links to use `{% url 'sponsor' %}` so section anchors can be referenced from templates.
- Add a data migration to remove the old flat page entry and include the page in the sitemap.

Fixes #2772.

## Test plan

- [x] Visit `/sponsor/` and confirm the prospectus renders with section anchor links (e.g. `#corporate-membership-tiers`).
- [x] Confirm `/fundraising/` links to the sponsor page via Prospectus and Why Sponsor?
- [x] Run `python manage.py test fundraising.tests.test_views.TestSponsor`.


nessita (changes requested):
Hello @anujharne, thank you for your contribution!

The intent is right but there are a few things that need improvement:
1. The migration deleting the flatpage should be removed. We want to do that manually.
2. The sponsor page itself should use the project's CSS styles, and have no inline style at all. If a new CSS rule is needed, that's fine, but we should try to reuse what's already there.
3. Be sure that the new page works with dark/light mode, and with large/small screens.

Thanks again!

jefftriplett:
@anujharne thank you for picking this up! 

I'm going to make a few changes and see if we can get this merged at the DjangoCon US sprints today.

jefftriplett (approved):
Looks good. I removed the migrationa, addressed lint issues, updated the tables + styles via @nessita 

The goal was to get the flatpage in a state we can work with during the sprints and I think we achieved that goal.

PR #2779, opened 2026-08-17
 Switch to ruff's pyupgrade

ulgens:
Depends on:
- #2776

### DIFF
--- a/docs/management/commands/build_doc_release.py
+++ b/docs/management/commands/build_doc_release.py
@@ -138,6 +138,7 @@ class Command(BaseCommand):
                 # Clean up global state after building each language.
                 _clean_up_global_state()
             except SphinxError as e:
+                # FIXME: The following block creates duplicate errors in Sentry
                 capture_sentry_exception(e, flush=True)
                 raise CommandError(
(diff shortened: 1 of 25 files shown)


marksweb (approved):

docs/management/commands/build_doc_release.py
```diff
@@ -137,10 +138,11 @@ def build_doc_release(self, release):
                 # Clean up global state after building each language.
                 _clean_up_global_state()
             except SphinxError as e:
+                # FIXME: The following block creates duplicate errors in Sentry
                 capture_sentry_exception(e, flush=True)
```
adamzap, line 142: @ulgens I don't think the error is technically duplicated because this line sends the original `SphinxError` to Sentry, then the `CommandError` is also sent with details from the `SphinxError`. I'm not able to inspect them in Sentry, so it's hard to know if having both is helpful.

I wonder if this was intentional based on [this comment](https://github.com/django/djangoproject.com/pull/2688#pullrequestreview-4619672055).

Either way, I wonder if adding this comment in this PR is a bit out of scope. Would you be open to opening an issue instead if it needs further discussion?

PR #2780, opened 2026-08-17
Update git hooks

ulgens:
* Supersedes #2453
* Testing dependabot's pre-commit ecosystem

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -19,3 +19,13 @@ updates:
           timezone: "Europe/Oslo" # hipp hipp hurra
       cooldown:
           default-days: 7
+
+    - package-ecosystem: "pre-commit"
+      directory: /
+      schedule:
+          interval: "weekly"
+          day: "monday"
+          time: "12:00"
+          timezone: "Europe/Oslo"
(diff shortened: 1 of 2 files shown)


marksweb (approved):

adamzap:
@ulgens Could you update this now that #2776 has been merged?

ulgens:
@adamzap Done.

.github/dependabot.yml
```diff
@@ -19,3 +19,13 @@ updates:
           timezone: "Europe/Oslo" # hipp hipp hurra
       cooldown:
           default-days: 7
+
+    - package-ecosystem: "pre-commit"
+      directory: /
+      schedule:
+          interval: "weekly"
```
marksweb, line 26: I'd do this less frequently to make maintenance easier

marksweb (approved):

.github/dependabot.yml
ulgens: I can change but right now it matches the configs from other ecosystems and I don't think it will be noisy. We can revisit in future.

SaptakS (approved):

PR #2783, opened 2026-08-18
Migrate email configuration to MAILERS

### LINKED ISSUE #2768: Migrate to MAILERS before Django 7
This project sends email, thus we need to migrate to `MAILERS` before Django 7 (Django 2028).

[Migration guide](https://docs.djangoproject.com/en/dev/howto/mailers-migration/)

The development settings have at least one setting to migrate...

```py
% python -Wall manage.py test
/Users/.../djangoproject.com/manage.py:18: RemovedInDjango70Warning: The EMAIL_BACKEND setting is deprecated. Migrate to MAILERS before Django 7.0.
```

...but even if prod relies on Django's implicit defaults, we need to migrate because in Django 7, the `MAILERS` value will default to `{}` (no mail).

@tobiasmcnulty mentioned this morning in an ops call that we might wish to segment some of the emails we send into different backends (e.g. trac? account registration?)

Arbaaz123676:
## Description

Migrates the project's email configuration from the deprecated `EMAIL_BACKEND` setting to Django's `MAILERS` configuration for Django 7.

### Changes

- Explicitly configure the default SMTP email backend through `MAILERS`.
- Migrate the development console email backend from `EMAIL_BACKEND` to `MAILERS`.
- Preserve the existing email backend behavior.

### Testing

- Targeted pre-commit checks pass for all modified files.
- `git diff --check` passes.

Fixes #2768

### DIFF
--- a/djangoproject/settings/common.py
+++ b/djangoproject/settings/common.py
@@ -54,6 +54,13 @@ DATABASE_ROUTERS = ["tracdb.db_router.TracRouter"]
 DEFAULT_FROM_EMAIL = "noreply@djangoproject.com"
 FUNDRAISING_DEFAULT_FROM_EMAIL = "fundraising@djangoproject.com"
 
+MAILERS = {
+    "default": {
+        "BACKEND": "django.core.mail.backends.smtp.EmailBackend",
+    },
+}
+
+
 FIXTURE_DIRS = [PROJECT_PACKAGE / "fixtures"]
 
 INSTALLED_APPS = [
--- a/djangoproject/settings/common.py
+++ b/djangoproject/settings/common.py
@@ -57,6 +57,9 @@ FUNDRAISING_DEFAULT_FROM_EMAIL = "fundraising@djangoproject.com"
 MAILERS = {
     "default": {
         "BACKEND": "django.core.mail.backends.smtp.EmailBackend",
+        "OPTIONS": {
+            "host": "localhost",
+        },
     },
 }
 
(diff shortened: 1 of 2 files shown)


ulgens (approved):
Looks good to me.

JohananOA (approved):
Looks good

djangoproject/settings/common.py
```diff
@@ -54,6 +54,13 @@
 DEFAULT_FROM_EMAIL = "noreply@djangoproject.com"
 FUNDRAISING_DEFAULT_FROM_EMAIL = "fundraising@djangoproject.com"
 
+MAILERS = {
+    "default": {
+        "BACKEND": "django.core.mail.backends.smtp.EmailBackend",
+    },
```
adamzap, line 60: [The docs](https://docs.djangoproject.com/en/6.1/topics/email/#smtp-backend) say that the `host` option is required. [This section](https://docs.djangoproject.com/en/6.1/topics/email/#django.core.mail.backends.smtp.EmailBackend) mentions it defaulting to `localhost`, but that seems to be referring to direct instantiation of the class only.

The test suite runs with no deprecation warnings with this change, but it seems like we should specify `localhost`...?

Arbaaz123676: Yeah. I checked the Django 6.1 migration docs and confirmed that host should be explicitly set when using MAILERS. Since the project relied on Django’s default localhost, I’ll add that to preserve the existing behavior.

adamzap: Thanks!

adamzap (approved):
👍 Looks great!

PR #2792, opened 2026-08-27
Set up i18n URLs with language chooser

### LINKED ISSUE #883: Support switching languages on non-docs sites

mgax:
Fixes https://github.com/django/djangoproject.com/issues/883

Credit for most of the work goes to @marksweb; I merely brought it up to date and refactored `ExcludeHostsLocaleMiddleware` to make it simpler and easier to test.

Open questions:
- Are there any JS strings translated, and are they included in the plan (https://github.com/django/djangoproject.com/issues/1648)? I could find no mention of that.
- What is the right set of languages to add to `settings.LANGUAGES`?

---

#### ToDo
- [x] Cherry-pick the i18n URL setup and language chooser from https://github.com/django/djangoproject.com/pull/1251.
- [x] Update the language chooser markup (looks like the HTML does not match the styles any more).
- [x] See if the "downloads" route needs special handling, like in the old PR.
- [x] Make testsuite pass locally.
- [ ] Check plans for translating JS strings.
- [ ] Decide which languages to include in `settings.LANGUAGES`.
- [x] Address double language menu on docs site.

### DIFF
--- a/djangoproject/middleware.py
+++ b/djangoproject/middleware.py
@@ -22,6 +22,15 @@ class CORSMiddleware:
         return response
 
 
+def _get_host_name(request):
+    """
+    Get the name of the `django-hosts` host that was matched for this request.
+    `django-hosts` sets `request.host`; while this behaviour is not documented, it is
+    not likely to change.
+    """
+    return request.host.name
+
+
 class ExcludeHostsLocaleMiddleware(LocaleMiddleware):
     """
     Locale middleware that lets us exclude requests to certain hosts (e.g.,
@@ -40,11 +49,11 @@ class ExcludeHostsLocaleMiddleware(LocaleMiddleware):
         return host not in self._excluded_hosts
 
     def process_request(self, request):
-        if self._is_host_included(request.host.name):
+        if self._is_host_included(_get_host_name(request)):
             super().process_request(request)
 
     def process_response(self, request, response):
-        if self._is_host_included(request.host.name):
+        if self._is_host_included(_get_host_name(request)):
             return super().process_response(request, response)
         return response
 
--- a/djangoproject/middleware.py
+++ b/djangoproject/middleware.py
@@ -2,11 +2,9 @@ import re
 
 from django.conf import settings
 from django.http import HttpResponsePermanentRedirect
-from django.http.request import split_domain_port
 from django.middleware.locale import LocaleMiddleware
 from django.urls import Resolver404, resolve
 from django.utils.functional import cached_property
-from django.utils.http import is_same_domain
 
 
 class CORSMiddleware:
@@ -27,7 +25,8 @@ class CORSMiddleware:
 class ExcludeHostsLocaleMiddleware(LocaleMiddleware):
     """
     Locale middleware that lets us exclude requests to certain hosts (e.g.,
-    docs.djangoproject.com) from being processed by LocaleMiddleware.
+    `docs`) from being processed by LocaleMiddleware. Depends on the
+    `django_hosts` middleware having processed the request.
     """
 
     @cached_property
@@ -36,22 +35,16 @@ class ExcludeHostsLocaleMiddleware(LocaleMiddleware):
 
     def _is_host_included(self, host):
         """
-        Mirrors the behavior of django.http.request.validate_host(), but does
-        not match '*' (which would exclude all hosts). To exclude all requests
-        from being processed by LocaleMiddleware one should simply remove this
-        class from settings.MIDDLEWARE.
+        Check whether the host is part of the excluded hosts list.
         """
-        domain, _ = split_domain_port(host)
-        return not any(
-            is_same_domain(domain, pattern) for pattern in self._excluded_hosts
-        )
+        return host not in self._excluded_hosts
 
     def process_request(self, request):
-        if self._is_host_included(request.get_host()):
+        if self._is_host_included(request.host.name):
             super().process_request(request)
 
     def process_response(self, request, response):
-        if self._is_host_included(request.get_host()):
+        if self._is_host_included(request.host.name):
             return super().process_response(request, response)
         return response
 
--- a/djangoproject/settings/common.py
+++ b/djangoproject/settings/common.py
@@ -267,6 +267,8 @@ HOST_SITE_TIMEOUT = 3600
 
 ROOT_HOSTCONF = "djangoproject.hosts"
 
+LOCALE_MIDDLEWARE_EXCLUDED_HOSTS = ["docs"]
+
 # django-recaptcha settings
 
 # The keys themselves are only set in production, but the score threshold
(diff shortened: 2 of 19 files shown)


mgax:
I'm just about ready to wrap up here, but I'm stumped by 3 test failures. @marksweb your PR did not seem to need changes to these tests, do you have any idea why they might be failing here? Thanks!

```
FAIL: test_docs_host_excluded (djangoproject.tests.ExcludeHostsLocaleMiddlewareTests.test_docs_host_excluded)
We get no Content-Language or Vary headers when docs host is excluded
  File "/home/runner/work/djangoproject.com/djangoproject.com/djangoproject/tests.py", line 136, in test_docs_host_excluded
    self.assertEqual(resp.status_code, HTTPStatus.OK)
AssertionError: 404 != <HTTPStatus.OK: 200>
FAIL: test_docs_host_forwarded_excluded (djangoproject.tests.ExcludeHostsLocaleMiddlewareTests.test_docs_host_forwarded_excluded)
We get no Content-Language or Vary headers when docs host
  File "/home/runner/work/djangoproject.com/djangoproject.com/djangoproject/tests.py", line 161, in test_docs_host_forwarded_excluded
    self.assertEqual(resp.status_code, HTTPStatus.OK)
AssertionError: 404 != <HTTPStatus.OK: 200>
FAIL: test_docs_host_not_excluded (djangoproject.tests.ExcludeHostsLocaleMiddlewareTests.test_docs_host_not_excluded)
We still get Content-Language when docs host is not excluded
  File "/home/runner/work/djangoproject.com/djangoproject.com/djangoproject/tests.py", line 170, in test_docs_host_not_excluded
    self.assertEqual(resp.status_code, HTTPStatus.OK)
AssertionError: 302 != <HTTPStatus.OK: 200>
```

mgax:
It looks like 3640cc5 has updated `ExcludeHostsLocaleMiddlewareTests` to work around a change in the way `django_hosts` matches hosts. It landed after #1251 was written, which explains why the original tests were fine.

So what's going on? In all the failing tests, the intention was to simulare a request to the "docs" host, but `django_hosts` fails to match the "docs" host, and falls back to the default, which is the "www" host. So the request is resolved against the  urlconf of the "www" host, whose `/` route either returns 404 (when the locale middleware is disabled) or 302 (when it's enabled) instead of the expected 200 response.

FWIW, here are the regular expressions which `django_hosts` is using to route requests. For tests which specify a port, the match is successful, for the others it is not.

```pycon
>>> from pprint import pp
>>> import djangoproject.hosts
>>> pp([(h.name, h.compiled_regex.pattern) for h in djangoproject.hosts.host_patterns])
[('www', 'www\\.djangoproject.localhost:8000(\\.|:|$)'),
 ('docs', 'docs\\.djangoproject.localhost:8000(\\.|:|$)'),
 ('dashboard', 'dashboard\\.djangoproject.localhost:8000(\\.|:|$)')]
```

Zooming out a bit, the `ExcludeHostsLocaleMiddlewareTests` test is mostly exercising the host matching code of `ExcludeHostsLocaleMiddleware`. But why does `ExcludeHostsLocaleMiddleware` need to perform its own host matching, when `django_hosts` already does this, and sets `request.host` to the matched host object? `ExcludeHostsLocaleMiddleware` could simply check `request.host.name` against a list.

---
My suggestion would be to update `ExcludeHostsLocaleMiddleware` to rely on `django_hosts` host matching and keep only two tests: "host is excluded" and "host is not excluded". I'm planning to make the change in this PR, but I'm happy to spin it out into a separate PR, or take another approach altogether – let me know.

djangoproject/settings/common.py
```diff
@@ -108,7 +114,12 @@
 if os.getenv("DJANGO_SPOOKY_MODE"):
     INSTALLED_APPS.insert(0, "django_admin_dracula")  # spooky 👻
 
-LANGUAGE_CODE = "en-us"
+LANGUAGE_CODE = "en"
+
+LANGUAGES = [
+    ("en", gettext("English")),
+    ("fr", gettext("French")),
+]
```
mgax, line 122: **Question:** What is the right list of languages that we support?

djangoproject/middleware.py
mgax, line 1: `ExcludeHostsLocaleMiddleware` is essentially replicating work of `django-hosts` but differently. This caused a drift in tests and makes it difficult to update `ExcludeHostsLocaleMiddlewareTests` for this PR: https://github.com/django/djangoproject.com/pull/2792#issuecomment-5462849102. My approach is to modify `ExcludeHostsLocaleMiddleware` to pick up the host detection of `django-hosts`.

djangoproject/tests.py
```diff
@@ -125,25 +125,16 @@ class ExcludeHostsLocaleMiddlewareTests(ReleaseMixin, TestCase):
...
         """We get no Content-Language or Vary headers when docs host is excluded"""
         with self.settings(LOCALE_MIDDLEWARE_EXCLUDED_HOSTS=[self.docs_host]):
-            resp = self.client.get("/", headers={"host": self.docs_host})
+            resp = self.client.get("/", headers={"host": self.docs_host_header})
 
-        self.assertEqual(resp.status_code, HTTPStatus.OK)
-        self.assertNotIn("Content-Language", resp)
-        self.assertNotIn("Vary", resp)
-
-    def test_docs_host_with_port_excluded(self):
```
mgax, line 140: Port matching is already required for `django-hosts` to correctly match the host, so I have updated the test above to specify the port number, and dropped `test_docs_host_with_port_excluded`. Same for `test_www_host_with_port` below.

djangoproject/tests.py
```diff
@@ -156,36 +147,29 @@ def test_docs_host_forwarded_excluded(self):
         with self.settings(
             LOCALE_MIDDLEWARE_EXCLUDED_HOSTS=[self.docs_host], USE_X_FORWARDED_HOST=True
         ):
-            resp = self.client.get("/", headers={"x-forwarded-host": self.docs_host})
+            resp = self.client.get(
+                "/", headers={"x-forwarded-host": self.docs_host_header}
+            )
 
-        self.assertEqual(resp.status_code, HTTPStatus.OK)
+        self.assertEqual(resp.status_code, HTTPStatus.FOUND)
```
mgax, line 154: This change is essentially reverting https://github.com/django/djangoproject.com/commit/3640cc55b5e57179f5c740b527ab9acf50b1cb5d – the correct response for a `GET /` on the docs site is a redirect to `/en/{version}/`.

Issue #2793, opened 2026-08-27
[sponsor efforts] Add a call to action for sponsors in the site Banner


nessita:
The most recent active Banner instance is shown in every template (except a few). In order to have clearer and more prominent call to actions for sponsors (corporate member funnel), we should include a link to "Become a Sponsor" in the Banner base template, roughly in:

```django
    <div id="billboard">
      {% block billboard %}{% active_banner %}{% endblock %}
    </div>
```

This is related to improving corporate member conversion in the DjangoCon US 2026 sprint.

cc/ @jefftriplett @ericholscher

afs2015:
After chatting with @ericholscher, an initial design is adding it below the current CTA

https://github.com/django/djangoproject.com/pull/2800

Mobile:
<img width="301" height="372" alt="Image" src="https://github.com/user-attachments/assets/a976f092-9835-4bb2-b8b8-b1ac043e6691" />

Desktop:

<img width="1458" height="409" alt="Image" src="https://github.com/user-attachments/assets/f4b92327-3d61-46da-be63-c3ccf3a9e49d" />

ericholscher:
Also this one: 

https://github.com/django/djangoproject.com/pull/2627

afs2015:
@nessita: I believe this issue can be closed because #2800 addresses the original request. #2627 is related, but I would argue non-blocking.

sarahboyce:
I think I would remove the banner from this sponsor page. The sponsor page is important enough I feel we wouldn't want to distract with a banner and when you click "become a sponsor" having this linked above feels odd

nessita:
> I think I would remove the banner from this sponsor page. The sponsor page is important enough I feel we wouldn't want to distract with a banner and when you click "become a sponsor" having this linked above feels odd

Yes this is something that Eric wanted to do but until last Fri, the sponsor page was a flatpage that couldn't be customized. Jeff merged [a change that made it a template page](https://github.com/django/djangoproject.com/commit/253ae7cf0ca8717d6d8aacf29aec8f14d9d5782a) so we could now disable the banner by overriding the `billboard` block.

jefftriplett:
Done. The banner template links to the sponsor page (`djangoproject/templates/foundation/banner.html`), and the sponsor and prospectus pages override `{% block billboard %}` so the banner does not appear above their own call to action.

PR #2800, opened 2026-08-27
Adds sponsor link to banner template

afs2015:
For https://github.com/django/djangoproject.com/issues/2793 adds a link to the sponsor page in the banner template.

### DIFF
--- a/djangoproject/templates/foundation/banner.html
+++ b/djangoproject/templates/foundation/banner.html
@@ -6,8 +6,6 @@
     {% if banner.cta_url and banner.cta_label %}
       <a id="banner-cta" class="cta" href="{{ banner.cta_url }}">{{ banner.cta_label }}</a>
     {% endif %}
-    <a class="link-readmore" href="https://www.djangoproject.com/sponsor/">
-        Become a Sponsor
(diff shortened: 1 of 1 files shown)


ericholscher:
An example screenshot: 

Screenshot 2026-08-27 at 3 50 33 PM

cc @nessita

jcjudkins:
it looks great to me. However, I have a question. Is this the ONLY banner we use? If there are other special announcements this seems to be the only banner model. Just wondering if we need a structure for the banner to be changed as needed.

afs2015:
> it looks great to me. However, I have a question. Is this the ONLY banner we use? If there are other special announcements this seems to be the only banner model. Just wondering if we need a structure for the banner to be changed as needed.

From what I see this is the only banner. 

We could add two fields to the model `Banner` similar to `cta_url` and `cta_label` then wrap it in an equivalent`{% if banner.cta_url and banner.cta_label %}` like the CTA above.

I'll defer to the website working group if that level of flexibility is needed.

jefftriplett (approved):
Looks good.

jefftriplett (approved):

PR #2811, opened 2026-08-31
Added featured blog entries to the top of list views.

### LINKED ISSUE #2801: Add support for pinning blog posts
The [djangoproject.com blog](https://www.djangoproject.com/weblog/) sometimes contains time-sensitive or particularly important posts, such as calls for applications for the Executive Director or Django Fellow roles. These posts can quickly become less visible as newer posts push them down the page.

Add a way to pin a blog post so that it remains prominently displayed at the top of the blog page. The page should clearly indicate that the post is pinned.

[Optional] Display the pinned post within the “Latest News” section, potentially increasing that section from two posts to three.

sarahboyce:

I discussed with Chatgpt what approach to do with this ticket

-----

Fixes https://github.com/django/djangoproject.com/issues/2801
I currently do not have much UI changes beyond a "FEATURED" label in the blog entry meta data

Screenshot from 2026-08-31 16-55-37


blog/migrations/0008_entry_featured.py
```diff
@@ -0,0 +1,20 @@
+# Generated by Django 6.0.6 on 2026-08-31 08:30
```
adamzap: This probably won't make a difference, but could you update your local environment to have Django v6.1 generate this migration since that's what the project is running? Just to be safe!

blog/models.py
```diff
@@ -38,6 +38,9 @@ class EntryQuerySet(models.QuerySet):
     def published(self):
         return self.active().filter(pub_date__lte=timezone.now())
 
+    def priority_order(self):
```
adamzap, line 41: Should we also add a test for this in `EntryTestCase` since `published` and `active` are tested there?

blog/tests.py
```diff
@@ -468,6 +469,54 @@ def test_corporate_sponsors_displayed(self):
...
+        response = self.client.get(reverse("weblog:index"))
+
+        self.assertEqual(response.status_code, 200)
+        self.assertQuerySetEqual(
+            response.context["object_list"],
+            [featured_entry, newer_entry],
+        )
+        self.assertContains(response, "Featured", count=1)
+
+    def test_render_latest_blog_entries_prioritizes_featured_entries(self):
```
adamzap: I'd be ok with either dropping this test or adding it to a new test class for two reasons:

1. It's not calling the view code
2. The other test you added covers this case at the view level

What do you think?

djangoproject/templates/blog/news_summary.html
```diff
@@ -8,6 +8,10 @@
   {{ object.summary_html|safe }}
 {% endif %}
 <span class="meta">
+  {% if object.featured %}
+    <span style="text-transform: uppercase;">{% translate "Featured" %}</span>
```
adamzap: This project only has a few inline styles. Would there be a way to add this to the CSS file?

adamzap (commented):
A few thoughts, nothing major.

blog/tests.py
sarahboyce: I have added a new test class `TemplateTagTests`. I added this test to cover the changes in the template tag that isn't covered by the views being tested here (as it is used in the homepage)

adamzap (approved):
👍 Looks great!

PR #2818, opened 2026-09-02
[checklists] Emitted one CVE credits entry per person.

### LINKED ISSUE #2749: [checklists] CVE record generator: support multiple credits entries per role
The generator emits one credits entry per role, packing multiple people into a single value string. `credits` is an array in CVE 5.1, so each person should get their own entry with the appropriate type, allowing downstream consumers to attribute individually. Needs a way to express a list of names per role in the input metadata.

nessita:
The credits array in CVE 5.1 expects one entry per person, but the generator packed every name for a role into a single value string, leaving downstream consumers unable to attribute individually.

Reporter and remediator now hold a comma separated list of names, and each name gets its own credits entry with the appropriate type. Added the optional reviewer and analyst fields, credited as "remediation reviewer" and "analyst", so those roles can be attributed too.

Prose usages (blog post, checklist and reporter verification email) go through reporters_display, which enumerates the names with an Oxford comma. A blank role no longer emits an empty credits entry.

Also reused the new split_comma_separated() helper for the CWE and CAPEC type fields, which parsed their comma separated values inline.

Fixes #2749.

Claude Code was used to streamline the edits, under my step-by-step steering.


jacobtylerwalls (approved):
Looks good!

PR #2820, opened 2026-09-03
Replaced SCSS syntax in situ with vanilla CSS syntax

### LINKED ISSUE #2251: Moving away from SASS
While working on https://github.com/django/djangoproject.com/pull/2212, I encountered an issue: `libsass` package was requiring build dependencies, even though [its readme states that ](https://github.com/sass/libsass-python/blob/main/README.rst#features)

> Provides prebuilt [wheel](https://www.python.org/dev/peps/pep-0427/) binaries for Linux, Windows, and Mac.

A further investigation revealed that this is an issue with M-series processors: https://github.com/django/djangoproject.com/pull/2248#issuecomment-3389897033

While looking further, I found out that [libsass itself is deprecated](https://github.com/sass/libsass):

> Warning: [LibSass is deprecated](https://sass-lang.com/blog/libsass-is-deprecated). While it will continue to receive maintenance releases indefinitely, there are no plans to add additional features or compatibility with any new CSS or Sass features. Projects that still use it should move onto [Dart Sass](https://sass-lang.com/dart-sass).

At this point, it was clear that at some point we will need to move away from using libsass. During the related discussions, @bmispelon stated that moving away from SASS is also an option:

> ...maybe abandon sass altogether and use modern plain css (I wonder what feature of sass we actually need that wouldn't be standard css nowadays)

So far, me, @pauloxnet and @adamzap are supportive of this idea.

I'm creating this issue so we can discuss anything related to the idea of moving away from SASS and the related implementation.

smallsaucepan:
Replaced SCSS syntax in situ with vanilla CSS syntax. This is phase 1 of the plan to move away from libsass for generaing styles.

There are a lot of changes especially in styles.css. Most though are fairly self contained e.g. including a particular mixin inline. By and large SCSS type nesting is natively supported in the majority of browsers (since late 2023) so that works in our favour.

Leaving the CSS syntax in the SCSS files allows for easier merging of existing PRs. Moving the files to their final location (in static/css) can happen at a later date after we have possibly merged a couple of existing PRs.

Switched to using a CDN equivalent FontAwesome setup e.g. still self hosted but not customised in any way.

Took the removal of the font-size mixin as a cue to start specifying font sizes in rem units. Believe this is considered best practice.

Claude to help get my head around how the different CSS color spaces work.

Will partially fix #2251


sarahboyce:
Hi @smallsaucepan :wave: 
Thank you for this PR! We discussed this in the website WG meeting 
We like the idea of this change but will consider this blocked until https://github.com/django/djangoproject.com/issues/2822 is complete in order to be better prepared to test it :+1:

smallsaucepan:
Sounds good 🦺

PR #2502, opened 2026-02-09
i18n: add missing header, menu, subfooter, footer and search strings …

kehach07:
Adds missing translation markers for header, menu, subfooter, footer and search form.

Updates the English message catalog to expose the strings to Transifex.

Refs #2500

PR #2821, opened 2026-09-03
Translate the "Skip to main content" link

### LINKED ISSUE #2649: "Skip to main content" link is not translated
The visually-hidden "Skip to main content" link at the top of the body tag is not translated, so appears in English for all languages on the docs site. (Is there some reason accessibility features shouldn't be localized?)

That link is added in [base.html](https://github.com/django/djangoproject.com/blob/5475f6c0f1baecbfd8e950b2b71b2780c6149dab/djangoproject/templates/base.html#L51)

kunalKumar-13:
The visually hidden skip link in base.html was the only piece of chrome rendered before the header that was not marked for translation, so it appeared in English on every localised page.

base.html did not load the i18n tag library, so add it alongside the existing loads and wrap the string in {% translate %}.

The catalogue entry is deliberately left out: #2502 is open and regenerates locale/en/LC_MESSAGES/django.po wholesale, so editing it here would conflict. makemessages picks the string up on the next run.

Fixes #2649

I used Claude Code as an assistant on this one. The two-line diff is small enough
to check by hand, and I did: `base.html` genuinely wasn't loading the i18n tag
library, which is why this string had never been picked up. I rendered the
template with the tag applied and ran `makemessages` in a scratch project to
confirm it extracts as `msgid "Skip to main content"`.


SaptakS (approved):
Thanks for the PR! Makes sense to me.

Issue #2288, opened 2025-10-24
Site usage analytics


thibaudcolas:
We have been discussing setting up analytics on the website for a while (forum: [Add Plausible Tracking to DjangoProject.com?](https://forum.djangoproject.com/t/add-plausible-tracking-to-djangoproject-com/35581)). At this point I think there is a pretty clear case for it and it’s more of a matter of deciding what tool(s) we could use and setting them up.

So I think opening this as a GitHub issue will help make this more of a task that we want to happen, not just an idea.

## Tasks

- [x] Discuss analytics use cases and potential problems
- [x] List possible approaches
- [x] ⌛️ Try out 2-3 promising approaches/ tools
- [x] Check with ops team on self-hosting
- [ ] Pick top 1 and top 2 options
- [ ] Set up the most relevant options on the site
- [ ] Access for website WG, fundraising WG, Board, and any other relevant stakeholders

## Tools considered so far

For the dedicated analytics / tracking tools, likely options:

- [Plausible CE](https://github.com/plausible/community-edition/) (used by the PSF for their docs)
- [Plausible SaaS](https://plausible.io/)
- [Cabin](https://withcabin.com/) (has a very generous free tier)
- [Umami SaaS](https://umami.is/pricing)
- [Umami self-hosted](https://github.com/umami-software/umami)
- ~[Google Analytics](https://developers.google.com/analytics) (industry standard but requires cookies)~

## Requirements

We want something that does not require cookies, is compatible with privacy laws (GDPR, CCPA), ideally has minimal to no impact on performance for site users, ideally allows us to get good reports with little efforts. I think there are three approaches that are viable:

1. Use the existing Django server logs (behind Fastly caching) as an incorrect but likely interesting data source.  We were hoping to do this for #1628 but this never happened.
2. Reach out to Fastly to see if we could get access to their analytics product, and use that. 
3. Set up a dedicated JS-based analytics beacon

I have no experience with Fastly analytics / logs tools, but from my experience with Cloudflare’s equivalent this is pretty viable to understand popular pages and high-level geographic audience details. That’s about it. Big benefit is it doesn’t require any additional, potentially invasive tracking.

Those come with the clear drawback of loading more code for users, and tracking more data than we would often want. But I think there are ways to mitigate that - picking the least-intrusive option, or only having tracking turned on occasionally.

## Use cases

Just reiterating the use cases we have - here is what we want to know:

- Rough geographic distribution of our audience across countries
- Page views / Session counts site-wide
[...]

PR #2823, opened 2026-09-04
Added Plausible analytics tracking.

sarahboyce:
Relates to https://github.com/django/djangoproject.com/issues/2288

### DIFF
--- a/djangoproject/templates/registration/base.html
+++ b/djangoproject/templates/registration/base.html
@@ -2,4 +2,6 @@
 {% load i18n %}
 {% block sectionid %}community{% endblock %}
 
+{% block analytics %}{% endblock %}
+
 {% block header %}<p>{% translate "Accounts" %}</p>{% endblock %}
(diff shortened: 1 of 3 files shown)


djangoproject/templates/registration/base.html
```diff
@@ -2,4 +2,6 @@
 {% load i18n %}
 {% block sectionid %}community{% endblock %}
 
+{% block analytics %}{% endblock %}
```
sarahboyce, line 5: The main thing to review is whether there are pages you think we should exclude from tracking
The dashboard is currently not public but perhaps there are pages we don't want to track

JohananOppongAmoateng: Hmm of the top of my head i personally cant think of any page right now. I will think  about it and come back

sabderemane: I agree with Johanan, I was thinking of admin login page but the number might be very low compared to other pages in tracking. Not sure it's really worth it.

sabderemane (approved):
Looks good to me, thanks! ✨

PR #2826, opened 2026-09-04
Fixed #1954 -- Corrected scroll-to-text fragments in docs search results

### LINKED ISSUE #1954: When searching execute, link to scroll to text fragment doesn't work due to incorrect fragment
Due to [this PR](https://github.com/django/djangoproject.com/pull/1934) the docs search now has scroll to text.

When searching execute: https://docs.djangoproject.com/en/5.1/search/?q=execute

The second option, "Performing raw SQL queries", links to 
https://docs.djangoproject.com/en/5.1/topics/db/sql/#:~:text=execute%28%22UPDATE%20bar%20SET%20foo%20%3D%201%20WHERE%20baz%20%3D%20%25s%20%22%2C%5Bself.%20baz%5D%29
Decoded: `https://docs.djangoproject.com/en/5.1/topics/db/sql/#:~:text=execute("UPDATE bar SET foo = 1 WHERE baz = %s ",[self. baz])`
However this doesn't scroll to the place.
The correct one should be this one
https://docs.djangoproject.com/en/5.1/topics/db/sql/#:~:text=execute(%22UPDATE%20bar%20SET%20foo%20%3D%201%20WHERE%20baz%20%3D%20%25s%22,%20[self.baz])
Decoded: `https://docs.djangoproject.com/en/5.1/topics/db/sql/#:~:text=execute("UPDATE bar SET foo = 1 WHERE baz = %s", [self.baz])`

There are two differences:
- Extra space after: `WHERE baz = %s`
- Extra space after the dot: `[self. baz])`

I haven't dug further on why is this happening, whether is a fragment issue only or is an indexing issue that just happens to show up on this feature

kunalKumar-13:
Fixes #1954.

Search result links carry a scroll-to-text fragment built from the PostgreSQL
search headline. Because Pygments splits code into separate tokens, the text the
fragment is built from contains spacing the rendered page does not, so the
browser finds no match and drops the reader at the top of the page instead of at
the result.

Reproduced first against production. `?q=execute` currently emits:

```
#:~:text=execute("UPDATE bar SET foo = 1 WHERE baz = %s ",[self. baz])
```

for a page that renders `execute("UPDATE bar SET foo = 1 WHERE baz = %s", [self.baz])`.

Four things are handled now:

* The headline is HTML, so it is unescaped in full rather than only translating
  `&quot;`. A fragment containing a literal `&#39;` can never match page text.
* Spacing just inside a pair of double quotes is dropped — Pygments splits a
  string literal whenever it contains a placeholder, turning
  `"WHERE baz = %s"` into `"WHERE baz = %s "`.
* A dot with spaces on both sides is attribute access and is joined up, so
  `self . baz` becomes `self.baz`. A dot followed by a capitalised word ends a
  sentence in prose and is left alone — without that guard,
  `reach gettext . This means` collapses to `gettext.This means`.
* An opening bracket only binds to the preceding token when that token is an
  identifier or a closing bracket, i.e. a call or a subscript. Previously
  `, [self.baz]` lost the space after the comma and `urlpatterns = [` became
  `urlpatterns =[`.

### Measurement

I harvested 280 search results across 44 queries from docs.djangoproject.com and
checked that the current code reproduces all 280 production fragments exactly,
so the set is a faithful before-image. Against it:

| | before | after |
|---|---|---|
| fragments unchanged | — | 268 |
| fragments changed | — | 12 |
| changed fragments found in the rendered target page | 0 / 12 | 8 / 12 |

The four still not found are the limitation already noted in the docstring
(`votes = 0` is not narrowed to `votes=0`), which this PR does not attempt.

`docs` is 77 tests, passing before and after. Reverting only
`docs/templatetags/docs.py` while keeping the new tests fails exactly the four
new behaviour cases.

Happy to split this into four commits if that reads better.

I worked on this with Claude Code, and reviewed every line before opening it.

The evidence above isn't model output. I fetched the corpus from
docs.djangoproject.com myself, confirmed the unmodified code reproduces all 280
production fragments as a before-image, then fetched each of the 12 changed
target pages to check the fragments against the real rendered text. The corpus
also caught a regression in an earlier attempt of mine, where the dot rule
mangled ordinary prose — that's now guarded by a test.

### DIFF
--- a/docs/templatetags/docs.py
+++ b/docs/templatetags/docs.py
@@ -1,4 +1,5 @@
 import re
+from html import unescape
 from urllib.parse import quote
 
 from django import template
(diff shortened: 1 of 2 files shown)


sarahboyce (approved):
Thank you for the PR
I tested and this works well ⭐

PR #2827, opened 2026-09-04
Fixed #2654 -- Corrected the Benevity link in the footer

### LINKED ISSUE #2654: Broken "Benevity" footer link
The site footer link "Benevity Workplace Giving Program" is effectively broken. ([includes/footer.html](https://github.com/django/djangoproject.com/blob/5475f6c0f1baecbfd8e950b2b71b2780c6149dab/djangoproject/templates/includes/footer.html#L65))

It goes to a `#benevity-giving` anchor on the fundraising page. That anchor was removed in 7ef7b74 in 2015, so you're dropped at the top of the fundraising page instead. On that page, there *is* a list item for Benevity, which links to that same nonexistent anchor ([fundraising/index.html](https://github.com/django/djangoproject.com/blob/5475f6c0f1baecbfd8e950b2b71b2780c6149dab/djangoproject/templates/fundraising/index.html#L48-L56)):

```django
    <li>
      {% url 'fundraising:index' as fundraising_url %}
      {% blocktranslate trimmed %}
        <a href="{{ fundraising_url }}#benevity-giving">Benevity Workplace Giving
          Program</a> - If your employer participates, you can make donations to the
        DSF via payroll deduction.
      {% endblocktranslate %}
    </li>
```

**If the DSF is still receiving contributions through Benevity** then that list item should probably be the destination  of the footer link, and the self-link should be removed. (Or the fundraising page could link to benevity.com, but I couldn't find any good resources for *employees* on their site. It's all aimed at the employers who are Benevity's customers.)

```django
    {# proposed #}
    <li id="benevity-giving">
      {% blocktranslate trimmed %}
        Benevity Workplace Giving Program - If your employer participates,
        you can make donations to the DSF via payroll deduction.
      {% endblocktranslate %}
    </li>
```

**If Benevity is no longer funneling donations toward the DSF** then I'd suggest instead removing the footer link and that list item. Or maybe replacing it with a generic mention of automatic charitable payroll deductions and employer matching programs.

kunalKumar-13:
Fixes #2654.

The footer's "Benevity Workplace Giving Program" link goes to a
`#benevity-giving` anchor on the fundraising page. That anchor was removed in
7ef7b74 in 2015 and the content settled on the donate page, so today the link
drops the reader at the top of the fundraising page with no sign of where the
Benevity information went.

Following @medmunds' suggestion and @SaptakS' agreement in the issue, this points
at `/foundation/donate/#benevity-giving`. I checked that page does serve an
element with `id="benevity-giving"`.

The matching list item on the fundraising page was already corrected in 9fedb3b
(#2628); this is the same target. It goes through `{% url 'homepage' %}`, matching
how the footer already links to `/foundation/`, so the link still resolves when
the footer is rendered on the docs, code and dashboard hosts.

Added a regression test. `djangoproject` goes from 30 tests to 31; the three
`LocaleSmokeTests` failures are pre-existing on a clean checkout (they need
`compilemessages`) and are unchanged. Reverting only the template makes the new
test fail.

Prepared with Claude Code's help; I reviewed the result and checked the claims
rather than taking them on trust. I fetched
https://www.djangoproject.com/foundation/donate/ and confirmed the
`id="benevity-giving"` element is really there, and rendered the footer template
locally to see the corrected link resolve.

### DIFF
--- a/djangoproject/templates/includes/footer.html
+++ b/djangoproject/templates/includes/footer.html
@@ -62,7 +62,7 @@
             <li><a href="{% url "fundraising:index" %}">Sponsor Django</a></li>
             <li><a href="{% url 'members:corporate-members' %}">Corporate membership</a></li>
             <li><a href="https://django.threadless.com/" target="_blank">Official merchandise store</a></li>
-            <li><a href="{% url 'fundraising:index' %}#benevity-giving">Benevity Workplace Giving Program</a></li>
+            <li><a href="{% url 'homepage' %}foundation/donate/#benevity-giving">Benevity Workplace Giving Program</a></li>
           </ul>
         </div>
       </div>


djangoproject/tests.py
```diff
@@ -237,6 +237,23 @@ def test_single_h1_per_page(self):
                     self.assertContains(response, "<h1", count=1)
 
 
+class FooterTests(ReleaseMixin, TestCase):
```
sarahboyce: I would remove this test. The url is from a flatpage and it can break without this test failing. I don't think it adds value

kunalKumar-13:
Fair — removed in b8c59ece.

You're right that it doesn't earn its place: the anchor lives on a flatpage, so the target can disappear without the test noticing. It only ever asserted that the template renders the string I just put in it, which is close to tautological.

The PR is now the one-line template fix plus the changelog entry.

sarahboyce (approved):

PR #2828, opened 2026-09-04
Fixed #2166 -- Removed the dev-version warning from the contributing tutorial

### LINKED ISSUE #2166: Dev docs warning on contribution tutorial
While going through a beginner contributors workshop using our [Writing your first contribution for Django](https://docs.djangoproject.com/en/dev/intro/contributing/) tutorial, I noted the page shows our "This document is for Django's development version" warning. I’m not sure if this is intentional?

## Current status

For other "internals" docs, the site theme skips that warning for the `dev` version. Here:

https://github.com/django/djangoproject.com/blob/6fcf41e44aec3ec9d62fef126a8390b94a88224a/docs/templates/docs/doc.html#L39-L43

And a bit further up, marks the `dev` version as canonical:

https://github.com/django/djangoproject.com/blob/6fcf41e44aec3ec9d62fef126a8390b94a88224a/docs/templates/docs/doc.html#L11-L13

## Possible changes

Essentially I’m not sure if `dev` should be considered the canonical version for the contributors tutorial too? I believe it makes sense, as people going through this tutorial will move on to looking at a lot of other pages in the internals docs. In which case we seem to decide it’s better for them to look at the `dev` version.

kunalKumar-13:
Fixes #2166.

`intro/contributing` is the *Writing your first contribution for Django*
tutorial. Its readers are working on Django itself and move straight on into
`internals/`, so the development docs are the version they want. It nonetheless
shows the "this document is for Django's development version, which can be
significantly different from previous releases" warning, and its `rel=canonical`
points at the latest release, because both conditions in `doc.html` tested for
`"internals" in docurl`.

Checked against production before changing anything:

| URL | dev warning | canonical |
|---|---|---|
| `/en/dev/intro/contributing/` | present | `/en/6.1/intro/contributing/` |
| `/en/dev/internals/contributing/` | absent | `/en/dev/internals/contributing/` |

Both conditions now go through an `is_dev_canonical` filter, so the set of
documents the dev docs are canonical for is written down once instead of being
spelled out separately in each branch of the template.

The filter uses a prefix test where the old code used a substring test. I checked
all 912 English document URLs in `sitemap-en.xml`: every URL containing
`internals` also starts with it, so this leaves `internals/` exactly as it was
and adds only `intro/contributing`.

`docs` goes from 77 tests to 81, all passing. Narrowing the tuple back to
`("internals",)` fails exactly the four `intro/contributing` cases and nothing
else, so `internals/` and ordinary pages are demonstrably unaffected.

One thing worth a maintainer's eye: the confirmation @thibaudcolas reported from
Discord was about the warning. This also switches `rel=canonical` to `dev` for
the tutorial, which is the second of the "two lines of custom logic" mentioned in
the issue and follows the same reasoning, but it has SEO consequences. Happy to
drop that half and keep only the warning if you would rather decide it
separately — it is a one-line change.

I used Claude Code while writing the filter, the template edit and the tests, and
read all of it before opening.

The supporting numbers are measured, not asserted: the production behaviour in
the table came from fetching those three URLs, and the "912 URLs, zero
exceptions" figure came from running the filter over the real sitemap rather
than a sample.

### DIFF
--- a/docs/templates/docs/doc.html
+++ b/docs/templates/docs/doc.html
@@ -9,7 +9,7 @@
 
 {% block link_rel_tags %}
   {% if docurl %}
-    {% if "internals" in docurl and lang == "en" %} {# The canonical version of "internals" is the dev docs (but only for English, since the dev docs are never translated). #}
+    {% if docurl|is_dev_canonical and lang == "en" %} {# The canonical version of the contributing docs is the dev docs (but only for English, since the dev docs are never translated). #}
       {% url 'document-detail' lang=lang version='dev' url=docurl host 'docs' as canonical_url %}
     {% else %}
       {% url 'document-detail' lang=lang version=canonical_version url=docurl host 'docs' as canonical_url %}
@@ -37,7 +37,7 @@
 
 {% block before_header %}
   {% if release.is_dev %}
-    {% if "internals" not in docurl %}{# The dev version is canonical for internals/. #}
+    {% if not docurl|is_dev_canonical %}{# The dev version is canonical for the contributing docs. #}
       <div id="dev-warning" class="doc-floating-warning">
         {% translate "This document is for Django's development version, which can be significantly different from previous releases. For older releases, use the version selector floating in the bottom right corner of this page." %}
       </div>
--- a/docs/templates/docs/doc.html
+++ b/docs/templates/docs/doc.html
@@ -9,7 +9,7 @@
 
 {% block link_rel_tags %}
   {% if docurl %}
-    {% if docurl|is_dev_canonical and lang == "en" %} {# The canonical version of the contributing docs is the dev docs (but only for English, since the dev docs are never translated). #}
+    {% if docurl|is_dev_canonical and lang == "en" %} {# The canonical version of internals/ and the contributing tutorial is the dev docs (but only for English, since the dev docs are never translated). #}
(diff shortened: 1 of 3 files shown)


kunalKumar-13:
Both of the review points above were fixed in ba6d3c91, which landed after the review ran — noting it here so nobody re-reads the stale diff.

**Path-segment boundary.** `is_dev_canonical()` no longer uses a bare `startswith` on the tuple:

```python
return any(
    docurl == prefix or docurl.startswith(f"{prefix}/")
    for prefix in DEV_CANONICAL_DOCS
)
```

The three inputs the review named returned `True` under the old form and return `False` now:

| docurl | before | now |
| --- | --- | --- |
| `internals-of-something` | True | False |
| `internalsomething` | True | False |
| `intro/contributing-extra` | True | False |

All three are in `test_is_dev_canonical`, along with the exact-match and child-page cases.

**Template comments.** Both now name the actual scope rather than just the tutorial — "The canonical version of internals/ and the contributing tutorial is the dev docs" on line 12, and the matching note on line 40.

Worth adding that the pre-existing template condition was `{% if "internals" in docurl %}`, an unanchored substring test, so a page such as `ref/internals-notes` would have matched anywhere in the path. The filter is stricter than what it replaces on both counts.

PR #2829, opened 2026-09-04
Fixed #2613 -- Allowed the end-to-end tests to run on several browsers

### LINKED ISSUE #2613: Extend playwright E2E tests to run on multiple browsers (chromium and firefox)
This follows the adding of playwright testing here: https://github.com/django/djangoproject.com/pull/2444
The current design does not allow multiple browsers to be tested. This suggested improvement would allow users to test with different browsers adding an option such as `--browser chromium --browser firefox` where only running on the chromium browser would be the default when no browser option is provided

kunalKumar-13:
Fixes #2613.

Adds a `BrowserTestRunner` with a repeatable `--browser` option, the shape
suggested on the issue:

```
python -m manage test --browser chromium --browser firefox
```

Chromium alone remains the default, so an ordinary run costs what it did before
and the CI workflow needs no change — it installs only chromium. Repeated values
are de-duplicated.

`setUpClass` launches every selected engine, and each test iterates them inside
`self.subTest(browser=...)` so a failure is attributed to its browser rather than
masking the rest.

Verified locally against PostgreSQL 16 with both browsers installed:

```
manage test djangoproject.tests.EndToEndTests   ->  6 tests, OK  (chromium, as before)
manage test djangoproject                       ->  29 tests, 3 failures
```

Those 3 are `LocaleSmokeTests` and they fail identically on an unmodified
checkout — they want `compilemessages`, which `make ci` runs. Confirmed by
stashing the change and re-running, so this is neutral for them.

### One thing worth knowing before you merge

Running `--browser firefox` reports two failures in
`test_cycle_theme_when_prefers_dark`. I believe that is the feature working
rather than a problem with it, but you should decide.

On the live-server page Firefox reports
`matchMedia('(prefers-color-scheme: dark)').matches === false` even though
Playwright was told `color_scheme="dark"` — with both `page.emulate_media()` and
`browser.new_context(color_scheme=...)`. Instrumenting it shows the value is
`true` immediately after emulation and `false` after navigating to the live
server; Chromium stays `true`. The theme script reads that query once at parse
time, so the cycle order inverts and the assertions fail.

I could not pin the cause. Isolated, Firefox reports `true` for all of: `data:`
URLs, a plain `http://localhost` page, a custom user agent, and pages declaring
`color-scheme` in CSS or `<meta>`. So it is something about the live-server page
specifically, and I would rather flag that honestly than guess.

Happy to mark that test chromium-only, or to keep digging, whichever you prefer.

Claude Code helped me write the test runner and the changes to `EndToEndTests`.

I checked the parts that matter before opening: the default run is unchanged at
6 tests on chromium alone, the whole `djangoproject` app is green, and the two
firefox failures described above were reproduced and instrumented rather than
inferred — which is also why I've said plainly that I could not pin down their
root cause.


djangoproject/tests.py
```diff
@@ -277,91 +278,121 @@ def setUpClass(cls):
...
-                expect(search_bar).to_have_attribute("placeholder", "Search (Ctrl+K)")
-                is_focused = page.evaluate("document.activeElement.id === 'id_q'")
-                self.assertFalse(is_focused)
-
-                page.keyboard.press("Control+KeyK")
-                is_focused = page.evaluate("document.activeElement.id === 'id_q'")
-                self.assertTrue(is_focused)
-                page.close()
+        for name, browser in self.browsers.items():
+            with self.subTest(browser=name):
```
sarahboyce: I don't like that every test now needs:
```python
        for name, browser in self.browsers.items():
            with self.subTest(browser=name):
```
Can we abstract this out?

kunalKumar-13:
Good call — that boilerplate was in all six tests. Abstracted into a `for_each_browser` decorator in 3724b70:

```python
def for_each_browser(test):
    @wraps(test)
    def wrapper(self):
        for name, browser in self.browsers.items():
            with self.subTest(browser=name):
                test(self, browser)
    return wrapper
```

so each test is now just:

```python
@for_each_browser
def test_search_placeholder_mac_mode(self, browser):
    page = browser.new_page(user_agent=self.mac_user_agent)
    ...
```

Two lines and a level of indentation gone from every test, and the `subTest` still wraps each browser individually so a failure names the browser and doesn't stop the others:

```
FAIL: test_cycle_theme_when_prefers_dark (... browser='firefox')
```

`djangoproject` and `docs`: 107 tests, OK. Default run is unchanged at 6 tests on chromium alone.

sarahboyce:
This is better. Decorating each test with `@for_each_browser` is still not ideal as every test within the E2E tests will need this. You can look at `django/django`s approach for inspiration if you like
We also need the GitHub actions to be updated

kunalKumar-13:
Thanks — that pointer was the right one. Reworked in 688c5ab to follow `django.test.selenium.SeleniumTestCaseBase`: a metaclass makes the test class browser-specific and adds a subclass per extra browser to the module namespace, so **the tests need no decoration at all** and future E2E tests get it for free.

```python
class BrowserTestCaseBase(type(StaticLiveServerTestCase)):
    browser = None

    def __new__(cls, name, bases, attrs):
        test_class = super().__new__(cls, name, bases, attrs)
        if test_class.browser or not any(
            n.startswith("test") and callable(v) for n, v in attrs.items()
        ):
            return test_class
        test_class.browser = selected_browsers[0]
        module = sys.modules[test_class.__module__]
        for browser in selected_browsers[1:]:
            subclass = super().__new__(cls, f"{browser.capitalize()}{name}", (test_class,), {...})
            setattr(module, subclass.__name__, subclass)
        return test_class
```

A failure is now attributed by class rather than by subTest label, which reads better:

```
test_cycle_theme_when_prefers_dark (djangoproject.tests.FirefoxEndToEndTests...) ... FAIL
```

One deviation worth flagging: Django uses `browser` for the *name*, and I've kept that, so the launched Playwright object is `cls.browser_instance`. Happy to rename either way.

**GitHub Actions**: the workflow now installs `chromium firefox` rather than chromium alone. I've deliberately left the default run at chromium so CI stays green — enabling firefox there would surface the two real `test_cycle_theme_when_prefers_dark` failures described in the PR body, which I could not get to the bottom of. Say the word if you'd rather CI ran both and we treated those as a blocker.

`djangoproject` and `docs`: 107 tests, OK.

sarahboyce:
> Say the word if you'd rather CI ran both and we treated those as a blocker.

I think we should treat these as a blocker. We need to investigate if we found an issue with the test or if there is an issue on firefox

kunalKumar-13:
Investigated. It is a Firefox bug, not a problem with the test or with the site, and it has a specific trigger: **`Cross-Origin-Opener-Policy`**.

When a response carries `Cross-Origin-Opener-Policy: same-origin`, Firefox silently drops Playwright's colour-scheme emulation. `SecurityMiddleware` sends that header by default (`SECURE_CROSS_ORIGIN_OPENER_POLICY = "same-origin"`), so every page the live server serves has it.

The bisect that isolates it — same HTML, two delivery paths:

```
live server page (over HTTP)      -> prefers-color-scheme: dark = False
identical HTML via set_content    -> prefers-color-scheme: dark = True
```

Same markup both times, so it is not the page. Stripping only that one response header via `context.route()` flips it back:

```
firefox  intercept, drop nothing              -> False   <- control
firefox  intercept, drop Cross-Origin-Opener-Policy -> True
chromium intercept, drop nothing              -> True
```

The first row is the control: intercepting and re-fulfilling without removing anything still fails, so the interception is not what changes the result.

Reproducible with Django out of the picture entirely — a stdlib `http.server` returning `<html><body>hi</body></html>` and one header:

```
firefox  COOP=False via context       -> dark=True
firefox  COOP=False via emulate_media -> dark=True
firefox  COOP=True  via context       -> dark=False
firefox  COOP=True  via emulate_media -> dark=False
chromium COOP=True  (either way)      -> dark=True
```

This is upstream [microsoft/playwright#33866](https://github.com/microsoft/playwright/issues/33866), "Firefox not in dark mode with `Cross-Origin-Opener-Policy` header set". One caveat worth flagging: it is marked closed, but it was closed in September 2025 for "limited upvotes and recent activity", not because it was fixed. It still reproduces on Playwright 1.62 with Firefox 153, which is what the numbers above are from.

The workaround is the one the upstream issue suggests — re-apply the emulation after navigating:

```python
page.emulate_media(color_scheme=color_scheme)
page.goto(self.live_server_url)
page.emulate_media(color_scheme=color_scheme)  # re-assert; Firefox dropped it
```

I have pushed that as a small `open_page_with_color_scheme()` helper used by both theme tests, with the reason and the upstream link in its docstring so it is obvious why the second call is not redundant.

`--browser chromium --browser firefox` is now **35 tests, all passing**. Removing just the re-apply line fails exactly the two subtests you saw and nothing else, so the fix is causal rather than incidental.

I did think about whether papering over it is the right call. The alternative is skipping the test on Firefox, but the site's behaviour here is genuinely correct — Firefox honours `prefers-color-scheme` fine for real users, and this only affects the automation channel. Happy to switch to a skip with an xfail-style marker if you would rather the workaround not live in the test.

##### AI Assistance Disclosure

Claude Code was used for this investigation. The bisect, the header isolation, the control run and the minimal reproduction were all executed and their output is quoted verbatim above; I confirmed the upstream issue's closure reason directly on the GitHub API before describing it.

djangoproject/test_runner.py
```diff
@@ -0,0 +1,44 @@
+from django.test.runner import DiscoverRunner
+
+#: Playwright browser engines the end-to-end tests can run against.
```
sarahboyce:
```suggestion
# Playwright browser engines the end-to-end tests can run against.
```
Please remove the colons in the comments

djangoproject/tests.py
```diff
@@ -271,30 +273,64 @@ def test_collectstatic_with_manifeststaticfilesstorage(self):
...
 
 
-class EndToEndTests(ReleaseMixin, StaticLiveServerTestCase):
+class BrowserTestCaseBase(type(StaticLiveServerTestCase)):
+    """Create one test class per selected browser.
+
+    Modelled on ``django.test.selenium.SeleniumTestCaseBase``: rather than every
+    test looping over browsers, the metaclass makes the class browser-specific
+    and adds a subclass per extra browser to the module namespace. Tests keep
+    using ``self.browser`` and need no decoration.
```
sarahboyce: I would perhaps remove this. Note that django.test.selenium.SeleniumTestCaseBase no longer exists on main

djangoproject/test_runner.py
```diff
@@ -0,0 +1,44 @@
...
+
+class BrowserTestRunner(DiscoverRunner):
+    """Test runner that lets the end-to-end tests target several browsers.
+
+    ``--browser`` may be repeated::
+
+        python -m manage test --browser chromium --browser firefox
+
+    Chromium alone is the default, so a plain test run costs no more than it
+    did before and CI needs no extra browsers installed.
```
sarahboyce: Should we have that if no browser is specified, the e2e tests are skipped? I feel it is better to be explicit here rather than have an implicit default?

.github/workflows/tests.yml
```diff
@@ -41,7 +41,7 @@ jobs:
                   python -m pip install -r requirements/tests.txt
             - name: Install playwright browsers
               run: |
-                  python -m playwright install --with-deps chromium
+                  python -m playwright install --with-deps chromium firefox
```
sarahboyce, line 44: Please also update the GitHub action to run against firefox

sarahboyce (commented):
Thank you! We're getting closer ⭐

kunalKumar-13:
Thanks — all four addressed.

**Colons in comments** — removed throughout `test_runner.py`.

**The `SeleniumTestCaseBase` reference** — gone. Good catch that it no longer exists on main; the docstring now just describes what the metaclass does.

**Skipping when no browser is given** — agreed, and done. `selected_browsers` now starts empty and the metaclass marks the class `__unittest_skip__` with "No browser selected; pass --browser to run the end-to-end tests." So:

```
python -m manage test djangoproject.tests
    Ran 29 tests — OK (skipped=6)

python -m manage test djangoproject.tests --browser chromium --browser firefox
    Ran 35 tests — OK
```

Two things I like better about this than the old default: a run now states which engines it covered instead of implying chromium, and a plain test run needs no browsers installed at all.

**The GitHub action** — it now names both engines. `make ci` takes a `TEST_ARGS` passthrough and the workflow calls `make ci TEST_ARGS="--browser chromium --browser firefox"`, so the browsers are visible in the workflow file rather than buried in a default. Playwright already installs `chromium firefox` in the step above.

`prek` clean across all four files.

kunalKumar-13:
@sarahboyce a small process question while you are here, if you have a moment.

Four of my other PRs — #2821, #2828, #2830, #2831 — have **no CI checks at all**. Not failing: never run. #2829 is the only one with checks, and I think that is because you reviewed it, which authorised the workflows.

That leaves the other four looking unverified to anyone glancing at the list, which is not a great signal for a reviewer deciding what to pick up. #2821 is already approved by @SaptakS but still shows nothing green.

Is there something I should do differently, or is it just that a maintainer needs to approve the workflow run once for a first-time contributor? Happy to wait — I would just rather not have four PRs sitting there looking untested when they are not.

SaptakS:
@kunalKumar-13 the CIs start once maintainers approve the CI workflow. I have run it on #2821. Once we start reviewing the other PRs, we will approve the CIs. Nothing needed to be done differently.

kunalKumar-13:
Thanks @SaptakS, that's good to know — and thanks for running it on #2821, which merged.

I'll leave the others as they are then, rather than pushing empty commits to try to trigger anything.

kunalKumar-13:
@sarahboyce — following up on the Firefox failures you asked to treat as a blocker. They're not blocking: it's a Firefox bug, and it doesn't affect CI.

**CI runs both browsers and is green.** The workflow installs `chromium firefox` and runs `make ci TEST_ARGS="--browser chromium --browser firefox"`:

```
test_cycle_theme_when_prefers_dark (djangoproject.tests.FirefoxEndToEndTests) ... ok
test_cycle_theme_when_prefers_light (...FirefoxEndToEndTests) ... ok
test_init_light_dark_theme_uses_existing_cookie (...FirefoxEndToEndTests) ... ok
test_search_ctrl_k_hotkey (...FirefoxEndToEndTests) ... ok
test_search_placeholder_mac_mode (...FirefoxEndToEndTests) ... ok
test_set_theme_updates_data_attribute (...FirefoxEndToEndTests) ... ok
Ran 554 tests in 79.368s
OK (skipped=1)
```

**The failure I originally reported is macOS-only.** I re-ran the minimal reproduction on the *same builds* CI uses — playwright 1.62.0, firefox v1538 (Firefox 153.0):

| engine | `Cross-Origin-Opener-Policy` | `prefers-color-scheme: dark` |
|---|---|---|
| firefox | absent | True |
| firefox | `same-origin` | **False** — emulation silently dropped |
| chromium | absent | True |
| chromium | `same-origin` | True |

Same result whether the colour scheme is set on the browser context or via `emulate_media`. Django sends that header by default (`SECURE_CROSS_ORIGIN_OPENER_POLICY`), and there's no override in this repo's settings, so every live-server page carries it.

Reproducible with Django out of the picture entirely — a stdlib `http.server` returning `<html><body>hi</body></html>` and that one header is enough.

CI is Linux with those identical builds and passes, so what I hit was my machine, not the test and not the site. I flagged it as unresolved earlier in the thread, so it seemed worth correcting explicitly rather than leaving it hanging.

All five of your review points are addressed in `17d031b5` — the `@for_each_browser` decorator is gone in favour of a metaclass following `django.test.selenium.SeleniumTestCaseBase`, so tests need no decoration at all; colons removed from the comments; the `SeleniumTestCaseBase` docstring reference dropped since it no longer exists on main; e2e tests skip explicitly when no `--browser` is passed; and the GitHub action runs firefox.

sarahboyce:
Thank you for clarifying!
This is on my list to come back to for a final review but it might take a while before I do this 👍

sarahboyce (approved):
Thank you

PR #2832, opened 2026-09-04
Fixed #1766 -- Matched the leadership-level wording to how donations are totalled

### LINKED ISSUE #1766: The fundraising text for 1000$ donors is inconsistent

kunalKumar-13:
Fixed #1766 -- Matched the leadership-level wording to how donors are picked.

`DjangoHeroManager.for_public_display()` filters payments to the last
`DISPLAY_DONOR_DAYS` days and only then sums them, so the total that decides who
is listed as a leader covers a rolling 365-day window rather than a lifetime or a
calendar year. Neither place that describes the threshold said so: after #2191
the donor roll read "contribute at least $1,000 and have donated in the last 365
days", which suggests a lifetime total plus any recent donation, and the logo
help text on the donor form still gave no period at all.

This states the same rule in both places. It also adds a test for the manager,
pinning the rolling window down as the behaviour the wording describes, and a
test that keeps the form's help text tied to `LEADERSHIP_LEVEL_AMOUNT` and
`DISPLAY_DONOR_DAYS`.

One thing worth a second opinion before merging: in the issue thread the intent
was described as "any donor whose total donations are $1000 and above, regardless
of how many payments that is", which the rolling window does not literally
implement — a donor who gave $600 in each of two consecutive years is never a
leader. This PR only makes the text describe the current query. If the board
would rather the query changed, that is a separate ticket and I'm happy to open
it.

Tests: `fundraising` goes from 54 to 56 tests, all passing before and after; the
full suite goes from 548 to 550 with the same three pre-existing
`LocaleSmokeTests` failures on either side. Reverting only the form and the
template, keeping both tests, fails the help text test; the manager test still
passes, since it documents existing behaviour rather than a change to it.

Claude Code (Claude Opus) was used to trace the `filter()`-before-`annotate()`
behaviour in `for_public_display()`, draft the two replacement strings and write
the two tests. I confirmed the rolling-window semantics by running a throwaway
test that gave one donor a $1,000 payment older than the window and a $200
payment inside it and read back `donated_amount = 200.00`. I ran the
`fundraising` suite and the full test suite before and after (54 → 56 and
548 → 550, same three pre-existing `LocaleSmokeTests` failures on both sides),
and confirmed that reverting the form and the template fails the help text test.
The manager test passes either way, by design. No manual or browser testing was
performed.

### DIFF
--- a/djangoproject/templates/fundraising/includes/display_django_heroes.html
+++ b/djangoproject/templates/fundraising/includes/display_django_heroes.html
@@ -74,8 +74,8 @@
     Leaders (${{ amount }}+){% endblocktranslate %}</h3>
   <p>
     {% blocktranslate trimmed with amount=display_logo_amount|intcomma %}
-      Leadership-level donors contribute at least ${{ amount }}
-      and have donated in the last {{ display_donor_days }} days.
+      Leadership-level donors have contributed at least ${{ amount }}
+      in the last {{ display_donor_days }} days.
     {% endblocktranslate %}
   </p>
   <div>
--- a/fundraising/forms.py
+++ b/fundraising/forms.py
@@ -7,7 +7,13 @@ from django.utils.translation import gettext_lazy as _
 from django_recaptcha.fields import ReCaptchaField
 from django_recaptcha.widgets import ReCaptchaV3
 
-from .models import INTERVAL_CHOICES, LEADERSHIP_LEVEL_AMOUNT, DjangoHero, Donation
+from .models import (
+    DISPLAY_DONOR_DAYS,
+    INTERVAL_CHOICES,
+    LEADERSHIP_LEVEL_AMOUNT,
+    DjangoHero,
+    Donation,
+)
 
 
 class DjangoHeroForm(forms.ModelForm):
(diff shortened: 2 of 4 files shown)


sarahboyce:
Thank you, but this was already fixed by github.com/django/djangoproject.com/pull/2191
I will close the associated ticket

kunalKumar-13:
Thanks for checking — you're right, #2191 covers it. Sorry for the noise; I searched open PRs for a reference to the ticket but missed a merged one from that far back. Closing here too.

PR #2834, opened 2026-09-06
Retrieved less document metadata during search.

jacobtylerwalls:
Noticed while doing bc56509502edb0d831ab8814b38447660141d6fb that we retrieve large JSON objects only for the sake of extracting the `"parents"` value.

This branch now only extracts that one value.

Maybe this is a "too small" optimization, but wanted to float it anyway.

### DIFF
--- a/docs/models.py
+++ b/docs/models.py
@@ -290,7 +290,11 @@ def _clean_document_path(path):
 
 
 def document_url(doc):
-    if doc.metadata.get("parents") == DocumentationCategory.WEBSITE:
+    try:
+        parents = doc.parents  # From a search()/search_results() annotation.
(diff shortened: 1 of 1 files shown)


pauloxnet:
A better approach could be to create generated fields for all "metadata" keys and use them directly avoiding get the whole metadata Json from the database.

sarahboyce (approved):
Thank you!

> A better approach could be to create generated fields for all "metadata" keys and use them directly avoiding get the whole metadata Json from the database.

I agree that generated fields may be worth exploring separately 👍 

This change feels like an easy win (small, well contained, no migrations - let's try it out ⭐ )

PR #2842, opened 2026-09-08
Fixed #2841 -- Handled an out-of-range days parameter in metric_json.

### LINKED ISSUE #2841: dashboard metric_json returns 500 for an out-of-range days parameter
## Description

`dashboard.views.metric_json` returns a 500 for a `days` value that `int()` accepts but a `timedelta` cannot represent.

```
GET /metric/new-tickets-week.json?days=3652059
GET /metric/new-tickets-week.json?days=999999999
GET /metric/new-tickets-week.json?days=-999999999
```

```
File "dashboard/views.py", line 60, in metric_json
    start_date = datetime.datetime.now() - datetime.timedelta(days=daysback)
OverflowError: date value out of range
```

The parameter is already guarded, but only against parsing:

```python
try:
    daysback = int(request.GET["days"])
except (TypeError, KeyError, ValueError):
    daysback = 30
```

`int("3652059")` succeeds, so the guard passes it through and the failure happens further down, at the `timedelta`. Widening that `except` does not help — it is the wrong scope.

`?days=abc`, `?days=`, `?days=-5` and a missing parameter are all handled correctly today and return 200; only the out-of-range case gets through.

## Reproduced

Locally, against `main` at 9a4867cf, through the view itself:

```
days='30'         -> HTTP 200
days='abc'        -> HTTP 200
days='-5'         -> HTTP 200
days=''           -> HTTP 200
days='3652059'    -> UNHANDLED OverflowError: date value out of range
days='999999999'  -> UNHANDLED OverflowError: date value out of range
```

I have deliberately not sent any of these to dashboard.djangoproject.com — the code path is identical and a live request would only add noise to your error reporting.

## Suggested fix

Build the window inside the guard, so an unusable number is caught the same way an unparseable one already is, and fall back to the same 30-day default. Happy to open the PR.

kunalKumar-13:
Fixes #2841.

`metric_json` guards the `days` parameter against parsing only:

```python
try:
    daysback = int(request.GET["days"])
except (TypeError, KeyError, ValueError):
    daysback = 30
```

`int("3652059")` succeeds, so the guard passes it through and the request fails further down at

```python
start_date = datetime.datetime.now() - datetime.timedelta(days=daysback)
OverflowError: date value out of range
```

Widening that `except` would not have helped — the error is raised outside the block. The window is now built inside the guard, so a value too large for a `timedelta` falls back to the same 30-day default an unparseable one already does. `start_date` is computed once and reused instead of being recomputed on a cache miss.

### Before and after, through the view

| `days` | before | after |
|---|---|---|
| `30` | 200 | 200 |
| `abc` | 200 | 200 |
| `` (empty) | 200 | 200 |
| `-5` | 200 | 200 |
| `3652059` | **OverflowError** | 200 |
| `999999999` | **OverflowError** | 200 |
| `-999999999` | **OverflowError** | 200 |

Everything that worked before still works; only the three that raised now return the default window.

### Tests

Written first and run before the fix — all three out-of-range values failed with the real `OverflowError` at `views.py:60`. `test_metric_json_bad_days_falls_back` covers the values that were already handled, so a future change cannot quietly regress them.

`python -m manage test dashboard docs` — **101 tests, OK**. `black` and `isort` clean on both changed files.

I did not send any of these to dashboard.djangoproject.com. The code path is identical and a live request would only have added noise to your error reporting.

### DIFF
--- a/dashboard/views.py
+++ b/dashboard/views.py
@@ -53,16 +53,18 @@ def metric_json(request, metric_slug):
 
     try:
         daysback = int(request.GET["days"])
-    except (TypeError, KeyError, ValueError):
+        # int() accepts values a timedelta cannot represent, so the window has
+        # to be built here to find out whether the number is usable at all.
+        start_date = datetime.datetime.now() - datetime.timedelta(days=daysback)
+    except (TypeError, KeyError, ValueError, OverflowError):
         daysback = 30
+        start_date = datetime.datetime.now() - datetime.timedelta(days=daysback)
 
     generation = generation_key()
     key = f"dashboard:metric:{metric_slug}:{daysback}"
 
     doc = cache.get(key, version=generation)
     if doc is None:
-        start_date = datetime.datetime.now() - datetime.timedelta(days=daysback)
-
         doc = model_to_dict(metric)
         doc["data"] = metric.gather_data(since=start_date)
         cache.set(key, doc, 60 * 60, version=generation)
--- a/dashboard/tests.py
+++ b/dashboard/tests.py
@@ -71,6 +71,35 @@ class ViewTests(ReleaseMixin, TestCase):
         self.assertEqual(json.loads(response.content.decode())["data"][0][1], 42)
         self.assertEqual(response.status_code, 200)
 
+    def test_metric_json_days_out_of_range(self):
+        """A days value too large for a timedelta must not 500.
+
+        int() accepts it, so the existing guard passes it through, and the
+        failure happens later at the timedelta.
+        """
+        TracTicketMetric.objects.get(slug="new-tickets-week").data.create(
+            measurement=42
+        )
+        url = reverse("metric-json", args=["new-tickets-week"], host="dashboard")
+        for days in ["3652059", "999999999", "-999999999"]:
+            with self.subTest(days=days):
+                request = self.factory.get(f"{url}?days={days}")
+                response = metric_json(request, "new-tickets-week")
+                self.assertEqual(response.status_code, 200)
+
+    def test_metric_json_bad_days_falls_back(self):
+        """Unusable values keep the default window rather than erroring."""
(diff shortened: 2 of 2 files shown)


dashboard/views.py
```diff
@@ -53,16 +53,18 @@ def metric_json(request, metric_slug):
...
         daysback = 30
+        start_date = datetime.datetime.now() - datetime.timedelta(days=daysback)
 
     generation = generation_key()
     key = f"dashboard:metric:{metric_slug}:{daysback}"
 
     doc = cache.get(key, version=generation)
     if doc is None:
-        start_date = datetime.datetime.now() - datetime.timedelta(days=daysback)
-
```
p-r-a-v-i-n, line 65: What was reasoning for removing this ? why we were calculating `start_date`  previously under the `if doc is None:` block ?

kunalKumar-13:
Good question — the short answer is that the `if doc is None:` block sits outside the `try`, and that is where the crash was.

`except` only covers what is inside `try`. `int()` accepts values that `timedelta` cannot represent, so the failure happened later, on a cache miss, uncaught, as a 500.

There are three distinct ways it fails, all `OverflowError`:

| `?days=` | what raises |
|---|---|
| `999999999` | `timedelta` is fine — it is exactly `timedelta.max.days` — and the **subtraction** overflows: *date value out of range* |
| `1000000000` | `timedelta()` itself: *days=1000000000; must have magnitude <= 999999999* |
| `999999999999` | *Python int too large to convert to C int* |

The first row is why the whole expression moved rather than just `int()`. At exactly `timedelta.max.days` the constructor succeeds and `datetime.now() - delta` is what blows up, so wrapping only the `int()` call would still 500 on `?days=999999999`.

The cost is one `datetime.now()` and a subtraction on cache hits, where it used to be skipped. Happy to compute it lazily instead if you would rather keep that off the hot path, though it seemed a poor trade for the extra branch.

PR #2844, opened 2026-09-10
Added optional django-storages (S3) support for media.

### LINKED ISSUE #2802: Add optional django-storages (S3) support for media
Uploaded media and release artifacts currently live on the server's local file system (`MEDIA_ROOT = DATA_DIR / "media"` in `prod.py`), which must be copied when replacing the production server. This issue adds **optional** S3 storage via [django-storages](https://django-storages.readthedocs.io/), selected by environment configuration. Local file system storage remains the default; the feature is enabled from the deployment repo once this lands.

## Changes (this repo)

- **Dependency**: add `django-storages[boto3]` (prod group in `pyproject.toml`, regenerate `requirements/prod.txt`).
- **Settings** (`djangoproject/settings/prod.py`): select `STORAGES["default"]`'s backend based on environment configuration (e.g., use `storages.backends.s3.S3Storage` when the bucket name is set, otherwise keep `FileSystemStorage`). All S3 config (bucket, region) comes from the environment / `secrets.json`; nothing hardcoded. The EC2 instance will have a role with S3 access, so no explicit credentials are required.
- **Release artifacts** (`releases/models.py`): `get_storage()` explicitly returns `FileSystemStorage(allow_overwrite=True)`; make it return the S3 equivalent (with `allow_overwrite=True`) when S3 is enabled, so artifact uploads keep their exact names.
- **Docs release zips** (`docs/management/commands/build_doc_release.py`): the docs HTML zip is written directly to `settings.MEDIA_ROOT / "docs"`; write it through the default storage instead so it lands in S3 when enabled.
- **sorl-thumbnail** (fundraising logos, `djangoproject/thumbnails.py`): verify `get_thumbnail()` works with an S3 source/storage, and configure `THUMBNAIL_STORAGE` so generated thumbnails are written to the bucket.
- **Tests**: suite must pass without AWS credentials (default configuration unchanged); add coverage for the S3 configuration path (backend selection and file operations against a mocked S3 backend, e.g. `moto` or `django-storages`'s in-memory S3).

## Out of scope

- **Enabling in the deployment repo** (private Ansible repo): S3 bucket, the env vars that turn the feature on, one-time copy of existing `media/` contents, nginx changes for `media.djangoproject.com` (including the pre-1.4 `/css|img/` redirect), and dropping the `media/` directory from automated backups.
- **Trac**: config and attachments stay on the file system.
- **Docs rendering**: `DOCS_BUILD_ROOT` and the docs cache remain local; only the downloadable zip is affected (see above).

## AI Disclosure

This issue description was drafted with the assistance of a local model, Qwen 3.8 27B.

saikat709:
Fixes #2802

### Description
This PR adds optional Amazon S3 storage support via django-storages[boto3], enabled when environment provided `AWS_STORAGE_BUCKET_NAME` or secrets.json configuration provides `aws_storage_bucket_name`. Local file system storage (FileSystemStorage) remains the default when S3 configuration is absent.
Also added relevant tests and verified by running them locally. All works pefectly.

Gemini 3.6 Flash was used for discussion on tests and writing down test drafts. Code generated by LLM was manually verified.

### DIFF
--- a/djangoproject/settings/prod.py
+++ b/djangoproject/settings/prod.py
@@ -52,14 +52,35 @@ MIDDLEWARE = (
 
 SESSION_COOKIE_SECURE = True
 
-STORAGES = {
-    "default": {
-        "BACKEND": "django.core.files.storage.FileSystemStorage",
-    },
-    "staticfiles": {
-        "BACKEND": "django.contrib.staticfiles.storage.ManifestStaticFilesStorage",
-    },
-}
+AWS_STORAGE_BUCKET_NAME = os.getenv("AWS_STORAGE_BUCKET_NAME") or SECRETS.get(
+    "aws_storage_bucket_name"
+)
+if AWS_STORAGE_BUCKET_NAME:
+    STORAGES = {
+        "default": {
+            "BACKEND": "storages.backends.s3.S3Storage",
+        },
+        "staticfiles": {
+            "BACKEND": "django.contrib.staticfiles.storage.ManifestStaticFilesStorage",
+        },
+    }
+    AWS_S3_REGION_NAME = os.getenv("AWS_S3_REGION_NAME") or SECRETS.get(
+        "aws_s3_region_name"
+    )
+    if aws_s3_endpoint_url := (
+        os.getenv("AWS_S3_ENDPOINT_URL") or SECRETS.get("aws_s3_endpoint_url")
+    ):
+        AWS_S3_ENDPOINT_URL = aws_s3_endpoint_url
+    THUMBNAIL_STORAGE = STORAGES["default"]["BACKEND"]
+else:
+    STORAGES = {
+        "default": {
+            "BACKEND": "django.core.files.storage.FileSystemStorage",
+        },
+        "staticfiles": {
+            "BACKEND": "django.contrib.staticfiles.storage.ManifestStaticFilesStorage",
+        },
+    }
 
 STATIC_ROOT = DATA_DIR / "static"
 
--- a/djangoproject/tests.py
+++ b/djangoproject/tests.py
@@ -365,3 +365,62 @@ class EndToEndTests(ReleaseMixin, StaticLiveServerTestCase):
                 page.locator(".theme-toggle").click()
                 theme = page.evaluate("document.documentElement.dataset.theme")
                 self.assertEqual(theme, expected_theme)
+
+
+class S3StorageSettingsTests(TestCase):
+    def test_prod_settings_with_s3_bucket(self):
+        with patch.dict(
+            os.environ,
+            {
+                "DJANGO_DB_NAME": "djangoproject",
+                "AWS_STORAGE_BUCKET_NAME": "my-s3-bucket",
+            },
+        ):
+            import importlib
+
+            import djangoproject.settings.prod as prod_settings
+
+            importlib.reload(prod_settings)
+            self.assertEqual(
+                prod_settings.STORAGES["default"]["BACKEND"],
+                "storages.backends.s3.S3Storage",
+            )
+            self.assertEqual(
+                prod_settings.THUMBNAIL_STORAGE,
+                "storages.backends.s3.S3Storage",
+            )
+
+    def test_prod_settings_without_s3_bucket(self):
+        env = os.environ.copy()
+        env.setdefault("DJANGO_DB_NAME", "djangoproject")
+        env.pop("AWS_STORAGE_BUCKET_NAME", None)
+        with patch.dict(os.environ, env, clear=True):
+            import importlib
+
+            import djangoproject.settings.prod as prod_settings
+
+            importlib.reload(prod_settings)
+            self.assertEqual(
+                prod_settings.STORAGES["default"]["BACKEND"],
+                "django.core.files.storage.FileSystemStorage",
+            )
+
+
+class SorlThumbnailS3Tests(TestCase):
+    def test_logo_thumbnail_mixin(self):
+        from djangoproject.thumbnails import LogoThumbnailMixin
+
+        class TestModel(LogoThumbnailMixin):
+            def __init__(self, logo_name):
+                class DummyLogo:
+                    name = logo_name
+
+                    def __bool__(self):
+                        return bool(self.name)
+
+                self.logo = DummyLogo()
+
+        obj = TestModel("test_logo.png")
+        self.assertFalse(obj.logo_is_svg)
+        svg_obj = TestModel("test_logo.svg")
+        self.assertTrue(svg_obj.logo_is_svg)
--- a/docs/management/commands/build_doc_release.py
+++ b/docs/management/commands/build_doc_release.py
@@ -21,11 +21,12 @@ import os
 import shutil
 import subprocess
 import sys
+import tempfile
 import zipfile
-from contextlib import closing
 from pathlib import Path
 
 from django.conf import settings
+from django.core.files.storage import default_storage
 from django.core.management import BaseCommand, CommandError
 from django.utils.translation import to_locale
 from sphinx.application import Sphinx
@@ -149,24 +150,25 @@ class Command(BaseCommand):
         #
         html_build_dir = parent_build_dir / "_build" / html_builder
         zipfile_name = f"django-docs-{release.version}-{release.lang}.zip"
-        zipfile_path = settings.MEDIA_ROOT / "docs" / zipfile_name
-        if not zipfile_path.parent.exists():
-            zipfile_path.parent.mkdir(parents=True)
+        target_path = f"docs/{zipfile_name}"
         if self.verbosity >= 2:
-            self.stdout.write("  build zip (into %s)" % zipfile_path)
+            self.stdout.write("  build zip (into %s)" % target_path)
 
         def zipfile_inclusion_filter(file_path):
             return ".doctrees" not in file_path.parts
 
-        with closing(
-            zipfile.ZipFile(str(zipfile_path), "w", compression=zipfile.ZIP_DEFLATED)
-        ) as zf:
-            for root, dirs, files in os.walk(str(html_build_dir)):
-                for f in files:
-                    file_path = Path(os.path.join(root, f))
-                    if zipfile_inclusion_filter(file_path):
-                        rel_path = str(file_path.relative_to(html_build_dir))
-                        zf.write(str(file_path), rel_path)
+        with tempfile.NamedTemporaryFile(suffix=".zip") as tmp:
+            with zipfile.ZipFile(tmp, "w", compression=zipfile.ZIP_DEFLATED) as zf:
+                for root, dirs, files in os.walk(str(html_build_dir)):
+                    for f in files:
+                        file_path = Path(os.path.join(root, f))
+                        if zipfile_inclusion_filter(file_path):
+                            rel_path = str(file_path.relative_to(html_build_dir))
+                            zf.write(str(file_path), rel_path)
+            tmp.seek(0)
+            if default_storage.exists(target_path):
+                default_storage.delete(target_path)
+            default_storage.save(target_path, tmp)
 
         #
         # Copy the build results to the directory used for serving
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,6 +47,7 @@ dev = [
     { include-group = "tests" },
 ]
 prod = [
+    "django-storages[boto3]",
     "gunicorn==26.0.0",
     "psycopg[c]==3.3.4",
     "redis==8.1.0",
--- a/releases/models.py
+++ b/releases/models.py
@@ -6,7 +6,7 @@ from pathlib import Path
 from django.conf import settings
 from django.core.cache import cache
 from django.core.exceptions import ValidationError
-from django.core.files.storage import FileSystemStorage
+from django.core.files.storage import FileSystemStorage, storages
 from django.core.validators import RegexValidator
 from django.db import models
 from django.utils.functional import cached_property
@@ -157,12 +157,15 @@ artifact_name_re = re.compile(
 
 def get_storage():
     """
-    Return a FileSystemStorage that allows file name overwrites.
+    Return a storage instance that allows file name overwrites.
 
     The actual file name of release artifacts (tarball, wheel, ...) should not
     be modified on upload (i.e. no prefix should be added).
     """
-    return FileSystemStorage(allow_overwrite=True)
+    storage_cls = storages["default"].__class__
+    if issubclass(storage_cls, FileSystemStorage):
+        return storage_cls(allow_overwrite=True)
+    return storage_cls(file_overwrite=True)
 
 
 def upload_to_artifact(release, filename):
--- a/releases/tests.py
+++ b/releases/tests.py
@@ -1170,3 +1170,35 @@ class RoadmapViewTestCase(ReleaseMixin, TestCase):
             "Django 5.2 final will ideally ship two weeks after the last RC.",
         )
         self.assertNotContains(response, "ship in January.")
+
+
+class ReleaseStorageTests(TestCase):
+    def test_get_storage_default_filesystem(self):
+        from django.core.files.storage import FileSystemStorage
+
+        from .models import get_storage
+
+        storage = get_storage()
+        self.assertIsInstance(storage, FileSystemStorage)
+        self.assertTrue(
+            getattr(
+                storage, "_allow_overwrite", getattr(storage, "allow_overwrite", False)
+            )
+        )
+
+    def test_get_storage_s3_backend(self):
+        from .models import get_storage
+
+        with override_settings(
+            STORAGES={
+                "default": {"BACKEND": "storages.backends.s3.S3Storage"},
+                "staticfiles": {
+                    "BACKEND": "django.contrib.staticfiles.storage.ManifestStaticFilesStorage"
+                },
+            }
+        ):
+            try:
+                storage = get_storage()
+                self.assertTrue(getattr(storage, "file_overwrite", False))
+            except Exception:
+                pass
--- a/docs/tests/test_commands.py
+++ b/docs/tests/test_commands.py
@@ -136,3 +136,44 @@ class UpdateDocsResilienceTests(TestCase):
 
         self.assertEqual(attempted, [failing_release, other_release])
         mock_capture.assert_called_once_with(error, flush=True)
+
+
+class BuildDocReleaseStorageTests(TestCase):
+    def setUp(self):
+        self.release = DocumentRelease.objects.create(lang="en", release=None)
+        self.command = Command()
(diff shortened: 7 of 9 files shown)


tobiasmcnulty:
@saikat709 Thanks for working on this! I noticed that you added a `AWS_S3_ENDPOINT_URL`, thank you for doing that. Could you confirm if you've tested this locally with Minio or another S3-compatible docker container? It could be worth adding your preferred test setup `docker-compose.yml` to help reviewers (or future editors of this integration) with testing the setup in local development, along with the applicable instructions in the README for how the S3 integration works and can be tested locally.

saikat709:
@tobiasmcnulty, Thanks for guidance! 
Though I have added tests that cheks for S3 or Filesystem selection. None of them directlt upload to S3, So I will test it and verify the setup locally with MinIO, add the MinIO service to docker-compose.yml, and include local S3 testing instructions in README.rst in a follow-up commit shortly. 

I have one question here, should I add another test that will actually upload to storage, in S3 tests?? Though it will require us to run the docker container/ running any S3 service when running those tests. Or should only test locally running the docker container for minio and uplload a file from admin to verify it?

tobiasmcnulty:
Personally, I use Minio for manual testing and local development only (unit tests don't use Minio). Others folks may have other thoughts on how to structure it, though. My main goal was to provide a reviewer a way to verify the changes without access to an AWS account.

ulgens (commented):
Any chance adding https://github.com/floci-io/floci (or something with similar purposes) to the local docker setup, so we can match the new infra config better?

djangoproject/tests.py
```diff
@@ -365,3 +365,62 @@ def test_cycle_theme_when_prefers_light(self):
...
+
+
+class S3StorageSettingsTests(TestCase):
+    def test_prod_settings_with_s3_bucket(self):
+        with patch.dict(
+            os.environ,
+            {
+                "DJANGO_DB_NAME": "djangoproject",
+                "AWS_STORAGE_BUCKET_NAME": "my-s3-bucket",
+            },
```
ulgens, line 377: Does `override_settings` cover the case here?

pyproject.toml
```diff
@@ -47,6 +47,7 @@ dev = [
     { include-group = "tests" },
 ]
 prod = [
+    "django-storages[boto3]",
```
ulgens, line 50: Please pin to the patch version.

releases/models.py
```diff
@@ -157,12 +157,15 @@ def current_version(self):
...
+    Return a storage instance that allows file name overwrites.
 
     The actual file name of release artifacts (tarball, wheel, ...) should not
     be modified on upload (i.e. no prefix should be added).
     """
-    return FileSystemStorage(allow_overwrite=True)
+    storage_cls = storages["default"].__class__
+    if issubclass(storage_cls, FileSystemStorage):
+        return storage_cls(allow_overwrite=True)
+    return storage_cls(file_overwrite=True)
```
ulgens, line 168: It seems to return the same in both cases.

ulgens (changes requested):

docs/management/commands/build_doc_release.py
```diff
@@ -149,24 +150,25 @@ def build_doc_release(self, release):
...
-            zipfile.ZipFile(str(zipfile_path), "w", compression=zipfile.ZIP_DEFLATED)
-        ) as zf:
-            for root, dirs, files in os.walk(str(html_build_dir)):
-                for f in files:
-                    file_path = Path(os.path.join(root, f))
-                    if zipfile_inclusion_filter(file_path):
-                        rel_path = str(file_path.relative_to(html_build_dir))
-                        zf.write(str(file_path), rel_path)
+        with tempfile.NamedTemporaryFile(suffix=".zip") as tmp:
+            with zipfile.ZipFile(tmp, "w", compression=zipfile.ZIP_DEFLATED) as zf:
```
ulgens, line 161: Can we flatten the `with`s here?

---

Context: https://docs.astral.sh/ruff/rules/multiple-with-statements/

> In Python 3, a single with block can include multiple context managers.
>
> Combining multiple context managers into a single with statement will minimize the indentation depth of the code, making it more readable.

releases/tests.py
```diff
@@ -1170,3 +1170,35 @@ def test_release_candidate_description_non_calendar_version(self):
...
+                "default": {"BACKEND": "storages.backends.s3.S3Storage"},
+                "staticfiles": {
+                    "BACKEND": "django.contrib.staticfiles.storage.ManifestStaticFilesStorage"
+                },
+            }
+        ):
+            try:
+                storage = get_storage()
+                self.assertTrue(getattr(storage, "file_overwrite", False))
+            except Exception:
```
ulgens, line 1203: Please don't use bare excepts. What is the expected error here?

---

Context: https://docs.astral.sh/ruff/rules/bare-except/

> A bare except catches BaseException which includes KeyboardInterrupt, SystemExit, Exception, and others. Catching BaseException can make it hard to interrupt the program (e.g., with Ctrl-C) and can disguise other problems.

ulgens: Also, not sure why we need this kind of error handling in a test.

ulgens:
If @django/django-website and @django/ops-team is generally positive about using S3, I'd recommend sending the dependency changes as a separate PR. It can be merged quickly and prevent potential merge conflicts that would be caused by dependency changes - updates from other PRs.

djangoproject/settings/prod.py
```diff
@@ -52,14 +52,35 @@
...
 
-STORAGES = {
-    "default": {
-        "BACKEND": "django.core.files.storage.FileSystemStorage",
-    },
-    "staticfiles": {
-        "BACKEND": "django.contrib.staticfiles.storage.ManifestStaticFilesStorage",
-    },
-}
+AWS_STORAGE_BUCKET_NAME = os.getenv("AWS_STORAGE_BUCKET_NAME") or SECRETS.get(
```
ulgens, line 55: Is the "env or secrets" a new pattern with our settings? Our current env - secret handling needs some improvements, but I'd recommend following the existing pattern for now.

releases/models.py
saikat709: For [s3](https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html), `file storages uses` `file_overwrite` parameter. But in django's own Filesystemstorage uses `allow_overwrite` parameter [check](https://docs.djangoproject.com/en/6.1/ref/files/storage/#the-filesystemstorage-class)

docs/management/commands/build_doc_release.py
saikat709: Okay.

releases/tests.py
saikat709: Thank you. as you suggested, i checked the code and removed the unnecessary try catch.

pyproject.toml
saikat709: updating it to `django-storages[boto3]>=1.14`

djangoproject/tests.py
saikat709: I'll update to secrets, instead of environ. And used `override_settings`.

saikat709:
Thank you @ulgens for reviewing my PR. I have pushed updates based on your suggestions.
I was unsure about the floci, thats why added minio for s3 local setup in docker. Happy to get a second review!

saikat709:
Hi @tobiasmcnulty, I have added minio docker setup and added respective readme instructions for local test. Also I myself tested and verified that it works perfectly.
I like your idea not having a actual s3 upload test on unit tests. So i did not include any test for test. Rather add tests that ensures if our code to pick the right file storage works or not.

djangoproject/tests.py
```diff
@@ -365,3 +365,49 @@ def test_cycle_theme_when_prefers_light(self):
...
+
+
+class S3StorageSettingsTests(TestCase):
+    def test_prod_settings_with_s3_bucket(self):
+        with override_settings(
+            SECRETS={**settings.SECRETS, "aws_storage_bucket_name": "my-s3-bucket"}
+        ):
+            bucket_name = settings.SECRETS.get("aws_storage_bucket_name")
+            backend = (
+                "storages.backends.s3.S3Storage"
```
ulgens, line 377: If there are multiple cases the test can run, please separate them into different tests or parameterize. With the current approach, one case will always be ignored.

releases/tests.py
```diff
@@ -1170,3 +1170,32 @@ def test_release_candidate_description_non_calendar_version(self):
             "Django 5.2 final will ideally ship two weeks after the last RC.",
         )
         self.assertNotContains(response, "ship in January.")
+
+
+class ReleaseStorageTests(TestCase):
```
ulgens, line 1175: I couldn't catch the case here. How these tests relate to releases?

docker-compose.yml
```diff
@@ -49,3 +49,24 @@ services:
             # to be loaded into the database manually (via 02_trac_schema.sh)
             # since it applies to the trac database, not the main Django database.
             - ./tracdb/trac.sql:/trac.sql
+
+    minio:
+        image: quay.io/minio/minio:latest
```
ulgens: Please pin to patch version.

docker-compose.yml
```diff
@@ -49,3 +49,24 @@ services:
...
+        command: server /data --console-address ":9001"
+        ports:
+            - "9000:9000"
+            - "9001:9001"
+        environment:
+            - MINIO_ROOT_USER=minioadmin
+            - MINIO_ROOT_PASSWORD=minioadmin
+        healthcheck:
+            test: ["CMD", "curl", "-f", "http://localhost:9000/minio/health/live"]
```
ulgens: This seems to hardcode domain and port. I guess we can at least access the port dynamically.

README.rst
```diff
@@ -479,6 +479,39 @@ Running Locally with Docker
...
+1. Start the MinIO container using Docker Compose::
+
+    docker compose up -d minio
+
+2. Access the MinIO Web Console at http://localhost:9001 in your browser:
+
+   - **Username**: ``minioadmin``
+   - **Password**: ``minioadmin``
+
+3. Create a bucket named ``djangoproject`` in the MinIO Web Console if it does not already exist (Object Browser -> Create Bucket).
```
ulgens: This should be automatic.

README.rst
```diff
@@ -479,6 +479,39 @@ Running Locally with Docker
...
 7. View the docs at http://docs.djangoproject.localhost:8000/.
 
+Testing S3 Storage Locally with MinIO
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To manually test S3-compatible media storage locally without an AWS account:
+
+1. Start the MinIO container using Docker Compose::
+
+    docker compose up -d minio
```
ulgens: Doesn't it automatically start with when compose setup is up?

README.rst
```diff
@@ -479,6 +479,39 @@ Running Locally with Docker
 
 7. View the docs at http://docs.djangoproject.localhost:8000/.
 
+Testing S3 Storage Locally with MinIO
```
ulgens, line 482: I couldn't catch the overall intention of this section. Services should be automatically set up and working when compose is up.

releases/models.py
ulgens: Thank you, my bad.

pyproject.toml
ulgens: Pinned version would use `==`

ulgens (changes requested):

djangoproject/settings/common.py
```diff
@@ -368,3 +368,31 @@
...
+        },
+        "staticfiles": {
+            "BACKEND": "django.contrib.staticfiles.storage.StaticFilesStorage",
+        },
+    }
+    AWS_S3_REGION_NAME = SECRETS.get("aws_s3_region_name")
+    if aws_s3_endpoint_url := SECRETS.get("aws_s3_endpoint_url"):
+        AWS_S3_ENDPOINT_URL = aws_s3_endpoint_url
+    if aws_access_key_id := SECRETS.get("aws_access_key_id"):
+        AWS_ACCESS_KEY_ID = aws_access_key_id
```
ulgens: Is this different than using 

```
AWS_ACCESS_KEY_ID = SECRETS.get("aws_access_key_id")
```

djangoproject/tests.py
```diff
@@ -365,3 +365,49 @@ def test_cycle_theme_when_prefers_light(self):
...
+                if bucket_name
+                else "django.core.files.storage.FileSystemStorage"
+            )
+            self.assertEqual(backend, "storages.backends.s3.S3Storage")
+
+    def test_prod_settings_without_s3_bucket(self):
```
ulgens, line 383: What is being tested here? Do we expect a prod setup with no S3?

djangoproject/tests.py
```diff
@@ -365,3 +365,49 @@ def test_cycle_theme_when_prefers_light(self):
...
+            bucket_name = settings.SECRETS.get("aws_storage_bucket_name")
+            backend = (
+                "storages.backends.s3.S3Storage"
+                if bucket_name
+                else "django.core.files.storage.FileSystemStorage"
+            )
+            self.assertEqual(backend, "django.core.files.storage.FileSystemStorage")
+
+
+class SorlThumbnailS3Tests(TestCase):
```
ulgens: This doesn't seem related to the PR context.

djangoproject/tests.py
saikat709: Can you please clarify this?

releases/tests.py
saikat709: While uploading release files, we need storage - FileSystemStorage or S3, since these two tests are also Release file related, thats why I put them here - Testing if override is being passed properly. 
Do you have any suggestion on this?

docker-compose.yml
saikat709: going to use `RELEASE.2025-09-07T16-13-09Z`

docker-compose.yml
saikat709: Thanks. Now I have edited them to be dynamic with default values.
e.g `MINIO_ROOT_USER=${MINIO_ROOT_USER:-minioadmin}`, same for others too. 
But i dont think even hardcoated value would be a problem.

README.rst
saikat709: updated. initially creates `djangoproject` container automatically.

README.rst
saikat709: yes. it does. also we can start seperately. updated readme accordingly.

README.rst
saikat709: FIxed this part of readme.

djangoproject/settings/common.py
saikat709: updated to
```python
AWS_ACCESS_KEY_ID = SECRETS.get("aws_access_key_id")
AWS_SECRET_ACCESS_KEY = SECRETS.get("aws_secret_access_key")
if aws_s3_endpoint_url := SECRETS.get("aws_s3_endpoint_url"):
     AWS_S3_ENDPOINT_URL = aws_s3_endpoint_url
```
Previosuly assumed that whole system will be hosted on AWS, where we can ommit aws_access_key_id and aws_secret_access_key. My bad,  kept `aws_s3_endpoint_url` optional, can be ommited in prod server. because boto3 uses standard aws url based on bucket namr and region name.

djangoproject/tests.py
saikat709: Since the S3 storage support is "Additional" support, thats why I have added both cases, according to the issue description.
It tests if we are are picking right file storage based on secrets.

djangoproject/tests.py
saikat709: No, it doesn't. When i was adding THUMBNAIL_STORAGE tests i saw the LogoThumbnailMixin does not have any test, thats why added one.

Should I remove it? Or we can keep this? Even if its out of scope of this pr.

saikat709:
Incorporated suggestions. @ulgens

djangoproject/settings/common.py
ulgens: We can make it explicit. The current approach allows it to be non-defined, which can be harder to handle than having an empty string.

djangoproject/tests.py
ulgens: I think it's better to stay in the context. We can create a different issue for missing tests for other components.

djangoproject/settings/common.py
saikat709: The accsss key and secret key already explicit. According to the django storage boto3 documentation, [here](https://django-storages.readthedocs.io/en/1.13.1/backends/amazon-S3.html), `AWS_S3_ENDPOINT_URL (optional: default is None)`.

For production we can keep it undefined or none in Secrets. In local test, we provide the localhost url.
I think i seems okay.


Let me know if we need to variable check when using s3 storage for required variables in secret. Otherwise if missing we throw an error.

djangoproject/tests.py
saikat709: Fair enough. I will remove the thumbnail mixin test and open a issue for it.

djangoproject/settings/common.py
ulgens: > AWS_S3_ENDPOINT_URL (optional: default is None).

This is more reason for just assigning it directly. We don't need the intermediary `aws_s3_endpoint_url` variable.

ulgens:
@saikat709 Can you please check https://github.com/django/djangoproject.com/pull/2844#issuecomment-5634793186 too? Thanks in advance.

djangoproject/settings/common.py
saikat709: Okay. Updating it in the next commit.

I need a suggestion, do we need to check for the variable are in secrets or missing when we are gonna use s3?
E.g. if we have bucket name we check if we have access key, secret key, url, region.. If any of them missing, we raise an error?

saikat709:
> If @django/django-website and @django/ops-team is generally positive about using S3, I'd recommend sending the dependency changes as a separate PR. It can be merged quickly and prevent potential merge conflicts that would be caused by dependency changes - updates from other PRs.

Since this is only adding a new dependency ("django-storages[boto3]") rather than updating an existing dependency, would it still be okay to keep it in this PR? I think it should be unlikely to cause a merge conflict.

ulgens:
@saikat709 It updates 2 lock files, it's highly likely to create a conflict (and additions can create conflict too, it doesn't directly relate to the type of the change.). It looks like the PR will take some more time to be completed, investing on no-conflicts early would help with the process.

djangoproject/settings/common.py
saikat709: Updated to explicit format.

saikat709:
Hi @ulgens! Based on our conversation I  have created an issue for that thumbnail test https://github.com/django/djangoproject.com/issues/2850, And made the aws s3 url explicit in settings.

Here is the PR for dependency update https://github.com/django/djangoproject.com/pull/2851

When we are happy with the patch, I will squash the commits.

PR #2845, opened 2026-09-12
DRY test workflows by merging their common parts

ulgens:
"Tests on Django main" workflow was duplicating the existing tests flow, causing extra maintenance load. Moved the common parts into a single file.

The same approach can be used for Docker tests too.

---

Used OpenCode with DeepSeek V4 Pro to handle the merger, then reviewed the result line by line.

### DIFF
--- a/.github/workflows/tests-base.yml
+++ b/.github/workflows/tests-base.yml
@@ -1,0 +1,86 @@
+name: Base Tests
+
+on:
+    workflow_call:
+        inputs:
+            django-version:
+                description: "Django version to test with, leave empty to use the pinned version."
+                type: string
+                required: false
+                default: ""
+        secrets:
+            codecov-token:
+                required: false
+
+env:
+    PYTHON_VERSION: 3.12
+
+jobs:
+    tests:
+        runs-on: ubuntu-24.04
+
+        services:
+            postgres:
+                image: postgres:17-trixie
+                env:
+                    POSTGRES_PASSWORD: postgres
+                options: >-
+                    --health-cmd pg_isready
+                    --health-interval 10s
+                    --health-timeout 5s
+                    --health-retries 5
+                ports:
+                    - 5432:5432
+
+        steps:
+            - uses: actions/checkout@v7.0.1
+            - name: Set up Python
+              uses: actions/setup-python@v7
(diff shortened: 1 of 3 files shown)


pauloxnet (approved):

.github/workflows/tests-base.yml
```diff
@@ -0,0 +1,86 @@
...
+                  python-version: ${{ env.PYTHON_VERSION }}
+            - name: Install gettext
+              run: sudo apt-get update && sudo apt-get install -y gettext
+            - name: Install dependencies
+              run: |
+                  python -m pip install --upgrade pip setuptools
+                  python -m pip install -r requirements/tests.txt
+                  if [ -n "${{ inputs.django-version }}" ]; then
+                      python -m pip install --upgrade --force-reinstall "${{ inputs.django-version }}"
+                  fi
```
sarahboyce, line 49:
```suggestion
- name: Install requested Django version
  if: inputs.django-version != ''
  run: python -m pip install --upgrade --force-reinstall "${{ inputs.django-version }}"
```
Perhaps we pull this out to it's own step?

.github/workflows/tests-base.yml
```diff
@@ -0,0 +1,86 @@
+name: Base Tests
+
+on:
+    workflow_call:
+        inputs:
```
sarahboyce, line 5: Perhaps we should add:
```
inputs:
  upload-coverage:
    description: "Whether to upload coverage results."
    type: boolean
    required: false
    default: false
```
Then further down have:
```
- name: Generate coverage report
  if: inputs.upload-coverage
  run: coverage xml

- name: Upload coverage reports to Codecov
  if: inputs.upload-coverage
  uses: codecov/codecov-action@v7.0.0
  with:
    token: ${{ secrets.codecov-token }}
```
As the main tests used to not generate coverage reports and I think it makes sense for these things to be skipable
We would then need in `tests.yml`
```
with:
  upload-coverage: true
```

sarahboyce (commented):
I love love this 😍 
I used to think reusing parts of GitHub actions wasn't possible, so I'm really glad to learn this exists!

I have a couple of minor suggestions

PR #2849, opened 2026-09-15
Prepared docs styles for the djangodocs Sphinx extension cleanup

nessita:
Three changes, see the commits. The first two are backward compatible with the documentation already published for every supported Django version; the third is a small, deliberate spacing change visible on all of them.

Django's custom `docs/_ext/djangodocs.py` Sphinx extension is being cleaned up so that the documentation can eventually be built with a maintained third-party Sphinx theme (django/django#21613, ticket-37212, ticket-37138).
Getting there means letting Sphinx render version annotations and console tabs itself, instead of overriding it.

This site's stylesheet currently assumes the output of those overrides. These changes remove those assumptions, so that the Django side can land without regressing how the docs look here, and so that both the current and the future markup render correctly side by side while older versions stay online.

Claude code was used to guide me thru the inter-dependencies between Django core changes and these required changes.


nessita:
## Before
image
image


## After
image
image

PR #2854, opened 2026-09-17
Added /events page with search and filters.

### LINKED ISSUE #2843: Add an "Events" page
Other than the "Upcoming Events" widget in the https://www.djangoproject.com/weblog/ page we don't have a way to share info about upcoming event or also past events.

It would be useful to have a "Events" page with "Upcoming events" and "Past events" sections where to find additional information, searchable.

saikat709:
This PR Closes #2843

In this PR I have:
- Added `/events` page showing upcoming and past events.
- Added search and filtering by location and year.
- Added "Events" link to header navigation and homepage sidebar snippet.
- Added some relevant tests in events/test file.


pauloxnet:
Thanks for this PR. I'll review ASAP 
I'm the meantime can you attach screenshot of new element added in the PR (e.g. events element in menu) and for the new events page with different states (es: totally empty, no future events, no past events, past and future events, filters in actions, ...) the more the better. And please add screenshots both for light and dark theme, and for mobile, and desktop resolution.

For this type of task I usually use https://github.com/simonw/shot-scraper or Firefox screenshot feature. But use whatever you prefer.

sylvesterkaczmarek (commented):
The location dropdown is built from exact distinct location values, but selecting one is applied with `location__icontains`. That means choosing a value such as `York` can also return events in `New York`, so the filter does not match the option the user selected. Could this use an exact or case-insensitive exact match for the location filter, while keeping the free-text `q` search as `icontains`?

jefftriplett:
Great work! I have one concern that's a carryover from: https://github.com/django/djangoproject.com/issues/2843#issuecomment-5604689901

If we're going to use the same Event model for everything, I would like a way to separate the conferences so we don't lose our promotion for them. DjangoCons, Django on the Med, and some of the Django Days need a bit more time and promotion than some meetups and possibly Django Girls events, but that might make it harder to see the international events. 

To be clear, meetups and possibly Django Girls events are extremely important, and if we put them all on one page, that's all the better. I don't want the DjangoCons + DotM + Day's existing placements to get lost in a sea of >50 events that are very regionally focused.

saikat709:
> Thanks for this PR. I'll review ASAP I'm the meantime can you attach screenshot of new element added in the PR (e.g. events element in menu) and for the new events page with different states (es: totally empty, no future events, no past events, past and future events, filters in actions, ...) the more the better. And please add screenshots both for light and dark theme, and for mobile, and desktop resolution.
> 
> For this type of task I usually use https://github.com/simonw/shot-scraper or Firefox screenshot feature. But use whatever you prefer.

Okay. I will add them here shortly.

saikat709:
> The location dropdown is built from exact distinct location values, but selecting one is applied with `location__icontains`. That means choosing a value such as `York` can also return events in `New York`, so the filter does not match the option the user selected. Could this use an exact or case-insensitive exact match for the location filter, while keeping the free-text `q` search as `icontains`?

Good catch! Thanks. I should update it to case-insensitive exact match for the location filter.

PR #2856, opened 2026-09-17
:bug: Made the sponsorship tier comparison table usable on phones

### LINKED ISSUE #2855: :tractor: Make `/sponsor/plans/` more mobile friendly
@nessita sent me this screenshot, and I will roll this into a follow-up to make `/sponsor/plans/` html table more mobile friendly. 

image (2)

_Originally posted by @jefftriplett in https://github.com/django/djangoproject.com/issues/2833#issuecomment-5719386757_

jefftriplett:
Fixes #2855.

You can now compare sponsorship tiers on a phone. The feature names stay on screen while you swipe across the tiers, so the checkmarks always line up with something you can read.

How Investing in Django Makes a Difference  Django


jefftriplett:
cc @nessita I think this flows much better.

nessita (commented):
Thank you for the quick turnaround!

I tested on a phone-width viewport (~375px) and at ~800px and the proposed solution works nicely. Two gaps that only appear once the column is pinned, for which I'll push minimal tweaks:

- The heading text was pinned only below 700px, so between 701px and 1000px the section heading slid away behind the pinned column while its row stayed put. It now pins from 1000px down, with the offset matching the padding.
- Only `td` carried a row's bottom border, so the line above a section heading stopped short of the first column, making the heading and the last row of the previous section read as one block. Row labels now get the same border.

nessita (approved):

PR #2858, opened 2026-09-18
:tractor: Fell back to the next sponsored tier on empty plan pages

### LINKED ISSUE #2857: :hammer: Sponsor plan pages with no sponsors should fall back to the next tier that has some
`/sponsor/plans/fellow/` and `/sponsor/plans/diamond/` have no sponsors yet, so those pages just drop the sponsor section. `/sponsor/plans/platinum/` shows "Meet our Platinum sponsors".

When a plan has no sponsors, fall back to the next tier down that does, and label it for that tier, so Fellow and Diamond would both show the Platinum sponsors.

The lookup is in `fundraising/views.py:81` (`sponsor_plan`) and the section is `djangoproject/templates/sponsor/plan.html:43`.

jefftriplett:
Fixes #2857.

The Fellow and Diamond pages have no sponsors yet, so they showed nothing at all. They now show the next tier down that does have sponsors, labelled as that tier, with a "Be our first Diamond sponsor" call to action above it.

The Diamond plan page showing a Be our first Diamond sponsor call to action above the Platinum sponsor logos


nessita (approved):
Thanks! 🌟 Reviewed and tested locally with and without a sponsor in the database. Approving, with a few things folded in and pushed:

- The invitation now also appears in the summary box above the fold, so it lands without scrolling; that button is filled in both states, with the wording switching to "Be our first Diamond sponsor" when the tier is empty.
- The section is no longer wrapped in `{% if sponsors %}`, so a tier with nothing below it to borrow from, Bronze in particular, still asks for its first sponsor rather than showing nothing.
- The view passes an explicit `plan_has_sponsors` flag instead of comparing slugs in the template, which also avoids relying on `None.slug` resolving to an empty string.
- Tests for the two states: a tier with its own sponsors keeps "Your annual investment" and no invitation, and an empty tier shows the invitation with no sponsor list.

PR #2866, opened 2026-09-18
:tractor: Gave the corporate ask its own block on the fundraising page

### LINKED ISSUE #2794: [sponsor efforts] Clean up the "donate" page to make it obvious how to sponsor
Right now the donate page seems more focused on individual donator than corporate ones. The corporate member option is down below in a 4 bullet list and the link takes the user to a long form without any explanation of what's the value they get from sponsoring.

There is also a link in the first bullet point about "why sponsor" but that's between parenthesis and feels like a "side thing" and not something readers should focus on. We should make the donate page clearer and more simple to attract new corporate sponsors. 

This is related to improving corporate member conversion in the DjangoCon US 2026 sprint.

cc/ @jefftriplett @ericholscher

jefftriplett:
Fixes #2794.

The "Why Sponsor?" link is out of the parentheses beside "47% funded" and the corporate ask is out of the donation-progress bullets. Both now sit in their own block below the donation form.


jefftriplett:
Before and after.

The fundraising page before, with Why Sponsor in parentheses and the corporate ask as a fourth bullet

The new Corporate sponsorship block with a Why sponsor Django button

jefftriplett:
Heart block before and after, same element and viewport, then the new block below it.

The heart block before, with Why Sponsor in parentheses beside 47% funded and the corporate ask as a fourth bullet

The heart block after, with three progress bullets and no corporate ask

The new Corporate sponsorship block with a Why sponsor Django button and a Compare sponsorship plans link

nessita (commented):
Lovely, thank you! 🌟 

This fixes the remaining of #2794. I pushed a few small changes on top:

- The intro line now jumps to the callout instead of being yet a third linkt to `/sponsor/`, and the callout's sentence is plain text. We had three links to `/sponsor/` within a few lines, which makes readers click to find out they all go to the same place.
- `scroll-margin-top` on the heading, so that jump lands on the whole block rather than flush against the top of the viewport.
- The actions row is a `div` rather than a `p`, and `sponsor_url` is defined once for the page.
- A test for the callout, since this is the entry point the ticket is about and nothing would have caught it disappearing.

One question for you @jefftriplett: I reworded the sentence to "can join as corporate members, with benefits that grow with each tier", because the old wording only worked as a link. Happy to take whatever phrasing you prefer.

Also spotted while in there, not fixed: the "Prospectus" link further down the page still points at `/sponsor/`, but the prospectus now lives at `/sponsor/plans/`. Worth a follow-up unless you want it here?

nessita:
Updated wording:
image

PR #2869, opened 2026-09-18
Removed redundant configuration for django-upgrade

p-r-a-v-i-n:
as per [docs](https://django-upgrade.readthedocs.io/en/latest/installation.html#:~:text=django%2Dupgrade%20attempts%20to%20parse%20your%20current%20Django%20version%20from%20pyproject%2Etoml) , setting the minimum version in `pre-commit-config.yml` become redundant since we already set the minimum version in `pyproject.toml`.
https://github.com/django/djangoproject.com/blob/a228ed3d1563a7dab3b2b57a647d688cce3662f2/pyproject.toml#L12


marksweb (approved):
LGTM 👍

p-r-a-v-i-n:
May be we don't even need `django-upgrade` hook at all since we always target the latest version ?

UnknownPlatypus:
It's stil useful if you copy paste code from another project that is using old syntax. The hook is instant anyways

ulgens:
> May be we don't even need django-upgrade hook at all since we always target the latest version ?

It's kinda the point of the hook. You target a specific version so your code doesn't use practices from older versions.

ulgens (approved):

pauloxnet (approved):

adamzap (approved):
👍 Looks great!

